PR #3, opened 2017-01-23
Feature: progress button

magalhini:
# Heads up: requires a new npm install

Imported AjaxButton as a renamed ProgressButton.

Styling for the loading needs a little bit of tweaking, but leaving this for later.
Also added the inline SVG loader to bring in the loader icon.

There were no tests for this component so for now it's just the bare minimum, needs a hella lot more coverage.


kyleshrives:
👍  for not adding tests right now whilst we refactor the components into this separate repo. Lets revisit tests as part of our refactoring process

PR #4, opened 2017-01-23
Feature: DropdownMenu & FontAwesomeIcon

magalhini:
This adds the DropdownMenu and a dependency of it, FontAwesomeIcon.

As such, the FontAwesome CSS library was also included as a dependency for Storybook.


kyleshrives:
Nice work. My only major concern is Font Awesome isn't kept separate from the core `gather-ui` codebase. I bet we could use FA as an external package - https://github.com/FortAwesome/Font-Awesome

magalhini:
Hah, I did see this repo but I assumed the "FortAwesome" was the newer (and different) version of the FontAwesome we're using. I now see that it seems to contain both, I'll check if they're interchangeable. If they are, I'll make the swap :)

kyleshrives:
What version do we need?

Here's the npm package if it helps - https://www.npmjs.com/package/font-awesome (again, depending on version)

PR #17, opened 2017-02-17
Feature hierarchical styleguide

magalhini:
So the refactoring (from a hierarchy perspective) isn't complete yet; but I'd really love your opinion @kyleshrives on the way this is going. Naturally the `Colours` and `Typography` stories merely act as placeholders for what will be there in the future, but it's more about the overall structure and approach here.

Let me know your thoughts :)


kyleshrives:
I like it a lot. Nice implementation with separating stories as `StoryItems` too. What is left to complete before this gets merged in?

magalhini:
All that was left was to move the remaining components to the same format, which was now complete, so I'll be merging this now. There's still a few scenarios that I'd like to explore regarding how we feature the components behaviour that I couldn't get to work, for e.g, having certain components only showing at the click of a button or something similar.

Basically, present the components based on some kind of `state`, but this doesn't seem to be possible inside a Story. I'm sure we'll find a way, though!

PR #18, opened 2017-02-17
Button component refactor

kyleshrives:
Not completely ready but wanted to get some visibility on it now. I just need to refactor the styling side of it. A refactoring of the tests are included too 👍 

Thoughts?

### DIFF
--- a/lib/Button/index.js
+++ b/lib/Button/index.js
@@ -1,29 +1,32 @@
 import React, { Component, PropTypes } from 'react';
 
-// Ideally, we'd just be doing this, but we need to account for Sass
-// mixins and settings... to study a bit more.
-// require('./styles.scss');
+/**
+ * Example Usage
+ *
+ * <Button
+ *  types={['clear', 'collapsed']}
+ *  clickHandler={() => {}}
+ * >
+ *   ...test goes here
+ * </Button>
+ */
 
 class Button extends Component {
-  static getClassName(type) {
-    switch (type) {
-      case 'primary':
-        return 'btn btn--primary';
-      case 'secondary':
-        return 'btn btn--secondary';
-      case 'danger':
-        return 'btn btn--warning';
-      case 'link':
-        return 'btn btn--link';
-      case 'clear':
-        return 'btn btn--clear';
-      case 'clear-and-collapsed':
-        return 'btn btn--clear btn--collapsed';
-      case 'neutral':
-      default:
-        return 'btn btn-default--light';
-    }
-  }
+  static defaultProps = {
+    types: ['primary'],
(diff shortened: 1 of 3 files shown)


lib/Button/index.js
```diff
@@ -1,29 +1,32 @@
...
-      case 'clear':
-        return 'btn btn--clear';
-      case 'clear-and-collapsed':
-        return 'btn btn--clear btn--collapsed';
-      case 'neutral':
-      default:
-        return 'btn btn-default--light';
-    }
-  }
+  static defaultProps = {
```
kyleshrives: Switched to this placement for default Props and prop validation. I felt like I was missing out on knowledge about the component by keeping these at the bottom of the file.

Moving it up like this means that I digest the contents of this component a little better. What do you think @magalhini?

lib/Button/index.js
```diff
@@ -33,42 +36,31 @@ class Button extends Component {
...
     this.handleOnClick = this.handleOnClick.bind(this);
+    this.getTypeClasses = this.getTypeClasses.bind(this);
   }
 
   handleOnClick() {
     this.setState({ disabled: this.props.disableOnClick });
     this.props.clickHandler();
   }
 
+  getTypeClasses() {
```
kyleshrives: I thought it'd be a neat idea to move the previous "type" prop for a "types" prop which allows us to define multiple types that are then used to string together the modifier classes e.g. `primary` and `collapsed`.

I wasn't super keen on the switch statement used in the previous component. Mostly because when I wanted to piece together types like `clear` and `collapsed` I had to add a brand new type.

However, the previous version did do a great job at describing the main types of Button you could use. What do you think @magalhini?

lib/Button/index.js
magalhini: Highly agreed! This makes it a bit more clear when opening a component's file and seeing what it's expecting 👍

lib/Button/index.js
magalhini: I see what you mean, there's a plus and a down side to this approach. On one hand it's a much cleaner (and more maintainable) approach; on the other we lost the ability to quickly scan the possibilities that are supported by our styleguide.

Maybe keeping the documentation comments at the top with all the possible options is a way to mitigate this, or to declare the component's usage for all cases in the actual style guide could also be a way.

kyleshrives: "or to declare the component's usage for all cases in the actual style guide could also be a way"

I think that's a great idea. We'd be doing this anyways and using clear language in the storybook means it should be easy to understand how to use the types prop.

That way we resolve the issue with the switch statement and make it clear what types we support!

PR #24, opened 2017-02-21
Add ITCSS structure

magalhini:
This refactors the existing CSS into alignment with the proposed hierarchy using the ITCSS structure. 

It also adds a script that gets build on `build` to copy over the Sass configuration variables into the `/dist/settings/` folders so they can be consumed from the main application.

**This needs a new `npm run install`.**


styles/main.scss
```diff
@@ -34,12 +34,3 @@
  * sparingly. Color overrides, forced floats, and text alignment classes can go here.
  */
 @import 'overrides/helpers';
-
-/* @ Styleguide
- * Styles for the components that exist only for styleguide purposes.
- */
-@import 'styleguide/main.scss';
```
magalhini: Great catch, thanks for noticing this!

PR #33, opened 2017-02-28
Notification Cleanup

kyleharper:


lib/Notification/styles.scss
```diff
@@ -1,9 +1,7 @@
 .notification {
-	font-size: 13px;
-	line-height: 1.75em;
-	padding: 10px 15px;
+	font-size: $typo-size-slight;
+	padding: $layout-spacing-base/2 $layout-spacing-base*0.75;
```
magalhini: Would it make sense to add fixed values for these, or to contain them in separate values without making loads of math already? :) (just pinging some thoughts, really!)

kyleharper: @magalhini I've got a great idea about how we can handle this moving forward. Going to propose something in the PR related to progress.

magalhini: Going to hold the merging of this PR then until we discuss that strategy so we can come back to this :)

I know what you mean about only using the base variables to perform the calculations, and I'm all up for that, but these edge cases really complicate things (like finding factors of multiplication to achieve the same effect. So far I've found calculations like `*2` or `/2` to be pretty maintainable, but multiplying by different factors can sometimes be hard to justify (what if a component calls for `* 0.65` in order to look alright?)

A chat for us to have later on!

kyleharper: I suggest we use the Style as Config approach I mention in #35. I will amend the styles to match that approach.

kyleharper:
@magalhini when ready review the changes and that should resolve the issue you raised, well caught. Note that I have abstracted the overrides into the App (check there for the PR).

PR #34, opened 2017-02-28
SearchInput Component Refactor

kyleharper:


lib/SearchInput/index.js
```diff
@@ -32,19 +31,19 @@ class SearchInputField extends Component {
     const { placeholder } = this.props;
     const hasSearch = this.state.query ? 'is-visible' : '';
 
-    const iconClass = (() => `fa fa-search c-input-search__icon ${hasSearch}`)();
+    const iconClass = (() => `fa fa-search search-input__icon ${hasSearch}`)();
```
magalhini: I didn't spot this initially, but this can be turned into a FontAwesome component since it is one! Not a biggie, but maybe we can easily add this in 😃

kyleharper: Good shout. I've switched this round and done a little cleanup on the other const :)

PR #35, opened 2017-02-28
Refactored progress bar

kyleharper:
As well as cleaning up any SCSS / JS for the progress component I've also come up with a way for how we can abstract our style inconsistencies away from `gather-content-ui`.

**Style as Config**

I forgot that I have done this before (in previous roles) but abstracting CSS prop values into a default config per component is a really useful way to allow us to keep the UI library styles consistent and easy to digest.

For example;

```
// Notification/styles.scss - in gather-content-ui

$notification-padding: 15px 10px !default;

.notification {
    padding: $notification-padding;
}

// settings/overrides/notification.scss - in app using package

$notification-padding: 13px 8px; // this padding will be used when 
```

**Most of the Work**

At the moment we're digesting the CSS from `gather-content-ui` as a complied CSS file. This strategy would require us to expose all SCSS files (like how we're exposing settings).

I kind of feel like we should do this anyways. Our only way to digest the CSS atm is to `require` it. Not the worst thing in the world but it'd be nice to have the ability to import our main SCSS file in the same way we expect `settings` to be imported.

**Proposal**

We don't expose the SCSS right away but use this style as config approach to set things up to be more flexible moving forward.

What do you think?


magalhini:
Like where this is going 👍

PR #40, opened 2017-03-03
Refactor: ModalConfirm

magalhini:
This brings the ModalConfirm component from UI in-app to the library.
I've renamed it Confirmation Modal, since I believe this makes a bit more sense considering its use case. There's no styling changes since this component uses Bootstrap only.

On the side, this PR also fixes the failing tests for the SearchInput due to classNames that were changed (one more reason I'd like for us to move to `data-attributes` for testing).


kyleharper:
Loving the change of name on this @magalhini 👍

kyleharper:
@magalhini My only piece of feedback is that it would be great to get a trigger on the story so we can open it up again. What do you think? Shouldn't be too difficult to do right?

Apart from that its all good.

PR #41, opened 2017-03-03
Refactor Components to use Style as Config

kyleharper:
I've gone through every components (except the DropdownMenu) and taken all of the values we'd want to override when we consume the package.

The biggest change on this to note is that I have moved Bootstrap overrides into the overrides directory as they're not styles for our modal component. I did something similar for the Tooltip styling (you can see this in `overrides/_tooltip.scss`.

As doing this I have noticed a few things that I will raise issues for so we can sort them out.


lib/Button/styles.scss
```diff
@@ -1,3 +1,27 @@
+/* ==========================================================================
+   Config
+   ========================================================================== */
+$color-btn-primary--bg: rgb(74, 181, 69) !default;
```
kyleharper: @magalhini these should be prefixed as `btn` not `color-btn`. Changing now 👍

lib/SearchInput/styles.scss
```diff
@@ -1,3 +1,11 @@
+/* ==========================================================================
+   Config
+   ========================================================================== */
+$search-input-clear-hover-bg: rgba($color-btn-default-light--bg, .5) !default;
```
magalhini: Did not know about this! Thanks for the lesson of the day 👍

PR #48, opened 2017-03-06
Fix linting errors

kyleharper:
All linting errors are now fixed. 97 in total 👍


kyleharper:
I've noticed a couple of bugs caused by this. Fixing now.

PR #52, opened 2017-06-05
Conversation Component

kyleharper:
This PR contains most of the component logic for conversations.


.storybook/config.js
```diff
@@ -1,4 +1,4 @@
-import { configure } from '@kadira/storybook';
+import { configure } from '@storybook/react';
```
kyleharper: This (plus everything related to webpack) is for the storybook upgrade.

kyleharper:
[[Editor] Comment Styling](https://trello.com/c/wTLZYydP/2276-%5Beditor%5D-comment-styling)

lib/BoundaryClickWatcher/index.js
```diff
@@ -0,0 +1,64 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+
+class BoundaryClickWatcher extends Component {
```
kyleharper: This component is responsible for calling functions (passed down as props) when a user clicks outside of a component. To optimise events the outside click checking isn't done until the initial click is done and the component is active.

lib/Button/index.js
```diff
@@ -55,13 +55,15 @@ class Button extends Component {
   }
 
   render() {
-    const { disabled, children, className } = this.props;
+    const { disabled, children, className, isSubmit } = this.props;
+    const type = (isSubmit) ? 'submit' : 'button';
```
kyleharper: When using buttons in forms that use `onSubmit` we get issues where all buttons submit the form. Adding this in resolves that issue.

lib/Button/styles.scss
```diff
@@ -89,9 +89,19 @@ $color-btn-default-dark--hover: #5e5e5e !default;
 }
 
 .button--link-default {
-  color: $color-btn-secondary--hover;
+  color: $color-brand-blue;
```
kyleharper: `button--link` doesn't give us a proper link styling but seeing as its used else where I have created a `link-default` modifier to get the blue text colour.

We also needed a link button modifier for danger links too.

lib/Conversation/styles.scss
```diff
@@ -0,0 +1,166 @@
...
+  .conversation__foot,
+  .conversation__resolve,
+  .comment {
+    display: block;
+  }
+}
+
+
+// Form
+.form__input {
```
kyleharper: Should we stick this in a form component scss file or maybe just prefix this with conversation?

kyleharper:
[[Editor] Field Styling](https://trello.com/c/z3MlFu1w/2287-%5Beditor%5D-field-styling)

lib/Conversation/styles.scss
```diff
@@ -116,9 +127,24 @@ $conversation-meta-text-color: $color-brand-dark !default;
...
   .form__actions {
     margin: 0 0 0 $layout-spacing-base/2;
   }
+}
+
+.form__drop-down {
+  position: absolute;
+}
+
+.form__button {
```
magalhini: Good job on the styling! I like the modularity of it all 👍 

My only concern is about the namespace of this `.form` component. What do you think about making it more specific in namespace? I've got a feeling that `.form__input` and `.form__actions` and so on will eventually accidentally clash once we flush out the actual Form components outside the conversations work. Thoughts?

kyleharper: Good point. My initial though is that we should be commanding the names of our classes so using `form` should be fine.

Saying that, if `form` is a little too generic then maybe for now we could hook this up with something like `comment-form` rather than `form`? What do you think?

lib/Conversation/index.js
```diff
@@ -29,41 +34,55 @@ class Conversation extends Component {
...
 
+  deactivateConversation() {
+    this.props.actions.deactivateConversation(this.props.id);
+    this.setState({
+      isActive: false,
+    });
+  }
+
   render() {
     const conversationClasses = (this.state.isActive) ? 'conversation--is-active' : '';
```
magalhini: Just a reminder that we have the `classnames` (cx) package available to help us handle conditional class names! It's okay for these because it's either on or off 😉

stories/components/Conversation.js
```diff
@@ -5,16 +5,16 @@ import StoryItem from '../styleguide/StoryItem';
 
 const mockComments = [{
   id: 11,
-  body: 'Comment body...',
-  createdAt: 'today',
+  body: 'Here is a decent size comment that was created by someone who wanted to comment.',
+  createdAt: 'Less than a minute ago',
   person: {
     avatar: 'https://gathercontent-production-avatars.s3-us-west-2.amazonaws.com/208205_yHGd7vA5HRxsnMQpES4UzjJ7Yxgn6Bp54165gqksRXyDJhuOnW88H6djhLJeE2BZ.jpg',
```
magalhini: Shall we leave handling the lack of an image URL falling back to the user's initials for later? (more than happy to do this a later stage)

lib/Conversation/index.js
```diff
@@ -49,7 +50,7 @@ class Conversation extends Component {
   }
 
   render() {
-    const conversationClasses = (this.state.isActive) ? 'conversation--is-active' : '';
+    const conversationClasses = (this.state.isActive || this.props.isActive) ? 'conversation--is-active' : '';
```
magalhini: A slight BEM concern; I'd argue that things like `is-active`, `is-collapsed` and whatnot are part of BEM state, not modifiers. This makes it a bit weird having in the SCSS nested classes inside modifiers.

For example, we have now:

```
.conversation--is-active {
  border-color: $conversation-border-color;

  .conversation__resolve {
    border-bottom: $conversation-border-width solid $conversation-border-color;
  }
```

It's a bit awkward deleting so much parent-like reponsibility to a modifier, instead of a state. Does this make sense? :)

lib/Conversation/styles.scss
magalhini: That was my initial thought too, something like comment-form is super declarative. Two 🌮 s for that!

lib/BoundaryClickWatcher/index.js
magalhini: Epic work here, I really like this component. Just out of curiosity, what was your thinking behind calling the `targetIsCounter` variable? Did you mean `targetIsContainer`?

kyleharper: Yep, that's a typo. Good find 👍

lib/Conversation/styles.scss
magalhini: I should have seen this comment earlier 😅 again, yeah let's go with the prefix for now. We could always extract this later to a comment-specific styling if it's not used anywhere else, too.

lib/Conversation/index.js
kyleharper: Yep that makes sense. We should handle this as a state class and not a BEM modifier 👍

stories/components/Conversation.js
kyleharper: Aye, lets handle this as another task 👍

PR #54, opened 2017-06-08
FileCard Component

magalhini:
This is the initial styling and very basic functionality for the File Cards with and without previews.
I didn't go too far yet writing the actions, as I feel like this would be a separate discussion still worth having (e.g., what happens when I click download, should this be handled by the component itself or always passed down as a prop action, etc).


lib/FileCard/index.js
```diff
@@ -0,0 +1,78 @@
+import React, { Component, PropTypes } from 'react';
```
kyleharper: use `prop-types` package 😄

lib/FileCard/index.js
```diff
@@ -0,0 +1,78 @@
...
+class FileCard extends Component {
+  static propTypes = {
+    image: PropTypes.string,
+    hasComments: PropTypes.bool,
+    filename: PropTypes.string,
+    label: PropTypes.string,
+    type: PropTypes.string,
+  }
+
+  constructor() {
```
kyleharper: constructor isn't doing anything, I believe this will cause an error with ESLint.

lib/FileCard/styles.scss
```diff
@@ -0,0 +1,123 @@
...
+  .file-card__action:focus {
+    @for $i from 1 through 4 {
+      .file-card__action:nth-child(#{$i}) {
+        transform: translateY(0);
+        transition-delay: #{$i/20}s;
+      }
+    }
+  }
+
+  &.has-comments {
```
kyleharper: It would be worth discussing where we should use state classes and BEM modifiers. Every BEM modifier could be argued as a state class but I'd rather keep these state classes to a select generic few e.g;

- `is-active`
- `is-disabled`
- `has-loaded` 

Things like `has-comments` could be abstracted into something like `file-card--highlight` to say what the styling is doing rather than why we're applying the style.

Pros and cons to both but we need to align as to when we should use either. E.g. is there a difference between the reason why we have `has-comments` and `file-card--no--preview`?

lib/FileCard/styles.scss
```diff
@@ -0,0 +1,123 @@
...
+.file-card__thumbnail--no-preview {
+  background-color: rgba(155, 155, 155, .1);
+  align-content: center;
+  justify-content: center;
+  font-size: 2em;
+  text-transform: uppercase;
+  display: flex;
+  padding-top: 35%;
+  overflow: visible;
+  font-weight: 600;
```
kyleharper: use variable instead of fixed value

lib/FileCard/styles.scss
```diff
@@ -0,0 +1,123 @@
...
+  &:hover {
+    cursor: pointer;
+  }
+}
+
+.file-card__thumbnail--no-preview {
+  background-color: rgba(155, 155, 155, .1);
+  align-content: center;
+  justify-content: center;
+  font-size: 2em;
```
kyleharper: use variable instead of fixed value

lib/FileCard/styles.scss
```diff
@@ -0,0 +1,123 @@
...
+  font-weight: 600;
+}
+
+.file-card__label {
+  background: $field-card-label-bg;
+  color: #fff;
+  position: absolute;
+  bottom: $field-card-actions-padding;
+  font-size: $field-card-label-font-size;
+  border-radius: 3px;
```
kyleharper: The same border-radius is used else where, maybe move to config?

lib/FileCard/styles.scss
```diff
@@ -0,0 +1,123 @@
...
+  text-overflow: ellipsis;
+  overflow: hidden;
+  white-space: nowrap;
+  align-self: center;
+  text-align: center;
+}
+
+.file-card__actions {
+  display: flex;
+  padding: .5rem;
```
kyleharper: Can this use the config variable instead?

lib/FileCard/styles.scss
```diff
@@ -0,0 +1,123 @@
...
+  justify-content: flex-start;
+
+  > .file-card__action {
+    @for $i from 1 through 4 {
+      &:nth-child(#{$i}) {
+        transform: translateY(-3rem);
```
kyleharper: It might be nice moving the transition stuff to config to manipulate the animation faster.

tests/FileCard/index.js
```diff
@@ -0,0 +1,45 @@
+import { React, expect, sinon, shallow } from '../setup';
+import FileCard from '../../lib/FileCard';
+
+describe('FileCard', () => {
+  let wrapper;
```
kyleharper: are we missing a test for when we don't have a preview?

lib/FileCard/styles.scss
```diff
@@ -0,0 +1,123 @@
...
+  background: rgba(20, 20, 20, .6);
+  border-radius: 3px;
+  transition: .2s all ease-in-out;
+
+  &:last-child { margin-right: 0; }
+
+  &:hover {
+    background: rgba(10, 10, 10, .9);
+  }
+}
```
kyleharper: There will be a number of ESLint errors around end of files needed a new line.

lib/FileCard/index.js
```diff
@@ -0,0 +1,78 @@
...
+    return <div className="file-card__thumbnail file-card__thumbnail--no-preview">{ext}</div>;
+  }
+
+  render() {
+    const { filename, image, type } = this.props;
+    const actions = this.getActions(type);
+
+    const url = `url(${image})`;
+    const classes = cx({
+      'file-card': true,
```
kyleharper: Not too sure I dig the "`file-card`: true" way of always rendering that class out. I'd prefer we just add it to `className` always. What are the pros?

lib/FileCard/index.js
```diff
@@ -0,0 +1,78 @@
...
+    filename: PropTypes.string,
+    label: PropTypes.string,
+    type: PropTypes.string,
+  }
+
+  constructor() {
+    super();
+  }
+
+  getActions(type) {
```
kyleharper: Really like this method 👍

lib/FileCard/index.js
magalhini: Ops, force of habit! Good shout 👍

lib/FileCard/index.js
magalhini: Yeah, it will (and it does). I'm just sure we'll need a class based component once we actually add more complexity to this component with the upload stuff. But you're right, and I'll follow my own advice regarding premature optimisation 👍

lib/FileCard/index.js
magalhini: In this particular case, maybe not as many pros, but more than two classes being checked against (meaning, no longer a ternary operator) and I begin to be a big fan of `classNames` since it removes the ambiguity of props => classes checking.

Since we only have two for now, I've no problem in making this a ternary operator :)

lib/FileCard/styles.scss
magalhini: I'll probably take care of this once (and if!) @nhqe approves the animation at all to begin with, if that's alright! No point in optimising it now if it will be taken away in the end :)

lib/FileCard/styles.scss
magalhini: I can dig that 👍 while the new state really does look like a simple state change (from non highlighted to highlighted), the underlying premised is that it received comments.

But this also leads me to an interesting discussion 😄 if we apply something like `file-card--highlight` to the state in which it has comments, and the design changes from something totally different than being highlighted, then the modifier name stops making sense... if it had been `file-card--with-comments`, then the modifier name would still make sense regardless of the visual styling.

Of course this is hardly the deepest CS issue 😂 but thought it would be worth trowing it into the pan and think about the pattern!

tests/FileCard/index.js
magalhini: There is one, but it's only testing against the number of possible actions. I'll add more to it, but in any case these will have to be revisited once we add more functionality to them :)

PR #55, opened 2017-06-12
Conversation Component Integration Changes

kyleharper:
During the integration we noticed a few things that had to be fixed. Here are the fixes.

--------------

After this we should patch his package and publish 👍

### DIFF
--- a/lib/Conversation/CommentForm/CommentFormActions.js
+++ b/lib/Conversation/CommentForm/CommentFormActions.js
@@ -30,7 +30,7 @@ const CommentFormActions = props =>
       </Button>
       <Button
         types={['secondary', 'size-sm']}
-        clickHandler={props.onSubmit}
+        clickHandler={() => {}}
         isSubmit
       >
         Send
--- a/lib/Conversation/CommentList.js
+++ b/lib/Conversation/CommentList.js
@@ -5,13 +5,13 @@ const CommentList = ({ comments, ...rest }) => {
   const list = comments.map(comment =>
     <Comment
       key={comment.id}
-      comment={comment}
       {...rest}
+      {...comment}
     />,
   );
 
   return (
-    <div className="comment-list">
+    <div className="conversation__comment-list">
       {list}
     </div>
   );
--- a/lib/Conversation/CommentList.js
+++ b/lib/Conversation/CommentList.js
@@ -1,4 +1,5 @@
-import React, { PropTypes } from 'react';
+import React  from 'react';
+import PropTypes from 'prop-types';
 import Comment from './Comment';
 
 const CommentList = ({ comments, ...rest }) => {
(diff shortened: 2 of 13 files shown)


kyleharper:
[[Editor] Field Styling](https://trello.com/c/z3MlFu1w/2287-editor-field-styling)

lib/Conversation/CommentList.js
```diff
@@ -5,13 +5,13 @@ const CommentList = ({ comments, ...rest }) => {
   const list = comments.map(comment =>
```
magalhini: The `comments` property on this component has no defaultProp set; maybe it will be okay anyway because an empty array is perhaps being sent from the parent, but for linting's sake it's probably worth adding a `defaultProps` to comments on this component.

lib/Conversation/index.js
```diff
@@ -15,6 +16,7 @@ class Conversation extends Component {
       activateConversation: PropTypes.func.isRequired,
```
magalhini: I can't comment on the specific line here, but I believe `id` should also have an `isRequired` to its propTypes. If there's a conversation being created, an ID truly is required for everything else to work :)

lib/Conversation/index.js
```diff
@@ -29,7 +31,7 @@ class Conversation extends Component {
   }
 
   addComment(commentBody) {
-    this.props.actions.addComment(commentBody, this.props.id);
+    this.props.actions.addComment(commentBody, this.props.id, this.props.user);
```
magalhini: Not a biggie, but `user` seems to be missing from prop validation.

lib/Conversation/CommentForm/CommentFormActions.js
```diff
@@ -30,7 +30,7 @@ const CommentFormActions = props =>
       </Button>
       <Button
         types={['secondary', 'size-sm']}
-        clickHandler={props.onSubmit}
+        clickHandler={() => {}}
         isSubmit
       >
         Send
```
magalhini: Why was the `props.onSubmit` removed in favour of an empty function here? I'm guessing it still makes sense to run the submit callback at this stage (and the cancellation one is still in place) :)

kyleharper: The form has a `onSubmit` property which calls the same function. Calling that function here causes the function to be called twice. We simply want a button that submits the form.

lib/Conversation/CommentList.js
kyleharper: `comments` is a required prop in the parent component so I don't think it needs a default prop. Instead I'll make it required 👍

lib/Conversation/index.js
kyleharper: Good catch, this has been done 👍

lib/Conversation/index.js
kyleharper: Great catch, this has been done 👍

lib/Conversation/CommentForm/CommentFormActions.js
magalhini: Ah, indeed it does! Then it can be removed from the props validation too, that was what initially caused my confusion, it being in two places :)

PR #56, opened 2017-06-12
Choices and Radios component styling

magalhini:
Basic styling for the fields regarding Choices and Radios.

Note: these will most likely have to be revisited once we hook them up and use them in the Item Editor (more control over properties, more custom states, etc).


kyleharper:
@magalhini There's a lot of `file-card` related work in this PR, is that correct?

lib/Form/Checkbox/styles.scss
```diff
@@ -0,0 +1,17 @@
+/* ==========================================================================
+   Config
+   ========================================================================== */
+$form-checkbox-margin: $layout-spacing-base/2 !default;
+$form-checkbox-font-weight: 400;
```
kyleharper: use global variables where possible e.g. this is `$typo-weight-roman`.

magalhini: @kyleharper There's not supposed to be... I've checked and there's a master merge that somehow brought in specific `filecard-component` into it, not sure what happened with that. 

But the merge is clean and should be good to go regardless of that (I'll investigate this further though, quite curious!). Only those 3 commits regarding these small components. I'll add the variable for the type weight, completely missed that one 👍

PR #61, opened 2017-07-05
Presence Avatars & Tooltip Components

magalhini:
This adds the more complex Avatars component, which can be used within an AvatarGroup component in order to form more groups of avatars.
It also implements a custom Tooltip element, mostly a wrapper around the Popover already provided by Bootstrap.

I expect some changes to be requested regarding the existing solo `Avatar` component, which now sits in isolation. I don't think there's a need to keep it around, as it's easily replaceable with this new one. Will likely rename the `AvatarPresence` to just `Avatar`... thoughts?


lib/AvatarPresence/index.js
```diff
@@ -0,0 +1,62 @@
...
+  <Tooltip className="avatar__tooltip" id={data.email}>
+    <p className="avatar__tooltip-name">{data.name}</p>
+    <p className="avatar__tooltip-email">{data.email}</p>
+    { isAssigned &&
+      <span className="avatar__pillbox">Assigned</span>
+    }
+  </Tooltip>
+);
+
+const AvatarPresence = (props) => {
```
kyleharper: What's your thinking around the name `AvatarPresence`?

I noticed that in the stories area we use this alongside a `AvatarPresenceGroup` component but the file is called `AvatarGroup`. Hitting it with a simple stick I thought of names such as;

- Avatar (with added initials & custom colour support)
- AvatarGroup (handles collapsing etc)
- AvatarWithTooltip (provides a tooltip with name/email + additional data if required e.g. assigned)

In the UI scope where does "presence" come into play? Once I understand that I think I'll get your thinking behind the name a little more 👍 

-------------------

A good example of the ComponentWithTooltip approach is the `ButtonWithTooltip` component.

kyleharper: Adding a little more to this I forgot to mention your comment;

> I expect some changes to be requested regarding the existing solo Avatar component, which now sits in isolation. I don't think there's a need to keep it around, as it's easily replaceable with this new one. Will likely rename the AvatarPresence to just Avatar... thoughts?

In my opinion there's still a need for the original Avatar component. AvatarPresence as it currently stands does 2 things;

- renders an avatar
- renders a tooltip

What if I wanted an Avatar without a tooltip? We could add extra props but I think its much cleaner to have 2 separate components one responsible for rendering an Avatar and handling all the custom options it supports (colours etc...) and then another component which uses the Avatar & Tooltip component. 

Again look at how `ButtonWithTooltip` works, the component name is super clear and functionality its just using component composition which is again easy to understand.

lib/Tooltip/index.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Popover from 'react-bootstrap/lib/Popover';
+
+const Tooltip = props => (
+  <Popover className={props.className} id={props.id} {...props}>
```
magalhini: Good catch! Realised later I had to actually spread all props but didn't remove the id 👍

lib/AvatarPresence/index.js
magalhini: That's exactly what I wanted to discuss :) and why I left it this way on purpose, in fact. 
The way I see it we can remove the simple Avatar component that's still currently there (just a wrapper for an `img` tag), and replace it with this one (would still be called `Avatar`).

I like the `AvatarWithTooltip` approach, I'll make the move in that direction 👍

lib/AvatarPresence/styles.scss
```diff
@@ -0,0 +1,101 @@
...
+}
+
+.avatar__image {
+  width: 100%;
+  height: 100%;
+  overflow: hidden;
+  border-radius: 50%;
+}
+
+.avatar-group-list {
```
magalhini: Hmm, interesting food for thought. I'd argue we should keep it as a higher parent component, since an `AvatarGroup` contains `Avatar` components and not the other way around. So `avatar-group` instead of `avatar__group` literally reflects the DOM hierarchy too.

Thoughts? :)

kyleharper: I've had this issue in the past where a "container" element surrounds the base element and how to handle it.

I've always kept things simple and treat the container as a element (e.g. `avatar__group`). Otherwise we're name spacing a component that only ever exists when using Avatar which can make things harder to manage.

BEM elements can be both children and parents of the base component IMO so I'd vote for `avatar__group`.

lib/AvatarPresence/styles.scss
```diff
@@ -0,0 +1,101 @@
...
+   ========================================================================== */
+.avatar {
+  border-radius: $avatar-border-radius;
+  width: $avatar-size;
+  height: $avatar-size;
+}
+
+.avatar--presence {
+  display: inline-block;
+  border-radius: 50%;
```
magalhini: Yeap, I'm replacing the original styles with these new ones anyway, so no more `avatar--presence` :)

magalhini:
@kyleharper I think I've addressed most of all the feedback. Let me know if these are okay for merging 😄

lib/Avatar/AvatarWithTooltip.js
```diff
@@ -0,0 +1,46 @@
+import React, { PropTypes } from 'react';
+import OverlayTrigger from 'react-bootstrap/lib/OverlayTrigger';
+import Tooltip from '../Tooltip/';
+import { Avatar } from '../';
+
+const AvatarWithTooltip = (props) => {
```
kyleharper: This component rocks 👍

lib/Avatar/AvatarWithTooltip.js
```diff
@@ -0,0 +1,46 @@
+import React, { PropTypes } from 'react';
```
kyleharper: Use `prop-types` package.

lib/Avatar/index.js
```diff
@@ -1,17 +1,47 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-const Avatar = props =>
-  <img
-    className={`avatar ${props.className}`}
-    src={props.src}
-    alt={props.altText}
-  />;
+const Avatar = (props) => {
```
kyleharper: Nice refactoring from the previous work.

lib/Avatar/styles.scss
```diff
@@ -3,15 +3,99 @@
...
+
+$avatar-show-more-background: rgb(175, 165, 175);
+$avatar-faded-background: rgb(60, 50, 60);
+$avatar-shadow: 0 0 0 1px rgba(100, 100, 100, 0.2) inset;
+$avatar-pillbox-padding: $typo-size-lead/3 $typo-size-lead/2 !default;
 
 /* ==========================================================================
    Styles
    ========================================================================== */
 .avatar {
```
kyleharper: Can we merge these styles with the other `.avatar` rule?

styles/settings/_colours.scss
```diff
@@ -37,4 +37,9 @@ $color-table-item--hover: #F1FCFF;
 /* ==========================================================================
    Borders
    ========================================================================== */
-$color-input-border: #eaeaea;
\ No newline at end of file
+$color-input-border: #eaeaea;
+
+/* ==========================================================================
+   Components
```
kyleharper: What's your thinking around this new section in settings?

tests/Avatar/AvatarPresence.js
```diff
@@ -0,0 +1,63 @@
+import { React, expect, sinon, mount } from '../setup';
+import Avatar from '../../lib/Avatar';
+import AvatarPresenceGroup from '../../lib/AvatarGroup';
```
kyleharper: Do we need to rename `AvatarPresenceGroup` to `AvatarGroup`?

tests/Avatar/AvatarPresence.js
```diff
@@ -0,0 +1,63 @@
...
+
+describe('Avatar/Presence', () => {
```
kyleharper: Might be worth renaming to match your refactor.

lib/Avatar/AvatarWithTooltip.js
magalhini: 🤦‍♂️

magalhini:
@kyleharper Thanks dude! Addressed things :)

PR #63, opened 2017-07-07
Start new status indicator

bagwaa:


petemcfarlane:
[[Editor] Manage Workflow Component](https://trello.com/c/yBpFMaYg/2355-editor-manage-workflow-component)

lib/StatusIndicator/index.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const StatusIndicator = props => 
+  <span>
```
kyleharper: It's be good if we add a base class to this for styling purposes e.g. `status-indicator`.

lib/StatusIndicator/index.js
```diff
@@ -0,0 +1,15 @@
...
+    <i className="fa fa-circle" style={{ color: props.color }}></i> { props.label }
```
kyleharper: If this will just be a coloured circle I don't think we need `fa`. Plus we'll move away from that at some point (not super keen on font icons).

How about using a BEM element class e.g. `status-indicator__circle`?

kyleharper: I really like the use of `props.label` here.

If there's going to be some design difference between the `label` and `children` prop then I recommend we wrap the label in a span and maybe give it a element className.

lib/StatusIndicator/index.js
```diff
@@ -0,0 +1,15 @@
...
+    { props.children }
+  </span>
+
+StatusIndicator.propTypes = {
+  color: PropTypes.string,
+  label: PropTypes.string,
```
kyleharper: `children` is a prop that also requires prop checking :)

kyleharper:
@bagwaa btw don't forget to add a test for this component 👍

lib/StatusIndicator/index.js
petemcfarlane: we weren't sure what to add here, as it might not always be a string, and it isn't required. Do you have any suggestions @kyleharper ?

kyleharper: Something like this could work, basically check to see if the prop is either a string or a node 👍 

```
children: PropTypes.onOfType([
    PropTypes.string,
    PropTypes.node,
])
```

lib/StatusIndicator/index.js
bagwaa: We were speaking about styling inside components, do you think we should add some base styles? We kind of thought that styling should be something done in the app?

lib/StatusIndicator/index.js
bagwaa: Yeah I think initially we just had this down as `isRequired` the above makes sense

lib/StatusIndicator/index.js
kyleharper: Styling for the component is handled in `gather-content-ui` not the project that consumes it.

If needs be you can override styles using SCSS variables with the `!default` keyword. For example of this look at the `Button` component style within `gather-content-ui`.

bagwaa:
I've implemented the feedback @kyleharper <3

lib/StatusIndicator/styles.scss
```diff
@@ -0,0 +1,39 @@
+/* ==========================================================================
+   Config
+   ========================================================================== */
+
+$size-status-indicator: 1em !default;
```
kyleharper: Such a minor thing but in other component styles we switch the name of the component with what it does 😂  e.g. `status-indicator-size` etc...

lib/StatusIndicator/styles.scss
```diff
@@ -0,0 +1,39 @@
...
+  display: inline-block;
+  vertical-align: middle;
+  color: $color-status-indicator-label--text;
+}
+
+.status-indicator__label {
+  margin-right: .25rem;
+}
+
+.status-indicator__children .dateTime {
```
kyleharper: I recommend we keep this flat and also inline with BEM so something like `status-indicator__date` would be good.

Failing that if you're trying to make the date look like a link (underline etc...) then I'd recommend using an anchor tag in the child markup and remove this styling all together.

petemcfarlane: Aye, I don't think we should add the styling of the .dateTime, as it's not part of the component - it's just what we added as a child in the stroybook as an example

lib/StatusIndicator/styles.scss
petemcfarlane: I've updated this now

petemcfarlane:
@bagwaa I added a test, is that testing enough? I'm not too sure what React tests should cover here... Appreciate all the feedback @kyleharper .

bagwaa:
nice, as for testing, might be worth a quick catch up with @kyleharper on Monday to make sure we are testing in the same way as other components.

PR #65, opened 2017-07-11
Component: Floating Button

magalhini:
A button component which floats, on demand. So not kind of like a boat, but close!


lib/FloatingButton/index.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import commentSVG from './../../assets/icons/comment.svg';
+
+const FloatingButton = (props) => {
+  return (
+    <button
+      style={props.position}
```
kyleharper: I don't think the positioning should be handled in `gather-content-ui`. With the current name (FloatingButton), this makes sense but when we resolve the other pieces of feedback then it will become clearer, the position will feel out more of place.

lib/FloatingButton/index.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import commentSVG from './../../assets/icons/comment.svg';
+
+const FloatingButton = (props) => {
```
magalhini: Interesting, that was my initial temptation too, but I had dismissed it due to the amount of complexity/responsibility the Button component would then need to handle.

I like the approach of something like `ButtonWithIcon`, though, and I might give that a go (as in, refactor) since there's already at least one more button (the + sign) that follows a similar pattern. Could be useful to have these as part of the same component sp they look and behave the same.

As part of the People Presence work, which involves that button, I'll try and get this refactored in a way that can handle both scenarios 👍

magalhini:
@kyleharper Let me know what you think mate 👍

PR #66, opened 2017-07-11
Fix all linting issues

magalhini:
These were growing like gangbusters in the repo. Getting it ready for CI integration.


lib/AvatarGroup/index.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable jsx-a11y/no-static-element-interactions */
```
kyleharper: I agree with this linting error. `div` should be a button, please fix 👍

lib/BoundaryClickWatcher/index.js
```diff
@@ -1,10 +1,12 @@
+/* eslint-disable jsx-a11y/no-static-element-interactions */
```
kyleharper: In this instance I can live with it a click handler on a div. Otherwise what's the alternative?

magalhini: Actually both these cases are easy to handle. We're running very old versions of the linters and apparently tons of discussions happened around this since then; both cases are fixable by adding `aria` roles when updating to the newer versions.

However... newer package versions also add more rules, so I now have again 100+ errors to fix 😂 

I think it's worth the hassle, though, so I'll update the packages and fix those.

PR #69, opened 2017-07-15
Fix AvatarGroup issues

magalhini:
Implementing the People component has shown some issues with regards to the AvatarGroup component, namely, its inability to display a single Avatar (had only tested with multiple). This fixes all the issues found during the work.

It also supports, by default, an array of colours to be passed so it can colour the Avatars dynamically.


lib/Avatar/styles.scss
```diff
@@ -35,6 +35,10 @@ $avatar-pillbox-padding: $typo-size-lead/3 $typo-size-lead/2 !default;
   &.is-faded {
     background: $avatar-faded-background !important;
   }
+
+  &.is-assigned {
```
kyleharper: It's ideal not to tie names in a UI library too much with business logic. Lets rename this to something a little more generic/abstract. For example adding a modifier such as `avatar--highlight`.

lib/AvatarGroup/index.js
```diff
@@ -17,28 +17,42 @@ class AvatarGroup extends Component {
   }
 
   render() {
-    const { children, maximum } = this.props;
-    const total = children.length;
-    const display = children.slice(0, maximum);
-    const remaining = children.slice(maximum, total);
+    const { children, maximum, className, colours } = this.props;
+    const total = children && children.length || 0;
```
kyleharper: If there was an issue with rendering a single component I would recommend adding this into the test file (rendering multiple & a single component).

lib/Avatar/index.js
```diff
@@ -1,8 +1,13 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import cx from 'classnames';
 
 const Avatar = (props) => {
-  const styleClass = props.fadedOut ? 'is-faded' : '';
+  const styleClass = cx({
+    'is-faded': props.fadedOut,
+    'is-assigned': props.isAssigned,
```
kyleharper: It's ideal not to tie names in a UI library too much with business logic. Lets rename this to something a little more generic/abstract. For example `avatar--highlight` could map to `props.isHighlighted`.

PR #70, opened 2017-07-18
Add some styling changes to the StatusIndicator

petemcfarlane:
Add flex to align the components, add an optional description, add `current` and `complete` props

### DIFF
--- a/lib/DropdownMenu/styles.scss
+++ b/lib/DropdownMenu/styles.scss
@@ -60,7 +60,7 @@ $dropdown-link-padding: 7px 15px !default;
   cursor: pointer;
   color: #6f7c82;
   font-weight: $typo-weight-roman;
-	font-size: $icon-size;
+	font-size: $typo-size-slight;
   background: none;
   border: 0;
   -webkit-appearance: none;
--- a/lib/DropdownMenu/styles.scss
+++ b/lib/DropdownMenu/styles.scss
@@ -60,7 +60,7 @@ $dropdown-link-padding: 7px 15px !default;
   cursor: pointer;
   color: #6f7c82;
   font-weight: $typo-weight-roman;
-	font-size: $typo-size-slight;
+  font-size: $typo-size-slight;
   background: none;
   border: 0;
   -webkit-appearance: none;
--- a/lib/StatusIndicator/index.js
+++ b/lib/StatusIndicator/index.js
@@ -1,15 +1,30 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-const StatusIndicator = props =>
-  (<span className="status-indicator">
-    <span
-      className="status-indicator__circle"
-      style={{ backgroundColor: props.color }}
-    />
-    <span className="status-indicator__label">{ props.label }</span>
-    <span className="status-indicator__children">{ props.children }</span>
-  </span>);
+const StatusIndicator = (props) => {
+  const classNames = ['status-indicator'];
+  let description;
+  if (props.completed) {
+    classNames.push('status-indicator--completed');
+  }
+  if (props.current) {
+    classNames.push('status-indicator--current');
+    description = <p className="status-indicator__description">{ props.description }</p>;
+  }
+  return (
+    <div className={classNames.join(' ')}>
+      <div className="status-indicator--flex">
+        <span
+          className="status-indicator__circle"
+          style={{ backgroundColor: props.color }}
+        />
+        <span className="status-indicator__label">{ props.label }</span>
+        <span className="status-indicator__children">{ props.children }</span>
+      </div>
+      { description }
+    </div>
+  );
+};
 
 StatusIndicator.propTypes = {
   color: PropTypes.string,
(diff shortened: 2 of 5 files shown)


lib/StatusIndicator/index.js
```diff
@@ -1,15 +1,30 @@
...
-  (<span className="status-indicator">
-    <span
-      className="status-indicator__circle"
-      style={{ backgroundColor: props.color }}
-    />
-    <span className="status-indicator__label">{ props.label }</span>
-    <span className="status-indicator__children">{ props.children }</span>
-  </span>);
+const StatusIndicator = (props) => {
+  const classNames = ['status-indicator'];
```
kyleharper: Whenever className conditional log gets a little much I recommend we use the `classnames` package which is already in this library.

It turns that conditional logic into the following;

```
const classNames = cx({
  'status-indicator--completed': props.completed,
  'status-indicator--current': props.current,
});
```

There are examples of this in the library (search for `cx`) 😄

lib/StatusIndicator/index.js
```diff
@@ -1,15 +1,30 @@
...
+  let description;
+  if (props.completed) {
+    classNames.push('status-indicator--completed');
+  }
+  if (props.current) {
+    classNames.push('status-indicator--current');
+    description = <p className="status-indicator__description">{ props.description }</p>;
```
kyleharper: Conditionals in JSX can be handled as such;

```
{ props.current &&
  <p className="status-indicator__description">{ props.description }</p>
}
```

This can then be moved down to where `{ description }` is. With the feedback around `classNames` I think this approach would be nicer 👍

lib/StatusIndicator/styles.scss
```diff
@@ -2,30 +2,55 @@
...
 
 /* ==========================================================================
    Styles
    ========================================================================== */
 
 .status-indicator {
-  display: inline-block;
-  vertical-align: middle;
   color: $status-indicator-label-color--text;
+  .due-date {
```
kyleharper: Most styles should use the BEM naming format for classes. For example if this is passed as the child component then maybe something like `status-indicator__content`. 

If you feel like the new description suits this better then maybe switch it round and have the description as the child component and the date as a prop on the component.

What do you think?

lib/StatusIndicator/styles.scss
```diff
@@ -2,30 +2,55 @@
...
+    color: #8492a6;
+    font-size: 12px;
```
kyleharper: Always try to use font-size variables from `settings/_typography.scss` when defining a font size. E.g. if you want this to be smaller then `$typo-size-slight` or `$typo-size-small` will suit your needs.

kyleharper: This is feedback for every `font-size` and also `font-weight` set in this file.

lib/StatusIndicator/styles.scss
```diff
@@ -2,30 +2,55 @@
...
 /* ==========================================================================
    Styles
    ========================================================================== */
 
 .status-indicator {
-  display: inline-block;
-  vertical-align: middle;
   color: $status-indicator-label-color--text;
+  .due-date {
+    color: #8492a6;
```
kyleharper: Always try to use colour variables from `settings/_colours.scss`.

This can be tricky at times but every single colour used should be derived from an abstracted variable defined in this settings file. Failing to do so will lead to multiple colours being added in a adhoc manner. This has already started to happen before so let's keep this in check.

If a colour variable doesn't exist then have a chat with the designer and come up with abstract names for them. Here are some examples;

- a yellow use for the workflow date: `$color-background-yellow` might be a good one to add.
- a new blue colour that is slightly off from the existing brand blue: use the existing `$color-background-blue` variable and speak with the team about this as we don't want to add multiple variations of a single colour without purpose.

**What happens when that new tint has a purpose?**

- create a new variable and add the purpose in the name e.g. a lighter tint of blue to help support interactions without being as powerful as the default brand blue, add a  `$color-background-blue-light` variable.

lib/StatusIndicator/styles.scss
```diff
@@ -2,30 +2,55 @@
...
+  align-items: center;
 }
 
 .status-indicator__circle {
-  display: inline-block;
-  vertical-align: middle;
   border-radius: 50%;
-  width: $status-indicator-size;
-  height: $status-indicator-size;
+  width: 12px;
```
kyleharper: Is there a reason why we removed the `$status-indicator-size` variable? That can still be used for this 👍

lib/StatusIndicator/styles.scss
kyleharper: This is feedback for every colour defined in this file.

lib/StatusIndicator/styles.scss
petemcfarlane: I think in this case going with `.status-indicator__children` might be best, the `.due-date` was coming from the storybook example, but I'll remove that there 👍

petemcfarlane:
Thanks @kyleharper I've made some changes following your suggestions. Still waiting on @nhqe for the feedback about text colours being too light/dark, but I didn't want that to block this work so we can tweak that in the future if needs be.

lib/DropdownMenu/styles.scss
```diff
@@ -60,7 +60,7 @@ $dropdown-link-padding: 7px 15px !default;
   cursor: pointer;
   color: #6f7c82;
   font-weight: $typo-weight-roman;
-	font-size: $icon-size;
+	font-size: $typo-size-slight;
```
kyleharper: Looks like we have some strange indent/spacing issue going off in this file. Can we fix that?

PR #77, opened 2017-07-25
Add black and white offline property for Avatar. Minor styling tweaks

magalhini:
This adds a few changes as discussed for the Avatar component, although not all of them are done.

- Black and white picture given an `offline` property (previously, `fadedOut`)
- Fall back to initials if the `url` property is Falsy
- Remove background if the image is shown (as not to show a leaking background colour)


petemcfarlane:
Awesome, thanks @magalhini ! 💥

PR #81, opened 2017-08-08
Avatar component updates

nhqe:
This PR adds finesse to the current presence component, and refactors Storybook with revamped stories.

PR #84, opened 2017-08-11
Add right margin to avatar group so other components don't overlap

petemcfarlane:


lib/AvatarGroup/styles.scss
```diff
@@ -0,0 +1,3 @@
+.avatar-group-list {
+  margin-right: 1em;
```
kyleharper: Ideal we'd use a spacing setting e.g. `$layout-spacing-base` but merge away as this will be cleaned up in another PR around Avatars.

petemcfarlane: The reason I used `1em` is because that is the value that we loose here https://github.com/gathercontent/gather-ui/blob/63a5085884ac582479dc048b0189033639108c3c/lib/Avatar/styles.scss#L93-L99
when the avatars are in a group - this doesn't add any extra margin around the group, it just resets the margin that is lost, if that makes sense 😄

kyleharper: Bah, something else that'll be resolved the #81 so 👍  Cheers dude ✋

PR #85, opened 2017-08-21
Added new icons and cleaned up the stories

kyleharper:
We have a couple of new icons defined in the Icon component;

- menu
- loader

I've also cleaned up how we handle the interactive and active states for icons as there are times where we want to use `fill` rather than `stroke`.


kyleharper:
[[Editor] Show active state of ItemList button](https://trello.com/c/rVBAXiPt/2390-editor-show-active-state-of-itemlist-button)

petemcfarlane:
LGTM @kyleharper! We can still port the back button, the menu drop down (three vertical dots) and probably others... but I can do them in some other PR 👍

PR #92, opened 2017-08-25
Add Toolbar component and icons

petemcfarlane:
This moves the Toolbar used by Quill from the App to this UI library. Most of the styling is applied by Quill so there's not a lot to see here.


lib/Toolbar/index.js
```diff
@@ -0,0 +1,50 @@
...
+import icons from './icons';
+
+class Toolbar extends Component {
+  constructor(props) {
+    super(props);
+    this.state = { showHeaders: false };
+  }
+
+  render() {
+    return (<div id="quill-toolbar">
```
kyleharper: `id` should use a prop.

kyleharper: Just a minor piece of feedback to but check linting as I would have thought that the div would need to be on a new line?

lib/Toolbar/index.js
```diff
@@ -0,0 +1,50 @@
...
+      <div className="ql-formats">
+        <button
+          className="ql-show-headers"
+          onClick={() => this.setState({ showHeaders: !this.state.showHeaders })}
+        >
+          { icons.aa() }
+        </button>
+        <div className={`ql-select-headers ${this.state.showHeaders ? 'show-headers' : ''}`}>
```
kyleharper: What do you think to `showHeadings` over `showHeaders`?

lib/Toolbar/index.js
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  render() {
+    return (<div id="quill-toolbar">
+      <div className="ql-formats">
+        <button
+          className="ql-show-headers"
+          onClick={() => this.setState({ showHeaders: !this.state.showHeaders })}
+        >
+          { icons.aa() }
```
kyleharper: Just curious, what does `aa` mean?

lib/Toolbar/index.js
```diff
@@ -0,0 +1,50 @@
...
+        </button>
+        <div className={`ql-select-headers ${this.state.showHeaders ? 'show-headers' : ''}`}>
+          <button className="ql-paragraph" />
+          <button className="ql-header" value="1" />
+          <button className="ql-header" value="2" />
+          <button className="ql-header" value="3" />
```
kyleharper: Just wondering if we should extract the toolbar "actions" as children rather than having them explicitly set in the toolbar component.

Having the UI component bound to `ql` means we store references to a 3rd party that doesn't exist in `gather-ui`. E.g. imagine seeing this component for the first time...what does `ql` stand for?

So when consuming you would;

```
<Toolbar id="some-id">
  <div className="ql-formats">
    <button className="ql-header" value="1" />
    <button className="ql-header" value="2" />
    <button className="ql-header" value="3" />
    ...etc
  </div>
</Toolbar>
```

If you need to then apply BEM classes to the children you could look into https://facebook.github.io/react/docs/react-api.html#react.children to map over and add additional markup with the BEM classes you need.

lib/Toolbar/index.js
petemcfarlane: The aa is for this first icon, for changing between p, H1, .. H6 screen shot 2017-08-25 at 17 16 10

lib/Toolbar/index.js
petemcfarlane: The trouble with separating the logic here is it is almost all business/action based, there's not much UI in it. Arguably if we just had `<Toolbar id="some-id">...children components...</Toolbar>` as you suggest, well why do we have a specific new component and why don't we just use a `<div>`? Or maybe we just have a Toolbar component back in the App as it was originally?

lib/Toolbar/index.js
petemcfarlane: Or better yet, I prefer `chooseHeadings`, what do you think?

petemcfarlane:
[[Editor] Floating toolbar component](https://trello.com/c/We5hWRsX/2309-editor-floating-toolbar-component)

lib/Toolbar/index.js
kyleharper: Well, the styling side of things is well worth keeping detached from the consuming app. The UI story would then demo the use of all the styles. 

Even the smallest component when handling UI logic or styling should be considered for the UI library, but again its just 2 paths ending at the same destination.

If we use the UI library I suggest we don't add 3rd party markup. Let's keep the concerns separate, otherwise when adding future components what is right or wrong? 😄

kyleharper: There are some more complex examples that we could take for this component, which would make it truly extendable/flexible for the future but I believe they'd be too complex for our needs at this moment 😄

assets/icons/ul.svg
```diff
@@ -0,0 +1,10 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" viewBox="0 0 16 16">
+  <g transform="translate(1 2)">
+    <rect width="11" height="2" x="3" rx=".5"/>
+    <rect width="11" height="2" x="3" y="5" rx=".5"/>
+    <rect width="11" height="2" x="3" y="10" rx=".5"/>
+    <circle cx="1" cy="1" r="1"/>
```
acairns: I think you mean `rx=".525"` here...

PR #93, opened 2017-08-29
Icon component clean up

kyleharper:
We've added some new icons as well as cleaned up the responsibilities a little.

The Icon component was doing a little too much including handling the interactive states when we didn't need it to. The green hover wasn't an initial requirement but instead was a temp style that got through QA.

Because of this we've cleaned up the Icon component altogether and moved the interactive logic into the Button component where it feels more natural to handle interactivity.

Also, sizes are no longer a thing for the Icon component. The size is determined from the SVG we import into the app. Which is exported from design etc... 👍


stories/components/Button.js
```diff
@@ -92,6 +92,17 @@ const button = storiesOf('Components', module)
               hideText
             />
           </Button>
+
+          <Button
+            types={['icon-only', 'contained', 'hover-transform']}
```
petemcfarlane: Make the `contained` and `hover-transform` their own story for consistency

PR #94, opened 2017-08-29
Status indicator bordered

Cyapow:
Have added the modifier to allow a bordered StatusIndicator. I have used a variable for the border color but not for padding or border-radius. Is this something that needs changing? @kyleharper


tests/StatusIndicator/index.js
```diff
@@ -41,4 +41,28 @@ describe('StatusIndicator', () => {
 
     expect(wrapper.find('.status-indicator').hasClass('status-indicator--completed')).to.equal(true);
   });
+
+  it('can be the last status', () => {
+    expect(wrapper.find('.status-indicator').hasClass('status-indicator--last')).to.equal(false);
+
+    wrapper = mount(<StatusIndicator color="#00ff00" label="Review" last />);
```
kyleharper: If we're just changing props, lets use `setProps` from enzyme. If this is the case then lets update all tests in this file.

lib/StatusIndicator/styles.scss
```diff
@@ -56,6 +50,12 @@ $status-indicator-horizontal-margin: .5rem !default;
   }
 }
 
+.status-indicator--bordered {
+  border: 2px solid $color-border-base;
+  border-radius: .25rem;
+  padding: 0.7rem;
```
kyleharper: Is there any way we can use a standardised value here? Spacing, font sizes and colours all should use settings where possible.

If you NEED to override consistent global settings then this can be done by adding this style to the components config area (the area at the top of this file with the section comment "config") and then overriding it in the consuming app.

PR #95, opened 2017-08-29
Status indicator bordered

Cyapow:


lib/StatusIndicator/styles.scss
```diff
@@ -5,6 +5,8 @@ $status-indicator-label-color--text: rgb(111, 121, 129) !default;
 $status-indicator-circle-size: $typo-size-small !default;
 $status-indicator-circle-size--current: $typo-size-base !default;
 $status-indicator-horizontal-margin: .5rem !default;
+$status-indicator-bordered-border: 2px solid #eaeaea !default;
```
kyleharper: Could this use a setting instead? e.g. `$color-border-base`

Cyapow: I did use color-border-base but it looked odd when it was on the actual webapp. I mean I'm happy to use that variable and override it in the app but is it worth doing that? What u think @kyleharper

kyleharper: Hmm, maybe use another colour if that one doesn't work but lets use an existing colour from our pallet 😄

PR #98, opened 2017-08-30
List Component Design Fixes

kyleharper:


lib/List/index.js
```diff
@@ -1,27 +1,35 @@
...
+  });
 
-    <div className="list__body">
-      { React.Children.map(props.children, child => (
-        <div className="list__row">
-          { React.cloneElement(child, {}) }
-        </div>
-      )) }
+  return (
+    <div className={`list ${listClasses}`}>
```
petemcfarlane: Maybe this would be a good opportunity to use `cx(list, {'list--bordered-right': props.borderedRight})` again?

PR #99, opened 2017-08-31
Add an other radio button option, for user input

petemcfarlane:


lib/Form/RadioButton/Other.js
```diff
@@ -0,0 +1,49 @@
...
+    return (
+      <span>
+        <Input name={name} id={id} checked={checked} onChange={() => onChangeHandler(this.state.otherOptionValue)} value={value} />
+        <Label label={label} id={id} />
+        { checked &&
+          <input
+            type="text"
+            id={`other-value-${id}`}
+            value={this.state.otherOptionValue}
+            onChange={(e) => {
```
kyleharper: Extract the `onChange` value into a separate method. Makes the JSX cleaner that way 👍

lib/Form/RadioButton/Other.js
```diff
@@ -0,0 +1,49 @@
...
+  constructor(props) {
+    super(props);
+    this.state = { otherOptionValue: props.value };
+  }
+
+  render() {
+    const { name, value, id, checked, label, onChangeHandler } = this.props;
+    return (
+      <span>
+        <Input name={name} id={id} checked={checked} onChange={() => onChangeHandler(this.state.otherOptionValue)} value={value} />
```
kyleharper: There's a ton of props here. Let's line break each one for readability 👍  (the same goes for Label too).

petemcfarlane:
Thanks @kyleharper, I've updated based on your feedback. Ok to merge now?

PR #101, opened 2017-09-05
Popover wrapper changes

Cyapow:


styles/overrides/_popover.scss
```diff
@@ -1,4 +1,6 @@
 .popover.bottom {
```
kyleharper: Remind me, is the `.bottom` class just for the bottom of the popover (e.g. the content) or is is when the popover is positioned at the bottom of the element you're hovering/clicking on?

lib/DropdownMenu/styles.scss
```diff
@@ -57,7 +57,7 @@ $dropdown-menu-link-padding: 7px 15px !default;
 }
 
 /** Custom **/
-.dropdown__link {
+.dropdown__link, .dropdown-menu a.dropdown__link {
```
kyleharper: Do we need this declaration to override some styling?

Cyapow: If we don't the styles get overridden by .dropdown-menu > li > a. Im open to other ways around it though

styles/overrides/_popover.scss
Cyapow: Hmm probably shouldn't use .bottom as its very specific. Should I add an extra class you think and use that?

PR #102, opened 2017-09-06
Feature choice fields

petemcfarlane:
This PR pretty much rewrites the RadioButtonGroup - which was only being used in once place in our App (part of the import flow that is as it stands unaccessible, and should be straightforward to refactor if needs be).

The idea of the RadioButtonGroup and CheckboxGroup is you pass them a list of `choices` (a JSON array) as a prop, and an `onChangeHandler` prop, which will receive the reduced list of choices including only the selected choices.

This will eliminate the need for some complex code in the App pertaining to adding/removing items from an array. It uses the components state to manage the currently selected items in the list. The RadioButtonList also has support for an optional `Other` choice field, with a user supplied input.


lib/Form/Checkbox/Group.js
```diff
@@ -0,0 +1,59 @@
...
+    const choice = this.state.choices.find(c => c.id === e.target.id);
+
+    const selected = this.choiceIsAlreadySelected(choice) ? this.removeChoice(choice) : this.addChoice(choice);
+    this.setState({ selected });
+
+    this.props.onChangeHandler(selected);
+  }
+
+  choiceIsAlreadySelected(choice) {
+    return this.state.selected.findIndex(c => c === choice) >= 0;
```
kyleharper: `findIndex` isn't support in IE: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex

Maybe use `filter` instead and look for the `id` within the selected objects?

lib/Form/Checkbox/Group.js
```diff
@@ -0,0 +1,59 @@
...
+    this.onChangeHandler = this.onChangeHandler.bind(this);
+    this.choiceIsAlreadySelected = this.choiceIsAlreadySelected.bind(this);
+    this.removeChoice = this.removeChoice.bind(this);
+    this.addChoice = this.addChoice.bind(this);
+  }
+
+  onChangeHandler(e) {
+    const choice = this.state.choices.find(c => c.id === e.target.id);
+
+    const selected = this.choiceIsAlreadySelected(choice) ? this.removeChoice(choice) : this.addChoice(choice);
```
kyleharper: When we get `choice` don't we already get the checked/selected state? So...;

```
const choice = this.state.choices.filter(c => c.id === e.target.id);

if (choice.selected) ?
  this.removeChoice(choice) :
  this.addChoice(choice);
```

It's also worth noting that `find` isn't supported in IE. So I have changed that in my explain above.

lib/Form/Checkbox/Group.js
```diff
@@ -0,0 +1,59 @@
...
+
+    this.props.onChangeHandler(selected);
+  }
+
+  choiceIsAlreadySelected(choice) {
+    return this.state.selected.findIndex(c => c === choice) >= 0;
+  }
+
+  addChoice(choice) {
+    return this.state.selected.concat(choice);
```
kyleharper: State should only be manipulated via setState, here's some docs to explain why;

- https://facebook.github.io/react/docs/react-component.html#setstate
- https://stackoverflow.com/questions/37755997/why-cant-i-directly-modify-a-components-state-really

kyleharper: With this in mind you could then pass `this.props.onChangeHandler(selected);` as a callback to the `setState` method chaining them nicely together. e.g.

```
addChoice(choice) {
  return this.setState({
    selected: this.state.selected.concat({
      ...choice,
      selected: true
    })
  }, this.onChangeHandler);
}

onChangeHandler() {
  this.props.onChangeHandler(this.state.selected);
}
```

kyleharper: The idea is shared with `removeChoice`.

lib/Form/Checkbox/Group.js
```diff
@@ -0,0 +1,59 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import Checkbox from './';
+
+export default class Group extends Component {
+  static propTypes = {
+    choices: PropTypes.arrayOf(PropTypes.shape({
```
kyleharper: You can get away with just doing;

```
choices: PropTypes.arrayOf(PropTypes.shape()),
```

Unless you're providing details of the object structure.

lib/Form/Checkbox/Group.js
kyleharper: This is linked with my comment about `addChoice`. It'll show how we can edit the selected value of the choice within state and then use that in here.

lib/Form/styles.scss
```diff
@@ -9,13 +9,26 @@ $form-checkbox-label-size: 1rem !default;
...
+  cursor: pointer;
+  margin-left: $form-checkbox-margin;
+  font-weight: $form-checkbox-font-weight;
+  font-size: $form-checkbox-label-size;
+  display: block;
 }
 
-.form__choice-element-wrapper {
-  margin: $form-checkbox-label-size/2;
+.form-checkbox__label__subtitle {
```
kyleharper: `form-checkbox__subtitle` would be fine 👍

lib/Form/Checkbox/Group.js
petemcfarlane: This isn't quite true, the `state.choices` will contain any choices that were initially checked when we first created the component, but if any choices are then checked/unchecked, we don't update the checked property on the choices, but add the whole choice to the `state.selected` array. Do you think we should worry about this inconsistency?

lib/Form/Checkbox/Group.js
petemcfarlane: Thanks, have done now :+1:

lib/Form/Checkbox/Group.js
kyleharper: We'll see when we give it a test. I'm not sure if it'll be an issue or not.

lib/Form/Checkbox/Group.js
```diff
@@ -0,0 +1,74 @@
...
+
+  componentWillReceiveProps(nextProps) {
+    if (nextProps.choices !== this.props.choices) {
+      const selected = nextProps.choices.filter(c => c.checked);
+      this.setState({ choices: nextProps.choices, selected });
+    }
+  }
+
+  onChangeHandler(e) {
+    const choice = this.state.choices.filter(c => c.id === e.target.id).shift();
```
kyleharper: Are we using `shift` to make sure we get one element only?

petemcfarlane: yes, is there a better way to do this?

kyleharper: I was just wondering if we should even worry about that? If it fails because it has more than 1 element I think we should see that error happen right?

What happens if this returns more than one choice and because we never allow that to error we get some corrupted experience that is very hard to debug?

PR #108, opened 2017-09-14
Feature radio button other option

petemcfarlane:
This is to match @nhqe designs for radio button other options


lib/Form/RadioButton/Other.js
```diff
@@ -22,6 +22,29 @@ class Other extends Component {
   render() {
     const { name, value, id, checked, label, onChangeHandler, onTextChangeHandler } = this.props;
```
kyleharper: I think in the future (and maybe for this PR) we begin spreading certain attributes/props rather than explicitly adding them to the component. This means any future attributes we add will always be used on the input e.g. `disabled` right now would require code changes for this to be reflected in the browser.

How would this look?

```
const { label, onChangeHandler, onTextChangeHandler, ...rest } = this.props; // rest = { name, value, id, checked }

<Input
  ...rest
  onChangeHandler={onChangeHandler}
/>
<input
  ...this one doesn't share the same props so set this as you are
/>
```
What do you think?

lib/Form/RadioButton/Other.js
```diff
@@ -30,20 +53,14 @@ class Other extends Component {
           checked={checked}
           value={value}
           onChangeHandler={onChangeHandler}
+          className="form-radiobutton--invisible"
```
kyleharper: I think there is a helper class of `is-hidden` or `hide` to do this 👍

lib/Form/RadioButton/Other.js
```diff
@@ -22,6 +22,29 @@ class Other extends Component {
   render() {
     const { name, value, id, checked, label, onChangeHandler, onTextChangeHandler } = this.props;
 
+    if (checked) {
```
kyleharper: I think this might be missing a test...and so is the contents of the conditional.

lib/Form/styles.scss
```diff
@@ -31,4 +35,12 @@ $form-checkbox-label-size: 1rem !default;
 
 .form-radiobutton__other {
   margin-left: $form-checkbox-margin;
+  border: none;
+  outline: none;
+  font-size: 1rem;
```
kyleharper: Please use a setting for the font-size 👍

stories/base/Base.js
```diff
@@ -12,7 +12,7 @@ const neutralColours = [
   {
     rgbValue: 'rgb(132, 146, 166)',
     hexValue: '#8492A6',
-    name: '$neutral',
+    name: '$neutral-base',
```
kyleharper: Good find 👍

lib/Form/RadioButton/Other.js
```diff
@@ -22,6 +22,29 @@ class Other extends Component {
...
+            value={value}
+            onChangeHandler={onChangeHandler}
+          />
+          <input
+            className="form-radiobutton__other"
+            type="text"
+            id={`other-value-${id}`}
+            value={value}
+            onChange={onTextChangeHandler}
+            autoFocus
```
kyleharper: Using this isn't a good idea for various reasons. One being that when loading the page this will auto focus for each other option checked.

What it is you wanted to achieve?

PR #119, opened 2017-09-27
Add prettier

kyleharper:
This PR adds the `lint:fix` command to automatically fix linting errors using prettier. It also updates a ton of eslint packages so we're up to date 😄


kyleharper:
If the updated linting rules remove them then I guess not. That'll be a eslint thing though I think. We could double check.

PR #121, opened 2017-10-05
File card changes

Cyapow:


tests/FileCard/index.js
```diff
@@ -25,8 +25,11 @@ describe('FileCard', () => {
 
   it('renders an element with the correct background-image', () => {
     expect(
-      wrapper.find('.file-card__thumbnail').props().style.backgroundImage
-    ).to.equal(`url(${props.previewSrc})`);
+      wrapper
+        .find('img')
+        .props()
+        .src.indexOf(props.previewSrc)
```
Cyapow: @kyleharper used indexOf here to check the file URL as at the beginning of the attribute, is this the way to go? I am adding ?rand=Math.random to the URL so I can't be sure what the URL will be

Cyapow:
[[Editor] Handle upload of file](https://trello.com/c/1Bh8jr6U/2432-editor-handle-upload-of-file)

lib/FileCard/index.js
```diff
@@ -36,19 +96,9 @@ class FileCard extends Component {
...
           <div className="file-card__label">{label}</div>
 
-          {previewSrc && (
-            <div
-              title={label}
-              style={{ backgroundImage: `url(${previewSrc})` }}
-              className="file-card__thumbnail"
-            />
-          )}
+          {previewSrc && this.backgroundImage()}
```
kyleharper: You may want to change from `previewSrc` to `this.state.imageHasLoaded`. That way your render logic is all around when the image is ready or not.

PR #122, opened 2017-10-09
Update Avatars

kyleharper:
Inline with some visual feedback, here are some updates to the Avatar component. This also includes some refactors that I think work pretty well at making the component easier to use.


lib/Avatar/index.js
```diff
@@ -5,66 +5,55 @@ import cx from 'classnames';
 /**
  * @usage
  *
- * <Avatar initials="MR" color="green" url="img/url" onlyInitials isHighlighted offline />
+ * <Avatar name="Seymour Butts" color="green" url="img/url" offline />
```
Cyapow: You been watching heavy weights? lol

kyleharper: This was just a name used in the stories previously. I might change it lol

lib/Avatar/index.js
```diff
@@ -5,66 +5,55 @@ import cx from 'classnames';
...
  */
-const Avatar = props => {
-  const offlineBg = Avatar.defaultProps.colour;
-  const displayImage = props.url && !props.onlyInitials && !props.offline;
-  const displayGreyedImage = props.url && props.offline;
-  const backgroundColor = props.offline ? offlineBg : props.colour;
+const getInitialsFromName = name => {
+  const names = name.split(' ');
+  return names.reduce((acc, value) => `${acc}${value.split('')[0]}`, '');
```
petemcfarlane: You can just do `value[0]` instead of `value.split('')[[0]` here, to get the first character from a string.

Also I think we pass the initials with the API anyway, and the logic Angus decided upon was we would just take the first letter from the first and last names, so we should always have two characters. What happens with the above reducer if the person has multiple names, represented by spaces? You could potentially get three or more characters.

kyleharper:
@petemcfarlane @Cyapow thanks for the feedback. All has been applied 👍

petemcfarlane:
👀 looks good
M E R G E

PR #123, opened 2017-10-10
Added additional avatar feedback

kyleharper:


petemcfarlane:
Not sure what the additional feedback was, but it works and looks ok to me

PR #130, opened 2017-10-25
Add -blue-darkest to colour pallet

petemcfarlane:
Adding a colour to the base storybook and to the settings.sass, going to be used by the new editor floating toolbar.


petemcfarlane:
Looks like my IDE has 'autoformat on save' on, which has changed a few other bits unintentionally - do you think this is ok? Do we have an eslint ruleset for this repo??

petemcfarlane:
[[Editor] Fix hyperlink within rich text fields](https://trello.com/c/WFPpPKEL/2528-editor-fix-hyperlink-within-rich-text-fields)

kyleharper:
@petemcfarlane No worries about the auto formatting. If linting was breaking the CI process should catch it 👍

petemcfarlane:
Thanks @kyleharper let me know when we can release a version so I can bump it in the App

PR #132, opened 2017-10-26
New conversation stories & tests

AmeeMorris:


lib/Conversation/Comment.js
```diff
@@ -14,19 +14,15 @@ class Comment extends Component {
...
   };
 
   static defaultProps = {
     user: {},
     focusFallback() {},
-    actions: {
-      editComment() {},
-      removeComment() {}
-    }
+    editComment() {},
```
kyleharper: Remove as these are required 👍

lib/Conversation/index.js
```diff
@@ -4,18 +4,16 @@ import pluralize from 'pluralize';
...
-    actions: PropTypes.shape({
-      resolveConversation: PropTypes.func.isRequired,
-      addComment: PropTypes.func.isRequired,
-      activateConversation: PropTypes.func.isRequired,
-      deactivateConversation: PropTypes.func.isRequired
-    }).isRequired,
+    resolveConversation: PropTypes.func,
+    onCancel: PropTypes.func,
+    addComment: PropTypes.func.isRequired,
+    activateConversation: PropTypes.func,
```
kyleharper: Lets abstract out `activateConversation` and `deactivateConversation` and treat BoundaryClickWatcher as another UI component that can be used to compose the interactions we desire.

lib/Conversation/index.js
```diff
@@ -37,85 +40,85 @@ class Conversation extends Component {
...
-            conversationId={id}
-            comments={comments}
-            focusFallback={this.focusFallback}
-          />
-
-          {!isActive &&
-            comments.length > 1 && (
-              <div className="conversation__reply-count">
+      <div className={`conversation ${isActiveClass}`}>
+        {typeof this.props.resolveConversation !== 'undefined' &&
```
kyleharper: Try just `!this.props.resolveConversation` 👍

tests/Conversation/Conversation.js
```diff
@@ -160,4 +150,9 @@ describe('Conversation', () => {
     wrapper.setProps({ userCanComment: false });
     expect(wrapper.find(CommentForm)).to.have.length(0);
   });
+
+  it('does not render a resolveConversation button', () => {
```
kyleharper: We also don't render a convoList 😄

lib/Conversation/index.js
```diff
@@ -24,98 +20,89 @@ class Conversation extends Component {
...
-            type="text"
-            className="conversation__focus-fallback"
-            ref={focusFallback => {
-              this.focusFallback = focusFallback;
-            }}
-            aria-hidden="true"
-          />
-        </div>
-      </BoundaryClickWatcher>
+        {!isActive &&
```
kyleharper: I wonder if we should change the naming of `isActive` to something like `showComments`. The original naming made sense as we were setting the active state internally but now we're just showing the comments or not. What do you think?

AmeeMorris:
Hey @kyleharper I've also added in the BoundaryClickWatcher refactoring to this PR

lib/BoundaryClickWatcher/index.js
```diff
@@ -67,10 +75,13 @@ class BoundaryClickWatcher extends Component {
...
         }}
-        onClick={this.handleInitialClick}
-        onKeyPress={this.handleInitialClick}
+        onClick={this.handleInnerClick}
+        onKeyPress={this.handleInnerClick}
       >
-        {this.props.children}
+        {typeof this.props.children === 'function'
+          ? this.props.children(this.state.boundaryIsActive)
+          : this.props.children
```
kyleharper: Is this so we can use this component as a means to trigger `initialClickHandler` & `outsideClickHandler` without passing the active state to the child?

tests/BoundaryClickWatcher/BoundaryClickWatcher.js
```diff
@@ -0,0 +1,35 @@
...
+    );
+  });
+
+  it('boundaryIsActive is true on click', () => {
+    expect(wrapper.state('boundaryIsActive')).to.equal(false);
+    wrapper.simulate('click');
+    expect(wrapper.state('boundaryIsActive')).to.equal(true);
+  });
+
+  it('passes boundaryIsActive to child', () => {
```
kyleharper: Did you want to test when a child is passed that isn't a function? If you can't imagine a use case for this then maybe we should expect a function to be passed as the child 👍

lib/BoundaryClickWatcher/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+##Usage
+
+BoundaryClickWatcher has a boolean (boundaryIsActive) which is set to true when it detects an inner click and back to false when it detects a click outside of the boundary. The value of this can be passed onto the child component like so:
+
+```
+<BoundaryClickWatcher>
+  {(boundaryIsActive) => {
+    return (
+      <My Component
```
kyleharper: Change to `<MyComponent` 😄

lib/BoundaryClickWatcher/README.md
```diff
@@ -0,0 +1,40 @@
+#BoundaryClickWatcher
+Wrap a component in BoundaryClickWatcher to detect clicks on our outside of the component. See an example of this in the Conversation story.
+
+##Usage
+
+BoundaryClickWatcher has a boolean (boundaryIsActive) which is set to true when it detects an inner click and back to false when it detects a click outside of the boundary. The value of this can be passed onto the child component like so:
```
kyleharper: I'd say its worth mentioning that it uses the render prop pattern and keep this simple. Something like: "`BoundaryClickWatcher` uses the render callback pattern and returns a boolean."

I think the usage shows just how to use it 😃

lib/BoundaryClickWatcher/README.md
```diff
@@ -0,0 +1,40 @@
...
+    return (
+      <My Component
+        boolProp={boundaryIsActive}
+      />
+    )
+  }}
+</BoundaryClickWatcher>
+```
+
+*(Optional)* It has two props `insideClickHandler` and `outsideClickHandler` you can pass in actions for added custom behaviour on inner or outer click.
```
kyleharper: I wonder if we should make this more concise?

Something like the following section in a different library: https://github.com/igorprado/react-notification-system#methods

e.g.

--------

## Props

### `insideClickHandler`

Executes when the user clicks inside the boundary.

### ```outsideClickHandler```

Executes when the user clicks outside the boundary.

--------

Or maybe a table like https://github.com/igorprado/react-notification-system/blob/master/README.md#creating-a-notification

What do you think?

lib/Conversation/README.md
```diff
@@ -0,0 +1,40 @@
+#Conversation
+A collection of components used to render conversations
+
+##Usage
+
+####Required props
```
kyleharper: I think a table would work better for props, what do you think?

PR #134, opened 2017-10-31
Feature comments styling

AmeeMorris:


lib/Conversation/CommentForm/CommentFormActions.js
```diff
@@ -1,31 +1,28 @@
...
-    </Button>
-
     <span className="comment-form__actions comment-form__actions--submit">
       <Button types={['danger', 'size-sm']} clickHandler={props.onCancel}>
         Cancel
       </Button>
-      <Button types={['secondary', 'size-sm']} clickHandler={() => {}} isSubmit>
+      <Button
+        types={['secondary', 'size-sm']}
+        disabled={props.hasValue ? false : true}
```
kyleharper: `!props.hasValue` would work just the same as this ternary operator. What do you think?

lib/Conversation/CommentForm/styles.scss
```diff
@@ -2,6 +2,7 @@
    Config
    ========================================================================== */
 $comment-form-padding: $layout-spacing-base/2;
+$comment-form-input-background-color: #FDFDFD !default;
```
kyleharper: We should using colour settings instead of hardcoding values, can this use one of our settings?

lib/Conversation/styles.scss
```diff
@@ -1,9 +1,16 @@
...
    ========================================================================== */
+$conversation-active-background-color: #ffffff !default;
+$conversation-active-shadow-color: rgba(158, 158, 158, 0.17) !default;
+$conversation-active-shadow-size: 0px 0px 5px 0px !default;
+
+
 $conversation-border-radius: 5px !default;
 $conversation-padding: $layout-spacing-base !default;
 
+$conversation-margin: 10px 0 !default;
```
kyleharper: For spacing (or any context) if a value is derived from a design pattern then we could use the base setting to create our desired output.

In simple terms, spacing is based on a `20px` setting called `$layout-spacing-base`. To get a 10px spacing you can do `$layout-spacing-base/2` and then that value is derived from one of our base settings. Making it more consistent etc...

lib/Conversation/styles.scss
```diff
@@ -1,9 +1,16 @@
 /* ==========================================================================
    Config
    ========================================================================== */
+$conversation-active-background-color: #ffffff !default;
+$conversation-active-shadow-color: rgba(158, 158, 158, 0.17) !default;
```
kyleharper: Is `rgba(158, 158, 158, ...)` based off an existing setting? If so, maybe `rgba($setting-name, ...)` 👍

lib/Conversation/styles.scss
```diff
@@ -57,7 +64,7 @@ $conversation-meta-text-color: $color-brand-dark !default;
 }
 
 .conversation__reply-count {
-  padding: 0 0 $conversation-padding $conversation-padding+$avatar-size;
+  padding: 0 0 $conversation-padding $conversation-padding+$avatar-size+10;
```
kyleharper: Look at the "setting" feedback about regarding the hardcoded spacing extension of `10`.

lib/index.js
```diff
@@ -2,6 +2,7 @@ export * as constants from './constants';
 export Avatar from './Avatar';
 export AvatarGroup from './AvatarGroup';
 export AvatarWithPopover from './Avatar/AvatarWithPopover';
+export BoundaryClickWatcher from './BoundaryClickWatcher';
```
kyleharper: 👍

lib/Conversation/README.md
```diff
@@ -11,10 +11,13 @@ A collection of components used to render conversations
...
 | comments            | Array         | `[]`     | No       | An array of comments that are in the conversation.                            |
+| removeComment       | Function      | `() {}`  | No       | Executes when the comment is removed.                                         |
+| editComment         | Function      | `() {}`  | No       | Executes when the comment is edited.                                          |
 | resolveConversation | Function      | `() {}`  | No       | Executes when the add resolve conversation button is clicked.                 |
 | onCancel            | Function      | `() {}`  | No       | Executes when the cancel button is clicked.                                   |
 | userCanComment      | Boolean       | `false`  | No       | Determines whether the user is allowed to comment and if to display the form. |
 | showComments        | Boolean       | `true`   | No       | Determines whether to show the conversation in the expanded format.           |
+| focusOnMount        | Boolean       | `true`   | No       | Determines whether to have the form in focus when the component mounts.       |
```
kyleharper: Nice new prop 👍

lib/Conversation/Comment.js
```diff
@@ -52,11 +53,11 @@ class Comment extends Component {
   }
 
   render() {
-    const { createdAt, createdBy, body, author, user } = this.props;
+    const { createdAt, createdBy, body, author, user, hasReply } = this.props;
     const userCanEdit = createdBy === user.id;
-
+    const hasReplyClass = hasReply ? 'has-reply' : '';
```
kyleharper: I can't see any styling for this. What is the purpose of this state class?

tests/Conversation/CommentForm/CommentForm.js
```diff
@@ -76,6 +76,7 @@ describe('Comment Form', () => {
   });
 
   it('renders CommentFormActions (with correct props)', () => {
```
kyleharper: There is a number of new props that have been added e.g. `onBlur` & `onFocus`. Should we add tests for what these do? 

I believe they set state so maybe it's a good idea?

lib/Conversation/CommentForm/CommentFormActions.js
```diff
@@ -1,31 +1,39 @@
...
+          Cancel
+        </Button>
+        <Button
+          types={['secondary', 'size-sm']}
+          disabled={!props.hasValue}
+          clickHandler={() => {}}
+          isSubmit
+        >
+          {props.submitted && !props.existingConversation ? (
+            <Icon name="loader" />
```
kyleharper: There is an existing component for handling showing progression when submitting. If you check out `ProgressButton` if might already provide you what you need 👍

lib/Conversation/CommentForm/index.js
```diff
@@ -13,6 +13,7 @@ class CommentForm extends Component {
       name: PropTypes.string
     }).isRequired,
     focusOnMount: PropTypes.bool,
+    existingConversation: PropTypes.bool.isRequired,
```
kyleharper: What's the thinking around this prop? What does `existingConversation` mean/do?

lib/Conversation/CommentForm/index.js
```diff
@@ -28,31 +29,43 @@ class CommentForm extends Component {
...
   }
 
   onSubmit(e) {
     e.preventDefault();
     this.props.onSubmit(this.state.inputValue);
-    this.setState({ inputValue: '' });
+    this.props.existingConversation && (
+      this.setState({ inputValue: '' })
+    );
+    this.setState({ submitted: true });
```
kyleharper: Does React combined these 2 setStates together? If not then we're calling setState twice rather than calling it once and maybe causing 2 renders for 1 UI update, is that an issue?

Something like this would combine the 2 👍 

```
const inputValue = this.props.existingConversation ? '' : this.state.inputValue
this.setState({ 
  submitted: true,
  inputValue,
})
```

lib/Conversation/CommentForm/CommentFormInput.js
```diff
@@ -20,14 +23,17 @@ class ContactFormInput extends Component {
   }
 
   render() {
+    const isSubmittedClass = this.props.submitted ? 'is-submitted' : '';
```
kyleharper: `isSubmitted` sounds kind of strange;

- is it submitting? (`is-submitting`)
- has it submitted? (`has-submitted`)

Clearing this up would make it easier to understand what it handles 👍

lib/Conversation/CommentForm/styles.scss
```diff
@@ -42,6 +42,21 @@ $comment-form-padding: $layout-spacing-base/2;
 /* ==========================================================================
    State Styles
    ========================================================================== */
-.has-value .comment-form__actions--submit {
-  display: inline-block;
-}
\ No newline at end of file
+   .new-conversation .is-submitted .button--secondary {
```
kyleharper: Is this styling for a disabled state? I wonder if this should be abstracted away from the submit form component and into the Button component, what do you think?

We could then update the type of button when required.

lib/Conversation/CommentForm/styles.scss
```diff
@@ -42,6 +42,21 @@ $comment-form-padding: $layout-spacing-base/2;
...
+   .is-active .new-conversation .is-submitted {
+     .icon svg {
+       height:$layout-spacing-base;
+     }
+     .icon path {
+       fill: white;
+     }
+   }
+
+   .new-conversation .comment-form__input.is-submitted {
```
kyleharper: Do we need the prefix of `.new-conversation` in the declaration?

I only ask because `.comment-form__input.is-submitted` feels like a nice state class that we only add when a certain scenario occurs, meaning that we shouldn't need additional classes in the declaration right?

It might just be confusing me due to the `is-submitted` class which I mention in some other feedback, maybe renaming that would clear it up.

PR #135, opened 2017-11-03
Design tweaks, focusOnMount Fix

AmeeMorris:


lib/Conversation/index.js
```diff
@@ -59,34 +59,38 @@ class Conversation extends Component {
       'is-active': showComments
     });
     const { id, comments, ...rest } = this.props;
+    const existingConversation =
+      this.props.resolveConversation && this.props.comments.length > 0;
+    const placeholderText = existingConversation
```
kyleharper: I feel like `placeholderText` should just be a prop that is passed in. What if we want to change the text to "Reply to this thread..."?

That way we remove the idea of an "existing" conversation too. Which simplifies the component.

lib/Conversation/CommentForm/CommentFormInput.js
```diff
@@ -7,19 +7,21 @@ class ContactFormInput extends Component {
...
 
   static defaultProps = {
-    focusOnMount: false,
     value: ''
   };
 
   componentDidMount() {
     if (this.props.focusOnMount) {
-      this.input.focus();
+      setTimeout(() => {
```
kyleharper: Why do we need a timeout here?

PR #139, opened 2017-11-08
Top bar component

Cyapow:


stories/components/TopBar.js
```diff
@@ -0,0 +1,25 @@
...
+import TopBar from '../../lib/TopBar';
+import Button from '../../lib/Button';
+
+storiesOf('Components', module).add('TopBar', () => (
+  <div>
+    <StoryItem
+      title="TopBar Wrapper"
+      description="TopBar is the top bar used throughout the whole site."
+    >
+      <TopBar>
```
kyleharper: I think we should add more stories so we can see some of the topbars that I know we need to support;

- left, center & right
- left, right (no centre)

And maybe show some of the contents we support in this component e.g. `Avatar`.

lib/TopBar/index.js
```diff
@@ -0,0 +1,28 @@
...
+      <div className={`top-bar__wrapper ${fixedClass}`}>
+        <Row className="top-bar__inner">{props.children}</Row>
+      </div>
+    </Grid>
+  );
+};
+
+TopBar.propTypes = {
+  fixed: PropTypes.bool,
+  children: PropTypes.shape()
```
kyleharper: Does children accept an object or is it a node/array of nodes? Your default prop for this sets it as an array so I believe this needs changing 👍

lib/TopBar/styles.scss
```diff
@@ -0,0 +1,125 @@
...
+  padding: 0;
+
+  @include respond-to($top-bar-main-breakpoint) {
+    height: $top-bar-height-medium;
+  }
+
+  .top-bar__wrapper {
+    background: #fff;
+    border-bottom: 1px solid $color-border-base;
+    box-shadow: 0 0 6px 0 rgba(230,234,237,0.29);
```
kyleharper: Could this use a box-shadow setting?

stories/components/TopBar.js
```diff
@@ -0,0 +1,67 @@
...
+
+storiesOf('Components', module).add('TopBar', () => (
+  <div>
+    <StoryItem
+      title="TopBar Wrapper"
+      description="TopBar is the top bar used throughout the whole site, can add a fixed attribute to fix it to the top."
+    >
+      <TopBar>
+        <Col className="top-bar__content top-bar__content--left" xs={10} md={5}>
+          <div className="top-bar__cell">
```
kyleharper: To make the `Topbar` more descriptive and easier to digest do you think we should add components for certain children?

Two possible candidates would be `top-bar__content` & `top-bar__cell`. It has a number of modifiers that could be props too so it fits pretty well.

So it would be something like;

```
<TopBar>
  <TopBarContent left>
    <TopBarCell bordered>
      <Icon name="backArrow" className="top-bar__action top-bar__action--collapse-left" />
    </TopBarCell>
  <TopBarContent>
</TopBar>
```

You see where I'm going with this 😄

Cyapow:
All feedback actioned. Have split off the content and cells into separate Components

PR #140, opened 2017-11-10
Feature/extending fields button style comments style

AmeeMorris:


lib/Field/styles.scss
```diff
@@ -69,4 +70,57 @@ $field-data-color: $neutral-base !default;
...
+    &:last-of-type {
+      margin-bottom: 0;
+    }
+  }
+
+}
+
+//statestyles
+
+.is-editor {
```
kyleharper: Normally state classes are added when a user performs an action e.g. `is-active` being added after user clicks a button. This feels like it should be a modifier for the `field` component.

Failing that, because the component itself doesn't really do much with the child component we could also just have element styles adding your styles to something like `field__editor`. What do you think? Do we even need to `is-editor` class?

kyleharper: Another idea, is to add modifiers to the elements. A really good example would be "instructions".

`field__instructions` & `field__instructions--input`.

stories/components/Field.js
```diff
@@ -62,5 +62,42 @@ storiesOf('Components', module)
...
+
+      <StoryItem
+        title="Field (Editing)"
+        description="A field component when in an editing state"
+      >
+        <Field
+          fieldId="123"
+          label="Field label text"
+          instructions="Instruction text to help inform the user of what they are mean't to add into this field."
+          isStructureEditor
```
kyleharper: Ideally we shouldn't add any domain language from external projects so generalising this would be great. Some examples would be;

- `canEdit`
- `isStatic`

stories/components/Field.js
```diff
@@ -62,5 +62,42 @@ storiesOf('Components', module)
...
+         >
+          <div>Some dummy text
```
kyleharper: What would normally go into here? I imagined something like Facebook text blocks to illustrate dummy content. Do we have those image assets? @nhqe we might need something from you here dude 👍

lib/Field/styles.scss
```diff
@@ -69,4 +70,57 @@ $field-data-color: $neutral-base !default;
...
+  .field__content {
+    box-shadow: $shadow-shallow;
+    background-color: white;
+    padding: 0;
+    border-radius: 5px;
+    border: 1px solid $neutral-light;
+  }
+  .field__child {
+    padding: 10px;
```
kyleharper: Should we be using a setting here?

lib/Field/index.js
```diff
@@ -14,29 +15,50 @@ const Field = props => {
...
 
-            {props.instructions && (
-              <div className="field__instructions">{props.instructions}</div>
-            )}
-          </div>
-        </Col>
-      </Row>
-    </div>
+              {props.instructions &&
+                (!props.isStructureEditor ? (
```
kyleharper: Do we need to add tests for this conditional? (and other works)

PR #142, opened 2017-11-15
Feature/choice field comments

AmeeMorris:


acairns:
[[Editor] Match design when commenting on choice fields](https://trello.com/c/awcCz0ls/2467-editor-match-design-when-commenting-on-choice-fields)

lib/Conversation/styles.scss
```diff
@@ -20,7 +20,7 @@ $conversation-meta-text-color: $neutral-base !default;
 /* ==========================================================================
    Styles
    ========================================================================== */
-.conversation {
+div.conversation {
```
kyleharper: Do we need a `div` here? If there is a clash elsewhere then lets resolve that by changing that instance of `.conversation`.

lib/Form/Checkbox/Group.js
```diff
@@ -19,10 +20,12 @@ export default class Group extends Component {
   constructor(props) {
     super(props);
     const selected = props.choices.filter(c => c.checked);
-    this.state = { choices: props.choices, selected };
+    this.state = { choices: props.choices, selected, highlightHover: false };
```
kyleharper: What is the `highlightHover` state used for? I've taken a look elsewhere but I didn't wan't to assume something incorrectly 😄

lib/Form/Label/index.js
```diff
@@ -2,9 +2,18 @@ import React from 'react';
...
+  id,
+  subtitle,
+  className,
+  highlight,
+  active,
+  highlightHover
+}) => {
   const textElementClasses = cx('form-input__text', {
     'form__input-text--highlight': highlight,
+    'highlight-active': highlightHover,
```
kyleharper: Is `highlight-active` meant to be a BEM modifier? If a state based class is required then maybe something like `is-highlighted` is something like that would be inline with other state class formats.

lib/Form/Label/index.js
```diff
@@ -2,9 +2,18 @@ import React from 'react';
...
 import cx from 'classnames';
 
-const Label = ({ label, id, subtitle, className, highlight, active }) => {
+const Label = ({
+  label,
+  id,
+  subtitle,
+  className,
+  highlight,
+  active,
```
kyleharper: I've forgotten but whats the difference between active and highlight again?

styles/base/_typography.scss
```diff
@@ -2,4 +2,13 @@ h1,
 h2,
 h3 {
     font-weight: $typo-weight-semi-bold;
-}
\ No newline at end of file
+}
+
+mark {
```
kyleharper: Nice new base styling 👍

lib/Form/Checkbox/Group.js
AmeeMorris: if its true it will show the hover style for the whole checkbox list, rather than just the individual checkbox option

lib/Form/Label/index.js
AmeeMorris: highlight just adds the highlight style, so the little yellow border, active will give it the active style ( the yellow background)

lib/Conversation/styles.scss
AmeeMorris: It was clashing with mark tags that shared the same class, I've renamed those

PR #146, opened 2017-11-21
Feature/comment design updates

AmeeMorris:


AmeeMorris:
[[Editor] Implement Design updates to Conversations](https://trello.com/c/PG7YHdOt/2567-editor-implement-design-updates-to-conversations)

lib/ExpandingTextArea/index.js
```diff
@@ -0,0 +1,105 @@
...
+        return this.setState({ rows: 1 });
+      }
+      return this.resizeTextArea();
+    }
+    return null;
+  }
+
+  setInitialRows() {
+    const rowSize = parseInt(
+      window.getComputedStyle(this.input).lineHeight,
```
kyleharper: Might be worth assigning this in a const above? e.g

```
const inputStyles = window.getComputedStyle(this.input);
```

Makes this method a little DRYer.

lib/ExpandingTextArea/index.js
```diff
@@ -0,0 +1,105 @@
...
+  static defaultProps = {
+    handleOnChange: () => {},
+    value: '',
+    focusOnMount: false,
+    setValue: false
+  };
+
+  constructor(props) {
+    super(props);
+    this.state = { inputValue: '', rows: 1, rowSize: 0, padding: 0 };
```
kyleharper: Is `rowSize` used as the physical size of the rows or is it the count of rows e.g. `rowCount`?

lib/ExpandingTextArea/index.js
```diff
@@ -0,0 +1,105 @@
...
+      newRows = 1;
+    }
+    return newRows;
+  }
+
+  resizeTextArea() {
+    const previousRows = this.state.rows;
+    this.input.rows = 1;
+    const newRows = this.calculateRows();
+    if (newRows === previousRows) {
```
kyleharper: Is the `previousRows` stored from a previous change event or is it the existing rowCount? I think `this.state.rows` can be confused with `this.state.rowSize` at times.

Maybe `if (newRows === this.state.rows)` reads better?

lib/ExpandingTextArea/index.js
AmeeMorris: Yeah thats confusing, I'll rename so they're more descriptive

PR #147, opened 2017-11-21
Avatar with popover changes

Cyapow:


lib/Avatar/AvatarWithPopover.js
```diff
@@ -29,9 +27,10 @@ const AvatarWithPopover = props => {
       placement={props.overlayPosition}
       overlay={popover}
     >
-      <span className="avatar__wrapper">
+      <button className="button button--icon-only avatar__wrapper">
         <Avatar {...props} className="avatar--with-toggle" />
-      </span>
+        {props.caret && <Icon name="caret" className="avatar--caret" />}
```
kyleharper: Is `avatar--caret` a modifier or an element?

Cyapow: Good point, have fixed

PR #149, opened 2017-11-22
Top bar amends

Cyapow:
@kyleharper take a look at this for me and see what you think. Not sure all the components are necessary but allows me to create the full story for TopBar in story book


lib/Branding/index.js
```diff
@@ -0,0 +1,31 @@
...
+import PropTypes from 'prop-types';
+import logoSVG from '../../assets/logo.svg';
+
+/**
+ * @usage
+ *
+ * <Branding url="http://example.com/myimage.png" alt="Alt Tag" />
+ */
+
+const Branding = props => {
```
kyleharper: `Branding` doesn't really say what this component outputs. It seems to output a logo so maybe `Logo`?

lib/Avatar/AvatarWithPopover.js
```diff
@@ -14,26 +15,48 @@ import { Avatar, Icon } from '../';
...
+    );
+
+    const onEnter = () => {
+      this.setState({ showing: true });
+    };
+    const onExited = () => {
+      this.setState({ showing: false });
+    };
+
+    const activeClass = cx({
```
kyleharper: I think in other instances we store all class logic in a single block like so;

```
const buttonClasses = cx('button button--icon-only avatar__wrapper', {
  'is-active': this.state.showing
});

<button className={buttonClasses}>...
```

This isn't consistent everywhere but I believe its been done like this in quite a few places recently so moving forward this makes sense to always do it like this when handling conditional classes.

lib/Navigation/index.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import cx from 'classnames';
+
+const Navigation = props => (
```
kyleharper: This looks almost identical to an existing component `NavTabs`: https://github.com/gathercontent/gather-ui/pull/143/files

I suggest we figure out how to use a single component and add modifiers where required.

lib/Navigation/index.js
```diff
@@ -0,0 +1,31 @@
...
+  <ul className="project-nav">
```
kyleharper: In `NavTabs` we use a `<nav>` tag which I believe is more inline with what this is.

lib/Navigation/styles.scss
```diff
@@ -0,0 +1,21 @@
+.project-nav {
```
kyleharper: This should be effected by the `NavTabs` feedback in a previous comment.

lib/Avatar/AvatarWithPopover.js
```diff
@@ -14,26 +15,48 @@ import { Avatar, Icon } from '../';
...
+      'is-active': this.state.showing
```
kyleharper: What styling is applied for `is-active`?

lib/ProjectSwitcher/index.js
```diff
@@ -0,0 +1,81 @@
...
+ * @usage
+ *
+ * <AvatarWithPopover
+ *  name="Poppy Cox"
+ *  email="poppycox@gmail.com"
+ *  >
+ *   ... popover content
+ * </AvatarWithPopover>
+ */
+class ProjectSwitcher extends Component {
```
kyleharper: This feels like it could be a variation on the existing component `DropdownSwitcher`. Is there any technical difference between the 2 or just visual styling?

lib/Avatar/AvatarWithPopover.js
Cyapow: Nice, didn't realise could do this. Fixed

lib/Branding/index.js
Cyapow: Renamed

lib/Avatar/AvatarWithPopover.js
Cyapow: Adds highlighting for the caret icon

lib/Navigation/index.js
Cyapow: Merged NavTabs into Navigation

lib/Navigation/index.js
Cyapow: Done

lib/ProjectSwitcher/index.js
Cyapow: Mainly visual style dropdowns look very different to popovers. For example there is no speech arrow to point to where it came from. We could always swap `DropdownSwitcher` to use `Popover` what are your thoughts @kyleharper

kyleharper: `DropdownSwitcher` accepts a `menu` prop. Maybe we could use the `Popover` for the prop and then add a modifier and style from there.

I think we should have everything required to migrate the `ProjectSwitcher` details into `DropdownSwitcher` 😄

Cyapow: @kyleharper have merged `ProjectSwitcher` and `DropdownSwitcher`

PR #151, opened 2017-11-24
Top bar fixes

Cyapow:


lib/Avatar/AvatarWithPopover.js
```diff
@@ -16,57 +16,106 @@ import { Avatar, Icon } from '../';
...
+    this.setState({ showing: true });
+  }
+
+  renderTrigger() {
     const classes = cx('button button--icon-only avatar__wrapper', {
       'is-active': this.state.showing
     });
 
+    const props = {};
+    if (typeof this.props.triggerEvent === 'string') {
```
kyleharper: Maybe this could do with being abstracted into a method? Something like the following would be quite readable;

```
const triggerEventProps = getOverlayTriggerEventProps()
```

That object could then be spread below too e.g. 

```
<button className={classes} {...triggerEventProps}>
```

lib/Avatar/AvatarWithPopover.js
```diff
@@ -16,57 +16,106 @@ import { Avatar, Icon } from '../';
...
+      props[this.props.triggerEvent] = this.togglePopover;
+    } else {
+      this.props.triggerEvent.forEach(event => {
+        if (event === 'onHover') {
+          props.onMouseEnter = this.showPopover;
+          props.onMouseLeave = this.hidePopover;
+        } else {
```
kyleharper: Try to refrain from using `else` for various reasons. Laracast did a decent video for this a while back: https://laracasts.com/series/simple-rules-for-simpler-code/episodes/2

By moving the logic into a method (previous feedback) I believe it should be quite simple and it'll make the logic really easy to understand e.g.

```
if (event === 'onHover') {
  return {
    onMouseEnter: this.showPopover,
    onMouseLeave: this.hidePopover
  }
}

return {
  [event]: this.togglePopover
}
```

Cyapow: Done these changes mate

PR #152, opened 2017-11-28
Feature/label click events

AmeeMorris:
Adding an 'overrideLabelDefault' prop to Form/Checkbox and Form/RadioButton which stops the label clicks firing onChangeHandler, and allows you to pass in a function for custom behaviour. In a similar vein you can now pass in props for MouseEnter and MouseLeave on labels for even more custom behaviour. 

Some clean up and renaming for clarity in both form choice fields and FileCard.


AmeeMorris:
[[Editor] Active conversations on choice fields and attachment fields](https://trello.com/c/b7RESdbM/2582-editor-active-conversations-on-choice-fields-and-attachment-fields)

lib/Form/Label/index.js
```diff
@@ -28,11 +30,24 @@ const Label = ({
     }
   };
 
+  const labelIsHovered = () => {
+    if (highlight) {
+      highlightIsHovered();
```
kyleharper: Why do we need to tell if we're hovering over the highlight? Doesn't this cause a loop as the parent set states...causing a re-render for all children and then we tell the parent to set the state here. Is that an issue?

kyleharper: It might be that we could make this simpler by saying;

- labels (buttons in this case) execute a function on mouse enter&leave
- If groups need to know when we're hovering over it then maybe the developer could handle that in the consuming app? E.g. passing a `highlighted` prop down to the group. The same could be said about being `active`...allow a active prop to be passed down.

By simplifying it like so it'll make the UI component easier to use and will remove the complexities around why we need `highlightIsHovered`, `highlightIsNotHovered` and the previous comment on this thread.

What do you think?

AmeeMorris:
@kyleharper PR amends done, let me know what you think

PR #153, opened 2017-12-04
Top bar fixes

Cyapow:
Various styling fixes which the app seems to break. 

Have also added a SectionHeader component for a standardised site wide title header.

### DIFF
--- a/lib/Avatar/styles.scss
+++ b/lib/Avatar/styles.scss
@@ -78,8 +78,13 @@ $avatar-pillbox-padding: $typo-size-lead/4 $typo-size-lead/3 !default;
 
 .avatar__caret {
   position: relative;
-  top: 10px;
-  margin-left: 10px;
+  top: 12px;
+  margin-left: $layout-spacing-base / 4;
+  opacity: .4;
+}
+
+.is-active .avatar__caret {
+  opacity: 1;
 }
 
 .avatar__target {
--- a/lib/Avatar/styles.scss
+++ b/lib/Avatar/styles.scss
@@ -26,7 +26,8 @@ $avatar-pillbox-padding: $typo-size-lead/4 $typo-size-lead/3 !default;
 }
 
 .avatar__wrapper {
-  display: inline-block;
+  display: inline-flex;
+  align-items: center;
 }
 
 .avatar--highlighted {
@@ -77,8 +78,6 @@ $avatar-pillbox-padding: $typo-size-lead/4 $typo-size-lead/3 !default;
 }
 
 .avatar__caret {
-  position: relative;
-  top: 12px;
   margin-left: $layout-spacing-base / 4;
   opacity: .4;
 }
--- a/lib/Button/styles.scss
+++ b/lib/Button/styles.scss
@@ -238,3 +238,15 @@ $color-btn-contained--bg: white !default;
     transform: scale(1.1);
   }
 }
+
+.button--plain {
+  background: none;
+  border: none;
+  box-shadow: none;
+  padding: 0;
+  transition: transform $animation-time-micro $animation-curve;
+
+  &:focus {
+    outline: none;
+  }
+}
--- a/lib/Button/styles.scss
+++ b/lib/Button/styles.scss
@@ -170,9 +170,14 @@ $color-btn-contained--bg: white !default;
 
 .button--clear {
   background: none;
+  padding: 0;
 
   &:hover {
     background: none;
+  }
+
+  &:focus {
+    outline: none;
   }
 }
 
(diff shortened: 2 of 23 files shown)


styles/overrides/_dropdown.scss
```diff
@@ -2,17 +2,24 @@
   box-shadow: $shadow-deep;
 }
 
-.dropdown-menu > li > a {
+.dropdown-menu > li > a,
+.dropdown-menu .dropdown__link,
+.dropdown-menu .dropdown__item .dropdown__link{
```
kyleharper: Normally the overrides area is used for overriding things we don't have control over. It looks like we're styling some BEM classes here. Are these styles for our components?

styles/settings/_typography.scss
```diff
@@ -6,6 +6,7 @@ $typo-stack-base: 'Whitney SSm A', 'Whitney SSm B', sans-serif;
 /* ==========================================================================
    Sizes
    ========================================================================== */
+$typo-size-header: 1.4rem;
```
kyleharper: This feels like a component setting rather than a global one. What do you think?

lib/Avatar/styles.scss
```diff
@@ -78,8 +78,13 @@ $avatar-pillbox-padding: $typo-size-lead/4 $typo-size-lead/3 !default;
 
 .avatar__caret {
   position: relative;
-  top: 10px;
-  margin-left: 10px;
+  top: 12px;
```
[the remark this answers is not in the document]
Cyapow: good shout, have amended

lib/DropdownSwitcher/styles.scss
```diff
@@ -29,21 +29,28 @@ $dropdown-switcher-title-margin-bottom: $layout-spacing-base/2 !default;
...
-  padding: 0 $layout-spacing-base / 2;
-  border-radius: 2px;
+  margin-left: $layout-spacing-base/4;
 }
 
-.dropdown-switcher .is-active .dropdown-switcher__caret {
-  background-color: $neutral-lightest;
-  opacity: 1;
+.dropdown-switcher__caret .icon {
+  vertical-align: middle;
```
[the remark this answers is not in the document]
Cyapow: Have used flex here

lib/Button/styles.scss
```diff
@@ -238,3 +238,15 @@ $color-btn-contained--bg: white !default;
     transform: scale(1.1);
   }
 }
+
+.button--plain {
```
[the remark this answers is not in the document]
Cyapow: Good shout, have changed to use clear but removed padding from it. And instead added a margin to the place its currently in use

styles/overrides/_dropdown.scss
Cyapow: They are used in our components but are being overridden by bootstrap. This enforces our own styling site wide for all dropdowns. I mean this could be moved to the app but I felt it would just be replicating the style we have here. Thoughts ? @kyleharper

kyleharper: That makes sense. The only thing that doesn't is I thought `.dropdown-menu > li > a` covered what we required? 

If the same styling is applied to both `.dropdown-menu > li > a` and the BEM related classes then I recommend duplicating the styles into the component. You could even use a `mixin` for the styles if you want it DRYer.

The reason why I suggest this is to keep the overrides responsible for overriding 3rd party libraries. If we start crossing the lines, we'll have issues later on I think.

lib/DropdownMenu/styles.scss
```diff
@@ -71,9 +72,9 @@ $dropdown-menu-link-padding: 7px 15px !default;
   padding: $dropdown-menu-link-padding;
   text-decoration: none;
   cursor: pointer;
-  color: #6f7c82;
-  font-weight: $typo-weight-roman;
-  font-size: $typo-size-small;
+  color: $color-text-base;
+  font-weight: $typo-weight-semi-bold;
+  font-size: $dropdown-menu-font-size;
```
Cyapow: Have updated the styling in this file to reflect the update to all dropdowns menus

lib/DropdownMenu/styles.scss
```diff
@@ -3,7 +3,8 @@
    ========================================================================== */
 $dropdown-menu-caret-margin: $layout-spacing-base/2 !default;
 $dropdown-menu-icon-size: $typo-size-small !default;
-$dropdown-menu-link-padding: 7px 15px !default;
+$dropdown-menu-link-padding: $layout-spacing-base/4 $layout-spacing-base/1.5 !default;
+$dropdown-menu-font-size: 13px !default;
```
kyleharper: Should this use a setting rather than a hardcoded value of 13px?

Cyapow: Only problem is $typo-size-base feels too big and $typo-size-small is too small lol. I mean I could always make a setting in between them two what you think @kyleharper ?

kyleharper:
Another thing @Cyapow ... do we have a README for these new components around the Topbar? If not it would be good to add those in.

lib/DropdownMenu/styles.scss
kyleharper: Get some design direction but we shouldn't be using hard coded values. We've had many issues in the past where hardcoded values have made the codebase less predictable and harder to work with.

Either;

- add a new setting using language and direction from designers
- derive font-size using a setting (we do this with spacing a lot)

PR #156, opened 2017-12-05
ShortcutTrigger Component

kyleharper:
This PR adds a new component for firing a function when the user presses a certain key combination. A readme has been added so to get more information, please read that 👍

### DIFF
--- a/lib/ShortcutTrigger/README.md
+++ b/lib/ShortcutTrigger/README.md
@@ -12,8 +12,8 @@ A component which executes a function when a combination of keys are pressed at 
 
 | Name                  | Type          | Default       | Required | Description                                         |
 | --------------------- |-------------- | ------------- | -------- |---------------------------------------------------- |
-| onShortcutTrigger    | Function      | ``      | Yes       | Executes when combination of keys are pressed.  |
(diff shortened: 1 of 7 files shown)


kyleharper:
@AmeeMorris after finding some bugs with the metaKey I have managed to resolve this moving over to the `keydown` event. Let me know what you think 😄

lib/ShortcutTrigger/README.md
```diff
@@ -0,0 +1,20 @@
...
+```
+<ShortcutTrigger shortcutKey="enter" onShortcutTrigger={func} withMetaKey />
+```
+
+### Props
+
+| Name                  | Type          | Default       | Required | Description                                         |
+| --------------------- |-------------- | ------------- | -------- |---------------------------------------------------- |
+| onShortcutTrigger    | Function      | ``      | Yes       | Executes when combination of keys are pressed.  |
+| shortcutKey   | string      | ``      | Yes       | Chosen key for the shortcut.  |
```
AmeeMorris: I've put 'N/A' for props that have no default, we should agree on a standard approach for signifying this.

kyleharper: Let's use 'N/A'. Updating now 👍

PR #157, opened 2017-12-06
changing focus-fallback styling

AmeeMorris:
A fix for conversation focusFallback causing scroll to shift if the bottom of the conversation (and the focusFallback itself) was out of view.

Simple fix, just making it so the fallback is always at the bottom of the window rather than the conversation itself.

### DIFF
--- a/lib/Conversation/styles.scss
+++ b/lib/Conversation/styles.scss
@@ -104,7 +104,8 @@ $conversation-meta-text-color: $neutral-base !default;
 }
 
 .conversation__focus-fallback {
-  position: absolute;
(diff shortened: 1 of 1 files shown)


AmeeMorris:
[[Editor] Bug: When deleting a comment and the reply + cancel buttons are out of the view port it throws the page around. See vid.](https://trello.com/c/SLyMV7R8/2614-editor-bug-when-deleting-a-comment-and-the-reply-cancel-buttons-are-out-of-the-view-port-it-throws-the-page-around-see-vid)

PR #162, opened 2017-12-12
Shortcuts popover tweaks

Cyapow:
Move UA Parser to the webapp so we can choose to show all shortcuts or not.

Slight margin tweak to plus icon


lib/Shortcut/ShortcutCommandKey.js
```diff
@@ -1,18 +1,12 @@
...
-  const ua = props.UAParser;
-
-  return (
-    <span className="shortcut__keyboard">
-      {ua.getOS().name === 'Mac OS' ? '⌘' : 'Ctrl'}
-    </span>
-  );
-};
+const ShortcutCommandKey = props => (
+  <span className="shortcut__keyboard">{props.mac ? '⌘' : 'Ctrl'}</span>
```
kyleharper: This seems so simple now that I would argue that we should just use `props.children` and let the consuming app decide what content should be rendered.

It does make the component extremely simple but that's fine.

Cyapow:
@kyleharper have done this, moved most of the work to the app and simplify the ui component

PR #163, opened 2017-12-18
stopping flex from shrinking avatar--supporting

AmeeMorris:
Easy fix for avatars appearing 'oval', this happened most notably in conversations when the conversation container was narrow.


AmeeMorris:
[Bug: Avatar heads after initial comment are distorted](https://trello.com/c/eFOBYs6b/2651-bug-avatar-heads-after-initial-comment-are-distorted)

PR #167, opened 2018-01-08
Field component changes

nhqe:
This PR adds visual structure to fields.


lib/Field/styles.scss
```diff
@@ -47,12 +47,32 @@ $field-data-color: $neutral-base !default;
 .field__instructions {
   font-style: italic;
   margin: $field-spacing-base 0 0;
+  padding: $field-spacing-base 0 0 0;
+  border-top: 1px solid $neutral-light;
+  transition: all 0.2s ease; // Does this go here?
```
kyleharper: This can go here but I wonder if we should use a setting? In `styles/settings/animations.scss` we have a setting for mirco animations at `.3s`. What do you think?

lib/Field/styles.scss
```diff
@@ -47,12 +47,32 @@ $field-data-color: $neutral-base !default;
...
 }
 
 .field__content {
-  padding-left: $field-spacing-base;
+  background: #F7F9FA;
```
kyleharper: Is there a setting for this?

lib/Field/styles.scss
```diff
@@ -47,12 +47,32 @@ $field-data-color: $neutral-base !default;
...
+  border:1px solid transparent;
+  border-radius: 4px;
+  padding: 30px; // Should be a variable?
```
kyleharper: I would say yes. Even though it seems strange at first if we change the global spacing in the future this should effect all components.

To get 30px you could do `$layout-spacing-base*1.5`.

lib/Field/styles.scss
kyleharper: Or we could change that setting to `0.2s`.

nhqe: Would prefer to set it to 0.2s, anytime I've used 0.3 it's felt too quick.

lib/Field/styles.scss
nhqe: Awesome, will do that now!

kyleharper: Another idea is that we could add some new settings for spacing 😄 

e.g.

```
$layout-spacing-base: 20px;
$layout-spacing-small: $layout-spacing-base/2;
$layout-spacing-large: $layout-spacing-base*1.5;
$layout-spacing-larger: $layout-spacing-base*2;
```

lib/Field/styles.scss
nhqe: Whoops, yep. Changed!

lib/Field/styles.scss
kyleharper: Sweet, change the setting and then use it in the 2 places you've added it 👍

nhqe:
Pushed up those changes 🤙

PR #174, opened 2018-01-12
Use ExpandingTextArea on field instructions instead of a plain input

AmeeMorris:
As title says and removing conditional which requires field to have existing instructions for the input to load in


AmeeMorris:
[[Structure] Add instruction fields when adding new text fields](https://trello.com/c/BiSXe21i/2704-structure-add-instruction-fields-when-adding-new-text-fields)

PR #176, opened 2018-01-15
New Tabs Component

kyleharper:
This PR adds a new component for tabs! I've got a couple of things to complete before this can be merged in;

- [x] add documentation


kyleharper:
@AmeeMorris Documentation added. Something else to note is that this component follows the compound component pattern as tabs and all the child component share internal state in order to set the active tab. This is why the usage follows that pattern so `<Tabs.Item>...` etc

PR #177, opened 2018-01-16
Adding Field and ExpandingTextArea config

AmeeMorris:
Adding a minRows prop to ExpandingTextArea, and an instructionsPlaceholder prop to Field.


AmeeMorris:
[[Structure] FE: Support guideline fields](https://trello.com/c/e2HrL6qU/2719-structure-fe-support-guideline-fields)

PR #183, opened 2018-01-22
LoadinOverlay

kyleharper:
This PR adds a new `LoadingOverlay` component which will be handy for handling our loading states across multiple other components.


kyleharper:
@AmeeMorris great shout about the README. I've added that in 👍

PR #189, opened 2018-01-31
DropdownMenu additions

AmeeMorris:
Adding a new 'outline' button style and a few bits to DropdownMenu:
- Changed closeDropdown as it was closing the dropdown regardless of where you clicked, overriding the toggleItems function.
- Adding a new 'withAdditional' item type
- Adding a downIcon prop incase you want to use that icon instead of the caret
- Adding a fullWidth prop to render the dropdown button and items full width
- Adding 'active' item functionality
- Tests now exist for this component

### DIFF
--- a/lib/DropdownMenu/index.js
+++ b/lib/DropdownMenu/index.js
@@ -45,6 +45,12 @@ class DropdownMenu extends Component {
         type = 'button';
       }
 
+      const activeContents = (
+        <span className="dropdown-item__tick is-active">
+          <Icon name="tick" size="micro" />
+        </span>
+      );
+
       switch (type) {
         case 'separator':
           return <li className="dropdown__separator" key={idKey} />;
@@ -52,11 +58,7 @@ class DropdownMenu extends Component {
         case 'link':
           return (
             <li className="dropdown__item" key={idKey}>
-              {item.active && (
-                <span className="dropdown-item__tick is-active">
-                  <Icon name="tick" size="micro" />
-                </span>
-              )}
+              {item.active && activeContents}
               <a className="dropdown__link" href={item.href}>
                 {item.name}
               </a>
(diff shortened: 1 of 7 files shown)


lib/DropdownMenu/index.js
```diff
@@ -47,16 +52,46 @@ class DropdownMenu extends Component {
...
           );
 
+        case 'withAdditional':
+          return (
+            <li className="dropdown__item" key={idKey}>
+              <button
+                className="dropdown__link dropdown__additional"
+                onClick={item.action}
+              >
+                {item.active && (
```
kyleharper: It seems like the contents for `item.active` is the same across all cases. It might be worth assigning it before the case and adding it in where required e.g.

```
static makeItems(items) {
  const activeContents = (
    <span className="etc">...</span>
  )

  // you can then add this into multiple cases.
}
```

lib/DropdownMenu/styles.scss
```diff
@@ -18,12 +18,52 @@ $dropdown-menu-font-color: $color-text-base !default;
...
+  .dropdown-menu  {
+    width: 100%;
+  }
+  .dropdown-menu__caret,
+  .dropdown-menu__down {
+    margin-left: auto;
+  }
+}
+
+.button--outline,
```
kyleharper: Should these be in the `Button` component styles file?

lib/DropdownMenu/styles.scss
```diff
@@ -18,12 +18,52 @@ $dropdown-menu-font-color: $color-text-base !default;
...
+  }
+}
+
 .dropdown-menu__button {
   display: flex;
   align-items: center;
+  &.button--outline,
+  &.button--link {
+    path fill {
+
```
kyleharper: This is empty. Is it required?

lib/DropdownMenu/styles.scss
```diff
@@ -120,3 +160,19 @@ $dropdown-menu-font-color: $color-text-base !default;
...
+.dropdown__additional {
+  display: flex;
+  .item-name {
+    display: block;
+    width: 65%;
+    white-space: nowrap;
+    overflow: hidden;
+    text-overflow: ellipsis;
+  }
+  .item-additional {
```
kyleharper: It might be worth sticking to the BEM naming convention here. Unless there's a reason not to?

AmeeMorris:
@kyleharper thanks for the review, changes done!

PR #190, opened 2018-02-01
Fixing long strings overflowing outside of conversations

AmeeMorris:
Really small PR

Added a bit of CSS to stop long strings popping out of the conversation text area. overflow-wrap and word-wrap are both used as firefox only likes word-wrap, and everything else likes overflow-wrap apparently. lol browsers


AmeeMorris:
[Bug: Comment overflow possible with long words etc.](https://trello.com/c/70k6JEsA/2759-bug-comment-overflow-possible-with-long-words-etc)

PR #191, opened 2018-02-05
Workflow amends

Cyapow:
Amendments to match design for StatusIndicator and added 2 new Icons. 

Have added an action prop to StatusIndicator to allow for object to show next to label. Have also fixed styling as well as text overflowing

### DIFF
--- a/lib/StatusIndicator/styles.scss
+++ b/lib/StatusIndicator/styles.scss
@@ -55,6 +55,8 @@ $status-indicator-margin-left: 1.3rem !default;
     color: $color-text-light;
   }
 
+  .status-indicator__circle {
+    background-color: $color-text-light;
   }
 }
 
--- a/lib/StatusIndicator/styles.scss
+++ b/lib/StatusIndicator/styles.scss
@@ -20,8 +20,8 @@ $status-indicator-margin-left: 1.3rem !default;
     color: $color-text-light;
     font-size: $typo-size-slight;
     margin-left: $status-indicator-margin-left;
-    font-weight: 500;
-    margin-top: 8px;
+    font-weight: $typo-weight-semi-bold;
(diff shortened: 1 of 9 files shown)


lib/StatusIndicator/styles.scss
```diff
@@ -17,8 +17,11 @@ $status-indicator-bordered-radius: .25rem !default;
   white-space: nowrap;
 
   .status-indicator__children {
-    color: $color-brand-dark;
-    font-size: $typo-size-small;
+    color: $color-text-light;
+    font-size: $typo-size-slight;
+    margin-left: $status-indicator-margin-left;
+    font-weight: 500;
```
kyleharper: Is there a setting for this weight?

lib/StatusIndicator/styles.scss
```diff
@@ -17,8 +17,11 @@ $status-indicator-bordered-radius: .25rem !default;
...
+    margin-top: 8px;
```
kyleharper: Maybe we could set this based on the base layout spacing setting? e.g. `$layout-spacing-base/2` = 10px

lib/StatusIndicator/styles.scss
```diff
@@ -61,11 +71,14 @@ $status-indicator-bordered-radius: .25rem !default;
...
-  border-radius: $status-indicator-bordered-radius;
-  padding: $layout-spacing-base/2;
+  border-bottom: $status-indicator-bordered-border;
+  padding-bottom: $layout-spacing-base/1.3;
 }
 
 .status-indicator__description {
+  white-space: initial;
   color: $color-brand-dark;
+  margin-top: 5px;
```
kyleharper: try using $layout-spacing-base to create this.

PR #194, opened 2018-02-06
Style tweaks to match design and to make popovers consistent

Cyapow:
Cleaned up StatusIndicator variables. 

Use consistent border and box shadow for all popovers


styles/settings/_layout.scss
```diff
@@ -10,6 +10,7 @@ $layout-spacing-base: 20px;
    ========================================================================== */
 $shadow-deep: 0 $layout-spacing-base/4 $layout-spacing-base/2 $neutral-light;
 $shadow-shallow: 0px 0px 6px 0px transparentize($neutral-light, 0.7);
+$shadow-base: 0 0 6px 0 rgba(230,234,237,0.29);
```
AmeeMorris: Minor thing, can the rgb code be the $neutral-light var (with transparentize) instead? Just makes it clearer what colour it is when quickly scanning

Cyapow: @AmeeMorris thanks, have updated :)

PR #202, opened 2018-02-13
Feature/message bar

AmeeMorris:
A few additions.

- A new MessageBar component used to display a simple messaging banner
- A new scrollToFixed prop on TopBar, which if true will only fix the TopBar once the user has scrolled to it. (Useful for when we have things, like the MessageBar, displaying above the TopBar)
- Some styling changes to TopBar dropdowns, so they are still positioned correctly if the TopBar isnt at the very top of the page
- A couple of new button styles, a slim one and a purple one
- A couple of linting fixes

### DIFF
--- a/lib/Button/styles.scss
+++ b/lib/Button/styles.scss
@@ -168,6 +168,22 @@ $color-btn-contained--bg: white !default;
   }
 }
 
+.button--purple-text {
+  background: white;
+  color: $primary-purple;
+  .icon path {
+    fill: $primary-purple;
+  }
+}
+
+.button--slim {
+  padding: $layout-spacing-base/10 $layout-spacing-base/2;
+  .icon {
+    margin-left: $layout-spacing-base/4;
+    vertical-align: text-bottom;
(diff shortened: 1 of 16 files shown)


lib/Button/styles.scss
```diff
@@ -168,6 +168,22 @@ $color-btn-contained--bg: white !default;
   }
 }
 
+.button--purple-text {
```
rikh42: Should this not describe the colour (so the colour can be changed one day) and instead describe the purpose?

AmeeMorris: Generally yes, but I don't think this one really serves a purpose (i.e. cancel/danger = red), its more I want a purple button because the rest of this is purple, I don't think purple signifies anything in particular in our UI. 

We also have a couple more that are named this way (light-grey, dark-grey)

kyleharper: Yeah, our `Button` component is legacy and needs a heavy refactor soon 😄

PR #208, opened 2018-02-27
MessageBar a tags should be underlined

AmeeMorris:
quick one liner, self explanatory!

### DIFF
--- a/lib/MessageBar/styles.scss
+++ b/lib/MessageBar/styles.scss
@@ -9,6 +9,9 @@
   a {
     color: white;
     text-decoration: underline;
+    &:hover {
+      text-decoration: none;
+    }
   }
   &__inner {
     display: flex;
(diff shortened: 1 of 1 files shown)


AmeeMorris:
[Bug: Link styling lost on alert bar](https://trello.com/c/yjM3Tn6O/2905-bug-link-styling-lost-on-alert-bar)

lib/MessageBar/styles.scss
```diff
@@ -8,6 +8,7 @@
   z-index: 3;
   a {
     color: white;
+    text-decoration: underline;
```
kyleharper: Have is the hover state? At the moment it won't do anything right?

AmeeMorris: @kyleharper this is true, however the bug was in certain places the a tag did not have the underline. 

But we should have a defined hover state for clarity, I'll check what people want and add it in.

PR #209, opened 2018-03-01
Issue: Cant access last item in TopBar dropdown 

AmeeMorris:
Just some simple css to subtract the TopBar height from the dropdown height


AmeeMorris:
[Bug: Can't access lowest item in the item list](https://trello.com/c/bvBfUDV1/2888-bug-cant-access-lowest-item-in-the-item-list)

PR #210, opened 2018-03-01
Issue: Comment editing content loss bug

AmeeMorris:
This fixes a bug where if you edited a comment without changing the value the comment content would disappear.

This was down to the inputValue state in CommentForm only getting a value if the input value had actually changed.


AmeeMorris:
[Bug: Comment content dissapearing when saved again](https://trello.com/c/3fPmQtML/2893-bug-comment-content-dissapearing-when-saved-again)

PR #218, opened 2018-03-07
fixing a few bugs around comment editing

AmeeMorris:
A few bugs fixed: 

1. The showComment state being lost when clicking on the edit Button (now a div)
2. You couldn't delete all the text in a CommentForm if you pass a value prop
3. The focus fallback would throw you down the page when focused (set back to fixed, if this was changed for a reason give me a shout, couldn't see any nastiness caused)

### DIFF
--- a/lib/Conversation/styles.scss
+++ b/lib/Conversation/styles.scss
@@ -121,7 +121,7 @@ $conversation-meta-text-color: $neutral-base !default;
 }
 
 .conversation__focus-fallback {
-  position: fixed;
+  position: absolute;
   bottom: 0;
   left: -999em;
 }
--- a/lib/Conversation/styles.scss
+++ b/lib/Conversation/styles.scss
@@ -121,7 +121,7 @@ $conversation-meta-text-color: $neutral-base !default;
 }
 
 .conversation__focus-fallback {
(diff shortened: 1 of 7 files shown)


AmeeMorris:
[Bug: View replies overlapping with cancel and send when message has collapsed](https://trello.com/c/uowE5cal/2894-bug-view-replies-overlapping-with-cancel-and-send-when-message-has-collapsed)

lib/Conversation/styles.scss
```diff
@@ -121,7 +121,7 @@ $conversation-meta-text-color: $neutral-base !default;
 }
 
 .conversation__focus-fallback {
-  position: absolute;
+  position: fixed;
```
kyleharper: This was set to absolute because it was pinging users to the top of the screen.

AmeeMorris:
@kyleharper changes done, I've moved focusFallback to each comment rather than per conversation. That way its very unlikely that a user would be shifted about since the controls that cause you to use the focusFallback are at the bottom of each comment, and if it did happen any shifting about should be minimal.

Also Travis fixed 🎉 it was Jest the whole time, see https://github.com/facebook/jest/issues/3855

PR #220, opened 2018-03-10
Field changes

kyleharper:
Just some style changes and reordering the validations / actions for fields.

------------

Also - CI seems to be passing but I have a failing test locally.


styles/helpers/_visibility.scss
```diff
@@ -0,0 +1,4 @@
+.hide-off-page {
+  position: absolute;
+  left: -999em;
```
rikh42: Is it polite to add display: none here, so screen readers don't keep trying to read it out - or will that prevent events from happening - not sure where this will be used.

kyleharper: Good shout on screen readers. However we'll resolve that by making them invisible via `aria-hidden=true`. Those roles will be applied when writing them elements themselves...nothing in this PR.

`display:none` causes multiple issues like not being able to access the node in the DOM. Hence the name of the class...we're not just hiding the element, we're sending it off page.

lib/Field/index.js
```diff
@@ -34,11 +34,11 @@ const Field = props => {
               {props.label}
             </EditableTextWrapper>
           )}
-          <FieldActions fieldId={props.fieldId} actions={props.actions} />
           <FieldValidations
             fieldId={props.fieldId}
             validations={props.validations}
           />
+          <FieldActions fieldId={props.fieldId} actions={props.actions} />
         </Col>
```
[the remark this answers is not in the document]
kyleharper: This is handle via the style changes (we use flex box & ordering) on this PR 👍 Check that out and approve if you're ok with that.

PR #222, opened 2018-03-12
Increased font-size of instructions and labels to $typo-size-base

nhqe:


nhqe:
[UX issue: Field instruction text is too small](https://trello.com/c/hWXXaKBE/2975-ux-issue-field-instruction-text-is-too-small)

PR #226, opened 2018-03-12
Item list updates

nhqe:


nhqe:
[UX Issue: Item list looks outdated](https://trello.com/c/qRTmE1Mf/2995-ux-issue-item-list-looks-outdated)

lib/List/styles.scss
```diff
@@ -49,7 +47,18 @@ $list-border: 1px solid $color-border-base;
   }
 
   .is-current > & {
-    background: #f9f9f9;
+    background: #EBF5FD;
```
kyleharper: Do we have a setting for this?

nhqe: @kyleharper we don't. I attempted to lighten up our $primary-blue colour but once lightened it gave off a purplish hue. Shall I add this to gather-ui as a $primary-light-blue colour?

PR #227, opened 2018-03-12
Updating icon hit area with pixel padding.

nhqe:
There's a slight issue here, as every button will have to have pixel values specified due to the discrepancies in icon sizing. Would love a chat about how we can make this applicable across every icon, but it may be that we have to size as and when.

### DIFF
--- a/lib/List/styles.scss
+++ b/lib/List/styles.scss
@@ -100,10 +100,10 @@ $list-border: 1px solid $color-border-base;
   .is-active svg {
     transform: rotate(0deg);
   }
+}
 
-  .button--icon-only-cover {
-  padding: 10px 8px;
-  }
+.list__button {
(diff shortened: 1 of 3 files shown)


nhqe:
[UX Issue: Expanding children in the item list, the hit area is a tad small](https://trello.com/c/PBOnmcUB/2979-ux-issue-expanding-children-in-the-item-list-the-hit-area-is-a-tad-small)

lib/List/styles.scss
```diff
@@ -100,6 +100,10 @@ $list-border: 1px solid $color-border-base;
   .is-active svg {
     transform: rotate(0deg);
   }
+
+  .button--icon-only-cover {
```
kyleharper: When styling other components we append a class to do so like `list__button`.

PR #228, opened 2018-03-12
Removed outline from search input • Added a border on search input focus

nhqe:
I've added `outline: none` to the search input, however it looks like buttons already have these applied, so maybe the structure button isn't using the same class as the buttons in gather-ui.


nhqe:
[UX Issue: The "Save structure" button, and the "Mark status as complete" button have that grey browser filth around them](https://trello.com/c/ZDrfD0yU/2977-ux-issue-the-save-structure-button-and-the-mark-status-as-complete-button-have-that-grey-browser-filth-around-them)

PR #229, opened 2018-03-12
EditableTextWrapper now goes into the editing view on text click

AmeeMorris:
Instead of just the pencil icon click


AmeeMorris:
[UX Issue: I expect to name a label by clicking. Having to press the pencil does not feel intuitive.](https://trello.com/c/1mQeoPfe/2987-ux-issue-i-expect-to-name-a-label-by-clicking-having-to-press-the-pencil-does-not-feel-intuitive)

PR #231, opened 2018-03-13
Updating grab SVG

nhqe:


nhqe:
[UX Issue: Field control buttons aren't the same size](https://trello.com/c/VqfYlFzU/2997-ux-issue-field-control-buttons-arent-the-same-size)

PR #233, opened 2018-03-15
Due date picker

Cyapow:
Have added a DueDatePicker component. Will most likely be used in the multiple places we choose due dates. 
Shows a popover with a calendar and a time picker. The popover is split into multiple components to make it easier to work with. 

DueDateLabel is a separate component which just displays the date


lib/DueDatePicker/DueDateHeader.js
```diff
@@ -0,0 +1,74 @@
...
+          listClassName="dropdown-menu-right"
+          items={[
+            {
+              action: props.removeDueDate,
+              name: 'Remove due date',
+              linkType: 'danger'
+            }
+          ]}
+        >
+          <Icon style={{ fontSize: '18px' }} name="menuDotted" />
```
kyleharper: Should this line style be here?

lib/DueDatePicker/DueDateHeader.js
```diff
@@ -0,0 +1,74 @@
...
+      nextWeek: 'dddd [at]',
+      lastDay: '[Yesterday at]',
+      lastWeek: '[Last] dddd [at]',
+      sameElse: 'MMMM Do YYYY [at]'
+    });
+
+    const time = props.dueTime.format('LT');
+
+    const classes = cx('duedate__header', {
+      'duedate__header--overdue': date < moment().set({ hour: 0, minute: 0 })
```
kyleharper: If you just want to set the text colour red there is a help class for that. Check our the helper styles 👍

lib/DueDatePicker/DueDateLabel.js
```diff
@@ -0,0 +1,54 @@
...
+        No due date set
+      </Button>
+    );
+  }
+
+  const classNames = cx('duedate__label', {
+    'duedate__label--overdue': dueDate.overdue
+  });
+
+  let icon = '';
```
kyleharper: It'd be worth using ternary here I think.

lib/DueDatePicker/index.js
```diff
@@ -0,0 +1,166 @@
...
+import moment from 'moment';
+import cx from 'classnames';
+import Button from '../Button';
+import DueDateHeader from './DueDateHeader';
+import DueDateButton from './DueDateButton';
+
+class DueDatePicker extends Component {
+  constructor(props) {
+    super(props);
+    this.handleDayClick = this.handleDayClick.bind(this);
```
kyleharper: We've started using arrow functions rather than binding this in the constructor 😄 

Try it out and see how you think it looks etc

lib/DueDatePicker/index.js
```diff
@@ -0,0 +1,166 @@
...
+    this.handleTimeClick = this.handleTimeClick.bind(this);
+    this.applyDueDate = this.applyDueDate.bind(this);
+    this.removeDueDate = this.removeDueDate.bind(this);
+
+    this.initialState = {
+      showing: false,
+      changed: false,
+      selectedDay: this.props.dueDate ? this.props.dueDate.toDate() : null,
```
kyleharper: I didn't think you had access to `this.props` in the constructor. That's why `props` is passed through as an argument. I'd test this a little further and use the argument.

lib/DueDatePicker/styles.scss
```diff
@@ -0,0 +1,349 @@
...
+$duedate-timepicker-height: 220px !default;
+/* ==========================================================================
+   Styles
+   ========================================================================== */
+.duedate__popover.duedate__popover {
+  max-width: $duedate-popover-width;
+  width: $duedate-popover-width;
+  margin-top: $layout-spacing-base / -2;
+
+  .arrow {
```
kyleharper: Is there a task to hide all arrows for popovers? If so shall we do this globally? 😄

Pinging @jamesdarracott & @nhqe for input.

lib/DueDatePicker/styles.scss
```diff
@@ -0,0 +1,349 @@
...
+  color: $color-text-light;
+
+  .button {
+    text-decoration: underline;
+    color: $color-text-light;
+  }
+
+  .icon {
+    margin-right: $layout-spacing-base / 4;
+    top: -2px;
```
kyleharper: Is this to centre the icon? Is so we could use flexbox.

lib/DueDatePicker/styles.scss
```diff
@@ -0,0 +1,349 @@
...
+      fill: $primary-red;
+    }
+  }
+
+  .icon path {
+    fill: $primary-red;
+  }
+
+  .icon {
+    top: -1px;
```
kyleharper: Is this to centre the icon? Is so we could use flexbox.

lib/DueDatePicker/styles.scss
```diff
@@ -0,0 +1,349 @@
...
+.duedate__popover .DayPicker {
+  padding: $layout-spacing-base / 2;
+
+  &:after {
+    content: "";
+    display: table;
+    clear: both;
+  }
+
+  .DayPicker-wrapper {
```
kyleharper: Is `DayPicker` is provided by the 3rd party? If so I wonder if this should be handled in the overrides styling area? What do you think?

lib/DueDatePicker/styles.scss
```diff
@@ -0,0 +1,349 @@
...
+    &.DayPicker-Day--selected {
+      background-color: $primary-red;
+      color: white;
+    }
+  }
+}
+
+/* DueDate Time picker dropdown */
+
+.duedate__popover .duedate__header--time-picker {
```
kyleharper: Is `duedate__header--time-picker` a modifier or a element? Reads like a element so maybe `duedate__time-picker`? Reducing the length of BEM selectors is always a win 😄

lib/DueDatePicker/styles.scss
```diff
@@ -0,0 +1,349 @@
...
+  color: $color-text-base;
+  text-decoration: underline;
+
+  &:hover {
+    color: $color-text-base;
+  }
+}
+
+.duedate__header--overdue .duedate__header--date {
+  color: $primary-red;
```
kyleharper: Nothing may need to change here but anywhere with just `color: $primary-red` is an area that may be better suited to use a helper class. Let me know what you think.

kyleharper: Doing at times can help reduce the amount of BEM bloat which can happen when you elements have multiple modifiers. One major goal is to make our codebase usable for anyone and a mixture of helper classes and BEM can work really well when balanced in the right scenarios.

package.json
```diff
@@ -59,12 +59,14 @@
     "eslint-config-prettier": "^2.6.0",
     "font-awesome": "^4.7.0",
     "linkifyjs": "^2.1.6",
+    "moment": "^2.21.0",
```
kyleharper: Make sure we're not suffering from this: https://github.com/jmblog/how-to-optimize-momentjs-with-webpack

lib/DueDatePicker/DueDateHeader.js
Cyapow: My bad not totally sure where I copied that from as I cant see it anywhere else lol

lib/DueDatePicker/styles.scss
Cyapow: @kyleharper if I add display: flex it seems to mess with the positioning of the popover. Not sure of how to get around it. I might leave it as is for now. And when we come to redo popovers we can redo it then maybe? what you think

kyleharper: Sounds good to me 👍

PR #236, opened 2018-03-16
New BoundaryClickWatcher mouse events

AmeeMorris:
I've just added some mouseEnter and MouseLeave props to it. Maybe we should just rename it BoundaryWatcher!

Also some minor styling changes for parent hover stuff


kyleharper:
I agree with using the default names 😄I've approved the changes, thanks for explaining a little more, it makes perfect sense 👍

PR #242, opened 2018-03-21
A few bits

AmeeMorris:
Quite a few new things

1. New AvatarWithInformation component
2. New SearchDropdown component
3. New Contributors component
4. New DropdownMenu item type (Avatar)
5. Icons can now display a little notification circle
6. Person SVG change


lib/Contributors/index.js
```diff
@@ -0,0 +1,145 @@
...
+            >
+              <Icon name="plusCircle" />
+            </Button>
+          )}
+        </div>
+        <div className="contributors__list">
+          <div className="contributors__subheading">
+            Assigned ({assigneeLength})
+          </div>
+          {this.getcurrentAssignees()}
```
kyleharper: This feels like it should be in a component of it's own.

lib/Contributors/index.js
```diff
@@ -0,0 +1,145 @@
...
+              fullWidth
+              handleOnChange={this.props.handleSearchChange}
+              onInputClear={this.props.handleClearResults}
+              resultsTitle="Assign..."
+            />
+          )}
+        {this.props.currentViewers.length > 0 && (
+          <div className="contributors__list contributors__viewing-list">
+            <div className="contributors__subheading">Viewing</div>
+            {this.getCurrentViewers()}
```
kyleharper: This feels like it should be in a component of it's own.

lib/Contributors/index.js
```diff
@@ -0,0 +1,145 @@
...
+
+    const classNames = cx('contributors', {
+      'contributors--search-open': this.state.showSearch
+    });
+
+    return (
+      <div className={classNames}>
+        <div className="contributors__heading">
+          Contributors
+          {this.props.userCanAddAssignee && (
```
kyleharper: Maybe rename `userCanAddAssignee` to a more generic prop, like `showAddUserButton`. We can then remove another mention of some business/domain language from the UI component.

lib/Contributors/index.js
```diff
@@ -0,0 +1,145 @@
...
+        url={viewer.avatar}
+        name={viewer.name}
+        initials={viewer.initials}
+        colour={viewer.colour}
+        smallSize
+      />
+    ));
+
+  getcurrentAssignees = () => {
+    const assignees = this.props.currentAssignees;
```
kyleharper: Maybe just deconstruct `currentAssignees` and map over that? Feels strange how we're assigning a new const here calling it `assignees` when we already have `currentAssignees`. I'm easy though, let me know if you disagree 👍

lib/Avatar/AvatarWithInformation.js
```diff
@@ -0,0 +1,87 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import cx from 'classnames';
+import { Avatar } from '../';
+
+class AvatarWithInformation extends Component {
```
kyleharper: I'm not too sure if we need to create a new component for showing information within the Avatar component. `AvatarWithPopover` requires some funky logic to get working but I believe this could be a simple case of rendering children into the `Avatar` component. e.g.

```
<Avatar>
  <AvatarInformation />
</Avatar>
```

Obviously this will have an affect on how the styling is currently setup and maybe we could add the `avatar__wrapper` in another way but I believe it'd be better if we didn't add another component for rendering additional content for the Avatar component itself.

The main reason is it becomes harder to discover where stuff happens but keeping it in one component with stuff like this is the best approach. When a component gets too large then that is a different issue but `Avatar` is pretty lean right now.

lib/Avatar/AvatarWithInformation.js
```diff
@@ -0,0 +1,87 @@
...
+          {this.props.name && (
+            <span className="avatar__name">{this.props.name}</span>
+          )}
+
+          {this.props.email && (
+            <span className="avatar__email">{this.props.email}</span>
+          )}
+        </div>
+        {this.props.additional &&
+          this.state.showAdditional && (
```
kyleharper: After adding my feedback about moving the information content into a child component I wondered if this is worth taking over to the Avatar component too?

If additional information was rendered but was visually hidden and shown via CSS...would that be bad? It would remove the need for more component logic...unless we need to remove the additional content from the DOM for a reason?

If so...how do screen readers see this content?

AmeeMorris:
@kyleharper changes done, let me know what you think

PR #244, opened 2018-03-26
Hiding popover arrows

AmeeMorris:
As the title says. Removing popover arrows to be in line with new designs

### DIFF
--- a/lib/PopoverWrapper/styles.scss
+++ b/lib/PopoverWrapper/styles.scss
@@ -1,0 +1,3 @@
+.popover .arrow {
+  visibility:hidden;
+}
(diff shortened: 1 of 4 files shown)


AmeeMorris:
[UX Issue: Arrows on dropdowns aren't styled correctly](https://trello.com/c/v302uGRD/2870-ux-issue-arrows-on-dropdowns-arent-styled-correctly)

lib/PopoverWrapper/styles.scss
```diff
@@ -0,0 +1,3 @@
+.popover .arrow {
```
kyleharper: Are these overrides? If so maybe move them there.

AmeeMorris:
@kyleharper very good point! 😅changes done

PR #245, opened 2018-03-26
FileCard label z-index

AmeeMorris:
The FileCard label would appear over other dom elements (i.e. a fixed position header).

So I've just removed the z-index, this seems to have no negative impact, the label is still visible with or without a preview image, not sure why it was there to begin with. 😅


AmeeMorris:
[Bug: A file label appears on the top bar when I scroll down](https://trello.com/c/tql9Ol76/3020-bug-a-file-label-appears-on-the-top-bar-when-i-scroll-down)

PR #246, opened 2018-03-27
Inline Comment Edits

kyleharper:
This PR updates the way we edit comments for conversations. It improves the editing experience.

### DIFF
--- a/lib/Conversation/Comment.js
+++ b/lib/Conversation/Comment.js
@@ -35,7 +35,7 @@ class Comment extends Component {
 
   showEditForm = () => {
     this.setState({ showEditForm: true });
-  }
+  };
 
   hideEditForm = () => {
     this.setState({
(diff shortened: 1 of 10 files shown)


lib/Conversation/Comment.js
```diff
@@ -88,62 +75,62 @@ class Comment extends Component {
...
+            )}
+
+            {userCanEdit &&
+              !this.state.showEditForm && (
+                <span className="conversation__actions">
+                  <Button
+                    className="conversation__meta conversation__edit-link"
+                    types={['link', 'link-default', 'collapse']}
+                    clickHandler={this.showEditForm}
+                  >
```
AmeeMorris: I made the edit Button a div previously due to how Quill's selection change doesn't get triggered when changing focus to a button element. Have you found a different work around? It wasn't a super obvious bug, but it would cause some weirdness

kyleharper: Ah ok, I forgot about that. Maybe we should add a test to document why this is the case. Will add a test now to say that we need this to be the case due to 3rd part libraries.

AmeeMorris: Sounds good! It is a bit of an annoying one 😅

PR #259, opened 2018-04-13
Feature/conversation mentions

AmeeMorris:
Adding mentions to the Conversation component.

Using https://github.com/signavio/react-mentions to do most of the heavy lifting, with a custom search and and suggestion display.

Also we're now checking comment body for any `@blah` substrings. If any exist they will be wrapped in a `span.mention`


AmeeMorris:
[@ing in comments](https://trello.com/c/RuhSdRKl/3102-ing-in-comments)

lib/Conversation/CommentForm/CommentFormInput.js
```diff
@@ -0,0 +1,103 @@
...
+
+  state = {
+    inputHeight: '0px'
+  };
+
+  componentDidMount() {
+    this.setState({ inputHeight: this.getInputHeight() }); // eslint-disable-line react/no-did-mount-set-state
+  }
+
+  getInputHeight = () => {
```
kyleharper: Are we using any of this input height logic? I can't see state being used in the rending or in the prop function passed. It may be left overs from the `ExpandingTextArea` component?

lib/Conversation/CommentForm/CommentFormInput.js
```diff
@@ -0,0 +1,103 @@
...
+            .split(' ')
+            .filter(subStr => subStr.lastIndexOf(term, 0) === 0).length > 0 ||
+          user.display.toLowerCase().lastIndexOf(term, 0) === 0
+      );
+      return filtered;
+    }
+    return this.props.users;
+  };
+
+  handleOnChange = () => {};
```
kyleharper: Should this be empty?

lib/Conversation/CommentForm/CommentFormInput.js
kyleharper: I think we may use `this.props.onRowCountChange` but I'm not sure.

kyleharper: Rather than copying `ExpandingTextArea` logic we could use the render prop pattern?

For example, here is some logic for using the new `CommentFormInput` which is wrapped in a `ExpandingTextArea` component, which accepts a function so it doesn't render a text area but instead shares methods/state.

```
<ExpandingTextArea onRowCountChange={this.props.onRowCountChange}>
  (calculateRowCount, height) =>
    <CommentFormInput onChange={calculateRowCount} />
</ExpandingTextArea>
```

Obviously there is more detail than what I have described there but this would remove the need to copy logic and would keep the source of truth for expanding text areas in 1 location 😄

lib/Conversation/CommentForm/index.js
```diff
@@ -58,8 +60,10 @@ class CommentForm extends Component {
     if (e) {
       e.preventDefault();
     }
-    this.props.onSubmit(this.state.inputValue);
-    const inputValue = !this.props.isSubmitting ? '' : this.state.inputValue;
+    this.props.onSubmit(this.removeMentionMarkup());
```
kyleharper: It might be worth doing `this.removeMentionMarkup()` and storing the value rather than doing it twice when `!this.props.isSubmitting` is true.

lib/Conversation/CommentForm/index.js
```diff
@@ -58,8 +60,10 @@ class CommentForm extends Component {
...
+    const inputValue = !this.props.isSubmitting
```
kyleharper: Could you share some knowledge for how the submitting/saving state should perform? It's always been a gap in my knowledge about Conversations.

lib/Conversation/CommentForm/CommentFormInput.js
AmeeMorris: We use onRowCountChange in the app for some positioning logic (that way the conversation doesn't overlap another conversation thats positioned directly below if the textarea expands).

The logic is quite different from expanding text area too, so its not really copied logic (although it achieves a similar result)

lib/Conversation/CommentForm/CommentFormInput.js
AmeeMorris: Nope, that shouldn't even be there, nice catch 😅

lib/Conversation/CommentForm/index.js
AmeeMorris: You know, I cant even remember. I suspect it was something that was previously used, but isn't any more? I can't see anywhere where its set to true. Loath to change it now just incase it somehow breaks it, and its not really hurting anything right now. Maybe worth reviewing on a future refactor? We've thrown around the idea of refactoring conversation a few times 🙂

AmeeMorris: Putting this here too for clarity: A while back we added functionality for a comment submitting state (isSubmitting) if isSubmitting is true we’d display the comment form still populated with text for a little longer with the text area greyed out/disabled and the submit button would become a spinner.

This was either never hooked up and forgotten about (likely) or taken out for other reasons

AmeeMorris:
@kyleharper changes/comments replied to, let me know what you think

PR #266, opened 2018-04-25
Adding a small version for CheckToggle

AmeeMorris:
And a displayChecked prop to give us a lovely green background when they're checked

### DIFF
--- a/lib/CheckToggle/styles.scss
+++ b/lib/CheckToggle/styles.scss
@@ -92,3 +92,29 @@ $borderColour: #F8FBFC;
 .toggle-switch--inline + label {
   display: inline-block;
 }
+
+.size-small {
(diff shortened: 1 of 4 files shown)


lib/CheckToggle/styles.scss
```diff
@@ -92,3 +92,29 @@ $borderColour: #F8FBFC;
...
+    &:after {
+      top: 2px;
+      left: 2px;
+      bottom: 4px;
+      width: 12px;
+      height: 12px;
+    }
+  }
+  .toggle-switch:checked + .toggle-switch__label:after {
+    margin-left: 10px;
```
kyleharper: If this is derived from our base spacing then maybe use the layout setting?

PR #270, opened 2018-05-01
Add a higher z-index for the top bar

Cyapow:
Adding a higher z-index will allow dropdowns it contains to sit above others used elsewhere in the app


Cyapow:
@AmeeMorris totally agree. it looks like elsewhere in the app we have crazy numbers in 1000s

PR #276, opened 2018-05-09
Item row updates

kyleharper:
This PR adds styles for boarded lists and stacked item rows.

There is a issue with bordered lists when having lists within lists. We can resolve this later.

### DIFF
--- a/stories/components/List.js
+++ b/stories/components/List.js
@@ -77,6 +77,21 @@ storiesOf('Components', module)
         </List>
       </StoryItem>
 
+      <StoryItem title="List (boarded)">
+        <List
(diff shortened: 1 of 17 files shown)


stories/components/List.js
```diff
@@ -66,66 +66,127 @@ storiesOf('Components', module)
...
+                    </ListItemText>
                   </ListItem>
                 </List>
               </ListItem>
             </List>
           </ListItem>
         </List>
       </StoryItem>
+
+      <StoryItem title="Boarded List: Full Example">
```
AmeeMorris: Should this be bordered? 😄

PR #277, opened 2018-05-09
Fix onCommentChange for onSubmit and changed icons

AmeeMorris:
As the title says, not sure if this was missed out when refactoring the commentForm for mentions recently.


AmeeMorris:
[Bug: Comment not deselecting.](https://trello.com/c/LytY36Ch/3229-bug-comment-not-deselecting)

PR #279, opened 2018-05-10
List fixes

kyleharper:
The issues in a previous PR (https://github.com/gathercontent/gather-ui/pull/276) came back to bite us a little too hard.

This PR fixes that issue. It adds the logic for splitting the children in `ListItem` so that `List` components are rendering in a different location.

This was required as when set as "current" the styling change would affect all children including list item within list items, which is not correct. "current" should only affect a single row at a time.


lib/List/ListItem.js
```diff
@@ -13,46 +12,65 @@ class ListItem extends Component {
...
+
+    const subListChildren = React.Children.map(children, child => {
+      if (child.type === List) {
+        return React.cloneElement(child);
+      }
+
+      return null;
+    });
+
+    const hasSubList = subListChildren ? subListChildren.length > 0 : false;
```
AmeeMorris: Neat!

PR #284, opened 2018-05-31
Ux/issues

Cyapow:
- Change all icons default colour (fixes the opacity to go with this on the dropdowns and avatars)
- Fix text overflow on avatar additional info
- Add collapse prop to TopBarContent and TopBarCell

### DIFF
--- a/lib/Avatar/styles.scss
+++ b/lib/Avatar/styles.scss
@@ -152,16 +152,16 @@ $avatar-pillbox-padding: $typo-size-lead/4 $typo-size-lead/3 !default;
     text-overflow: ellipsis;
     white-space: nowrap;
   }
-}
 
-.avatar__wrapper--additional-visible {
-  .avatar__additional {
-    display: block;
-    margin-left: auto;
+  &.is-visible {
+    .avatar__additional {
+      display: block;
+      margin-left: auto;
 
-    button[class*="button--link"] {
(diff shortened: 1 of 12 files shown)


lib/Avatar/styles.scss
```diff
@@ -153,7 +152,9 @@ $avatar-pillbox-padding: $typo-size-lead/4 $typo-size-lead/3 !default;
     text-overflow: ellipsis;
     white-space: nowrap;
   }
+}
 
+.avatar__wrapper--additional-visible {
```
kyleharper: Always a grey area in BEM but maybe `avatar__wrapper--additional-visible` could use a state class instead e.g. `is-visible` or `is-open`. 

The reason for this is 1 disadvantage of BEM is that you can stick too much to the modifier principle. Making it less legible. This is how I see it;

**Use a modifier** 

- when a components take a new state (e.g. component--highlighted)

**Use a state class**

- when a components element or modifier has altered due to a user interaction
- examples: `is-open`, `is-active`, `is-visible`

There isn't no right answer (as always) but seeing as you are adding to an existing modifier I recommend using a state class 😄

kyleharper: You can see this principle in the same component...`.is-active .avatar__caret` works better than `avatar__caret--active` 😄

PR #294, opened 2018-06-12
Issue/react16 fix warnings

AmeeMorris:
As the title says, some more may crop up, but these are the ones I've come across so far. Also fixed any prop type errors that I came across with our stories

### DIFF
--- a/lib/DropdownSwitcher/index.js
+++ b/lib/DropdownSwitcher/index.js
@@ -4,6 +4,7 @@ import cx from 'classnames';
 import Dropdown from 'react-bootstrap/lib/Dropdown';
 import Overlay from 'react-bootstrap/lib/Overlay';
 import Popover from 'react-bootstrap/lib/Popover';
+import CustomToggle from './CustomToggle';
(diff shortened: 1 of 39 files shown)


AmeeMorris:
[[React 16] Fix warnings around React 16](https://trello.com/c/h5zXUnex/3414-react-16-fix-warnings-around-react-16)

lib/DropdownSwitcher/index.js
```diff
@@ -51,9 +52,9 @@ class DropdownSwitcher extends Component {
   dropdown() {
     return (
       <Dropdown className="dropdown-switcher" id={this.props.id}>
-        <div bsRole="toggle">
+        <CustomToggle bsRole="toggle">
```
kyleharper: What's the story with this new `CustomToggle`?

AmeeMorris: Ah its a work around for a warning. Having bsRole on a div threw up a warning, the whole new dom attributes thing complained about it all not being lower case, but bootstrap needs it.

PR #295, opened 2018-06-12
added tooltips to file cards

kyleharper:
This PR adds tooltips to FileCards...I had an issue when adding BoundaryClickWatchers to the mix but that was a hang up from bootstrap & our own styling.

This PR sorts that too!


kyleharper:
[Revisions: Add viewing of 'Who' made changes with each revision](https://trello.com/c/uP89IlNz/3119-revisions-add-viewing-of-who-made-changes-with-each-revision)

PR #296, opened 2018-06-14
Dont pass the wrong prop to Avatar

AmeeMorris:
as the title says!


AmeeMorris:
[Bug: My avatar isn't being pulled into the assigned users components](https://trello.com/c/TpmMFHrp/3128-bug-my-avatar-isnt-being-pulled-into-the-assigned-users-components)

PR #298, opened 2018-06-15
Only highlight mentions that exist in the users prop

AmeeMorris:
As the title says, slightly smarter mention highlighting!


AmeeMorris:
[UX Issue: You can appear to @ someone who doesn't exist](https://trello.com/c/awuH7Dsw/3263-ux-issue-you-can-appear-to-someone-who-doesnt-exist)

PR #301, opened 2018-06-22
Add userCanSetDueDate prop to DueDatePicker

AmeeMorris:
And don't let users set a due date if that's false


AmeeMorris:
[BUG: Permission issues](https://trello.com/c/W9GjmIrP/3441-bug-permission-issues)

PR #302, opened 2018-06-22
userCanResolve prop

AmeeMorris:
adding this prop for some more control over users being able to resolve a conversation

### DIFF
--- a/lib/Conversation/index.js
+++ b/lib/Conversation/index.js
@@ -89,15 +89,16 @@ class Conversation extends Component {
                     <ResolveIcon /> Resolved
                   </span>
                 )}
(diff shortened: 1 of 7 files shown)


lib/Conversation/index.js
```diff
@@ -63,7 +65,8 @@ class Conversation extends Component {
     const showComments = this.state.showComments || this.props.showComments;
     const additionalClasses = cx(this.props.className, {
       'is-active': showComments,
-      'is-resolved': this.props.resolved
+      'is-non-interactive':
```
kyleharper: The user can still click on it to change state? Does this class name still make sense?

AmeeMorris:
@kyleharper good point! changed to is-read-only

PR #305, opened 2018-07-02
Read only ui for StatusIndicator

AmeeMorris:
Also a new icon and the ability to pass a className to Notification


lib/StatusIndicator/styles.scss
```diff
@@ -41,12 +41,17 @@ $status-indicator-margin-left: 1.3rem !default;
...
 
 .status-indicator__label {
-  text-overflow: ellipsis;
-  overflow: hidden;
   white-space: nowrap;
-  width: 100%;
   margin-top: 0;
   margin-bottom: 0;
+  width: 100%;
+  span.status-indicator__labelspan {
```
kyleharper: Does this class need to be attached to a span? Seeing as this covers text too maybe something like `status-indicator__label-text` reads well?

PR #313, opened 2018-07-09
add the delay props to TooltipWrapper

AmeeMorris:
As the title says really


AmeeMorris:
[UX Issue: Formatting options need hover over tooltips](https://trello.com/c/veXbrFan/3406-ux-issue-formatting-options-need-hover-over-tooltips)

PR #317, opened 2018-08-03
added modal title styles for images being added

kyleharper:


kyleharper:
[Bug - Letters show up three at a time when typing](https://trello.com/c/FlBEbrOb/3569-bug-letters-show-up-three-at-a-time-when-typing)

PR #322, opened 2018-08-10
Add a failure state to comments

AmeeMorris:
As the title says.

Comments can have a `hasFailed` property which will display a little failure message and retry button (which will call our brand new retryComment prop)


AmeeMorris:
[UX Issue: I thought I lost my comment](https://trello.com/c/95Kj7FBT/3452-ux-issue-i-thought-i-lost-my-comment)

PR #323, opened 2018-08-13
Fix clicking conversation reply count with boundary click watcher

AmeeMorris:
This fixes an issue where using Conversation with BoundaryClickWatcher wont work properly if the reply count is clicked to activate the conversation, directly after clicking the 'cancel' button to deactivate the conversation. 

We don't render the reply count if the conversation is active, so when the boundary watcher checks if event.target is inside the container, it doesn't think it is, because it wasn't or wont be rendered. Hiding it with some css instead achieves the same result, but without this weird bug!

### DIFF
--- a/lib/Conversation/index.js
+++ b/lib/Conversation/index.js
@@ -114,20 +114,16 @@ class Conversation extends Component {
             </div>
           )}
 
-        {!showComments &&
-          comments.length > 1 && (
-            <div className="conversation__reply-count">
-              <Button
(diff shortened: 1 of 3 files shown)


AmeeMorris:
[Bug: View reply link doesn't work on first click (sometimes)](https://trello.com/c/4zM6vNOw/3611-bug-view-reply-link-doesnt-work-on-first-click-sometimes)

PR #329, opened 2018-08-27
Dont show sub list unless required

kyleharper:
does what it says on the tin 😄


kyleharper:
[Bug: item list performance is affecting writing experience.](https://trello.com/c/Fxz7hc7I/3636-bug-item-list-performance-is-affecting-writing-experience)

PR #331, opened 2018-08-29
A few mainly stylistic changes to search

AmeeMorris:
Styles, missing props etc!

### DIFF
--- a/lib/Search/styles.scss
+++ b/lib/Search/styles.scss
@@ -1,3 +1,6 @@
+$top-bar-height: 95px;
+$search-min-width: 250px !default;
+
 .search {
   position: relative;
 }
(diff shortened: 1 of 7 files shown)


lib/Search/styles.scss
```diff
@@ -1,3 +1,6 @@
+$top-bar-height: 95px;
```
kyleharper: Hmm this seems like a leak from a consuming app. Search is a standalone component in `gather-ui` is there a way we can do this in an abstract manner?

kyleharper: One idea would be to create a story using both `TopBar` and `Search` together and then join the 2 with a new element like `top-bar__search-dropdown`.?

Then apply the styles to that - it'll then be clear that in our library we expect the 2 to be used together where as right now search can only ever be used in the top bar reducing it's usability to a single case.

AmeeMorris:
[[Search] Build UI components](https://trello.com/c/tnSC2dpb/3502-search-build-ui-components)

AmeeMorris:
@kyleharper yep, good thinking, changes done :)

PR #337, opened 2018-09-04
add MentionsForm component

AmeeMorris:
Adding a new MentionsForm component, so we can have a reusable `@ing` input. 

Its mostly copied from the current CommentForm components, but made a little more generic and with some more functionality (add mention button)

After some testing we can replace CommentForm with this 👍


lib/BoundaryClickWatcher/index.js
```diff
@@ -18,7 +19,8 @@ class BoundaryClickWatcher extends Component {
     onMouseEnter() {},
     onMouseLeave() {},
     className: '',
-    alwaysListen: false
+    alwaysListen: false,
+    ignoreClass: null
```
petemcfarlane: should this be an empty string `''` rather than `null` if it's PropType is `string`?

lib/Dropdown/index.js
```diff
@@ -19,12 +19,14 @@ class Dropdown extends Component {
...
-    className: PropTypes.string
+    className: PropTypes.string,
+    ignoreClass: PropTypes.string
   };
 
   static defaultProps = {
     onToggle: () => {},
-    className: ''
+    className: '',
+    ignoreClass: null
```
petemcfarlane: Again, should this be `''` not `null` to match the propType?

lib/MentionsForm/MentionsFormSearch.js
```diff
@@ -0,0 +1,62 @@
...
+        <Dropdown.ActionGroup>
+          <span className="mention-form__search-heading">Mention...</span>
+        </Dropdown.ActionGroup>
+        <Dropdown.ActionGroup className="mention-form__search-wrapper">
+          <input
+            className="mention-form__search-input"
+            onChange={this.searchForUsers}
+            type="text"
+            placeholder="Search people..."
+          />
```
petemcfarlane: Do you think it'd be nice if this input was autofocused?

lib/MentionsForm/MentionsFormSearch.js
```diff
@@ -0,0 +1,62 @@
...
+        </Dropdown.ActionGroup>
+        <Dropdown.ActionGroup className="mention-form__search-results">
+          {this.state.users && this.state.users.length > 0 && users}
+          {this.state.users &&
+            this.state.users.length === 0 && (
+              <span className="no-results">No users found!</span>
```
petemcfarlane: Could this be padded/centred to look a bit more consistent with the search results?

AmeeMorris:
@petemcfarlane good catch with the z-index, should be fixed now!

As for the keyboard controls, its a known issue that we need to put in for a few areas on the app. Arrow controls work when you type `@` since thats given to us for free with the package we use. But the mentions button is all our own custom stuff, so we need to build it in ourselves everywhere else, its kind of a separate issue which is known about!

lib/MentionsForm/MentionsFormSearch.js
```diff
@@ -0,0 +1,62 @@
...
+
+  render() {
+    const users = this.state.users.map(user => (
+      <Dropdown.Action
+        action={() => this.props.addMention(user.id, user.display)}
+        key={user.id}
+        isSubmit={false}
+      >
+        <Avatar url={user.avatar} initials={user.initials}>
+          <AvatarInformation email={user.email} name={user.name} />
```
petemcfarlane: Just realised we don't show the username here, is that a bug? And we don't show the email in the MentionsFormInput, only the username/name. They should be the same, right?

lib/MentionsForm/styles.scss
```diff
@@ -164,6 +164,10 @@ $mention-suggestions-max-height: 500px;
 
 }
 
+.mention-form__input__suggestions {
+  z-index: 2 !important;
```
petemcfarlane: ❗️ important

lib/BoundaryClickWatcher/index.js
AmeeMorris: Doesn't really matter, it'll just treat null as the prop not existing, and wont throw up any propType errors

lib/MentionsForm/MentionsFormSearch.js
AmeeMorris: Done! One thing to note is because the search box isnt re-rendered each time the dropdown is opened and closed, we couldn't use a simple `autoFocus` prop 👎

lib/MentionsForm/MentionsFormSearch.js
AmeeMorris: Thanks for spotting, I shoved that in then forgot about it, wonderfully styled up now!

lib/MentionsForm/MentionsFormSearch.js
AmeeMorris: Good point, I've added a prop which will determine whether both should display the email or username (although I'd imagine at some point we'll just end up displaying both ;))

lib/MentionsForm/styles.scss
AmeeMorris: Gross I know, the mentions package adds in some inline styles, so its the only way to override 🤢

PR #343, opened 2018-09-14
Feature/tag user component

AmeeMorris:
A few things here, probably looks a bit bigger than it actually is due to moving things about. If you want me to split it up or walk through it give me a shout.

Mainly we've got a new TagUserForm component, similar to the MentionForm, but users are added in a more 'tagging' style, rather than `@`ing. Because of similarities between the two I've moved MentionsFormSearch out to a new UserSearch component, so both can use it, and made some cleanup in MentionsForm too!

There's a new button style, a new icon, and some global styling changes that were requested for placeholders, and the slight font-size.

One thing to note is there's a couple of timeouts in TagUserForm. I've had to add those in due to a timing issue that occurs if its placed inside a Dropdown.


AmeeMorris:
[[Loop someone in] UI Components](https://trello.com/c/HFHw6LEG/3695-loop-someone-in-ui-components)

lib/Button/ButtonWithIcon.js
```diff
@@ -0,0 +1,27 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import { Button, Icon } from '../';
+
+const ButtonWithIcon = props => (
+  <div className="button-icon">
```
kyleharper: If this is part of the button component then maybe something like `button__wrapper button__wrapper--icon` may keep us inline with our BEM patterns.

PR #347, opened 2018-09-24
stop un-mounting sub lists

kyleharper:
A simple PR that stops un-mounting sub lists so we retain the open hierarchy until the users refreshes the page of course 😄


kyleharper:
[Issue: I can't see which item I'm on in the item list](https://trello.com/c/RXt0122x/3561-issue-i-cant-see-which-item-im-on-in-the-item-list)

PR #348, opened 2018-09-24
add wrapperClassName prop to TooltipWrapper

AmeeMorris:
As the title says, we already pass the className prop to the tooltip itself


AmeeMorris:
[UX Issue: Formatting Toolbar animation](https://trello.com/c/N97Yiv8K/2965-ux-issue-formatting-toolbar-animation)

PR #349, opened 2018-09-25
Feature/avatar animations

AmeeMorris:
Adding an animate prop to avatar, which will give it a lovely animation on render!


AmeeMorris:
[UX Issue: It's easy to miss when people connect to the item I'm on](https://trello.com/c/KM32rF60/2998-ux-issue-its-easy-to-miss-when-people-connect-to-the-item-im-on)

PR #350, opened 2018-09-27
Make guidelines max-height dynamic

AmeeMorris:
Pretty much as the title says. 

We're setting the max-height with js now, that way we can still have the sliding animation with dynamic heights.

### DIFF
--- a/lib/Guideline/styles.scss
+++ b/lib/Guideline/styles.scss
@@ -44,6 +44,7 @@ $guideline-title-size: $typo-size-large !default;
 
 .guideline.is-active {
   .guideline__body {
-    max-height: 400px;
+    max-height: 100%;
+    overflow-y: auto;
   }
 }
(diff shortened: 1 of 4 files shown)


AmeeMorris:
[Guidelines field content is being cut off when displayed on the page](https://trello.com/c/YEROWn2y/3731-guidelines-field-content-is-being-cut-off-when-displayed-on-the-page)

lib/Guideline/styles.scss
```diff
@@ -44,6 +44,7 @@ $guideline-title-size: $typo-size-large !default;
 
 .guideline.is-active {
   .guideline__body {
-    max-height: 400px;
+    max-height: 100%;
```
kyleharper: I thought this would have handled the animation without the need for JS but I gather it doesn't 😂

PR #362, opened 2018-10-19
Feature/card component

AmeeMorris:
May as well add this now! 

This adds a new Card component. 

I've disabled some linting in the main card file, this is because the linting was complaining the div didnt have a role, even though it did! I'm guessing its because the role I gave it was a ternary.


AmeeMorris:
[[Content] build <ProjectFiles />](https://trello.com/c/D8rdeKJ3/3778-content-build-projectfiles)

PR #363, opened 2018-10-23
Styling changes to CollectionsTable and Card

AmeeMorris:
Couple of styling tweaks, adding a selected style to Card and CollectionsRable and making FileCards work within CollectionsTable

### DIFF
--- a/lib/CollectionsTable/styles.scss
+++ b/lib/CollectionsTable/styles.scss
@@ -4,6 +4,7 @@
 $collections-table-spacing: $layout-spacing-base/2;
 $collections-table-content-padding: $layout-spacing-base/4 $layout-spacing-base/2;
 $collections-table-border: 1px solid $color-border-base;
+$collections-table-selected-border: 2px solid $primary-blue;
 $collections-table-border-radius: 4px;
 $collections-table-indent-level-support: 10;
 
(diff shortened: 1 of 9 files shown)


lib/CollectionsTable/styles.scss
```diff
@@ -48,31 +49,64 @@ $collections-table-indent-level-support: 10;
...
+    padding: $collections-table-content-padding;
+  }
   &:first-child {
     padding: 0;
     border: 0;
   }
 
   &:first-child > .collections-table__cell-content {
-    padding: $collections-table-spacing;
+    padding: ($collections-table-spacing + 1) $collections-table-spacing ($collections-table-spacing + 1) ($collections-table-spacing + 1);
```
kyleharper: Why do we need the + 1px?

lib/CollectionsTable/styles.scss
```diff
@@ -48,31 +49,64 @@ $collections-table-indent-level-support: 10;
...
   }
+
 }
 
 .collections-table__cell-content {
   display: flex;
   align-items: center;
   position: relative;
   min-width: 0;
+  .file-card {
```
kyleharper: When normally deal with component composition styling we apply a class like `collections-table__file-card`. This can be seen in this stylesheet e.g. `collections-table__status-indicator`.

I'm down for this way to styling components within components but I just want to know if we're doing this moving forward so I can follow that pattern 😄 

------

I guess it's easier to manage if we just use the already defined classes...thoughts?

lib/CollectionsTable/styles.scss
AmeeMorris: so it doesnt "jiggle" when adding the extra 1px border

lib/CollectionsTable/styles.scss
AmeeMorris: @kyleharper Ah see I think there's an argument for both. If its a component we include ourselves then adding that composition styling makes total sense. 

However if its just a component that can live inside there, but isn't really part of it by default, and its totally up to the user then this way makes more sense. What do you think?

AmeeMorris: By the latter scenario making more sense I mean that our components should just work with each other (within reason) without having to add extra classNames. 

Butttttttt I'm easy either way, just musings.

kyleharper: Exactly my thoughts 👍 😄

PR #364, opened 2018-10-26
Add a new ConfirmationDropdown and SelectionBar components

AmeeMorris:
Couple of simple components. The confirmation dropdown has been used a few times now, so it makes sense we have a ui component for it!

SelectionBar is pretty basic for now, we can make it a bit more sophisticated if needed


lib/SelectionBar/SelectionBarLeft.js
```diff
@@ -0,0 +1,28 @@
+import React, { Fragment } from 'react';
+import PropTypes from 'prop-types';
+import pluralize from 'pluralize';
+import { Button } from '../';
+
+const SelectionBarLeft = props => (
```
kyleharper: `SelectionBarLeft` applies that there is some logic or at least style hooks for the left position. It seems though that isn't the case. Should the name be more directed at the contents of the component?

lib/SelectionBar/SelectionBarRight.js
```diff
@@ -0,0 +1,21 @@
+import React, { Fragment } from 'react';
+import PropTypes from 'prop-types';
+import { ShortcutTrigger } from '../';
+
+const SelectionBarRight = props => (
```
kyleharper: See feedback about `SelectionBarLeft` 😄

AmeeMorris:
[[Content] build <ProjectFiles />](https://trello.com/c/D8rdeKJ3/3778-content-build-projectfiles)

PR #369, opened 2018-12-06
Feature/comment threading

AmeeMorris:
Adding a preview of the most recent reply to conversations.

### DIFF
--- a/lib/Conversation/styles.scss
+++ b/lib/Conversation/styles.scss
@@ -72,7 +72,7 @@ $conversation-meta-text-color: $neutral-base !default;
 }
 
 .conversation__reply-count {
-  padding: 0 0 $conversation-padding $layout-spacing-base*2;
+  padding-left: $layout-spacing-base*2;
   margin: -$conversation-padding 0 0;
   font-size: $conversation-meta-text-size;
   .is-active & {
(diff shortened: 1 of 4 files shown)


AmeeMorris:
[Issue: Comment threading](https://trello.com/c/c1vr4XEe/3562-issue-comment-threading)

lib/Conversation/styles.scss
```diff
@@ -216,25 +221,36 @@ $conversation-meta-text-color: $neutral-base !default;
...
+  margin-bottom: 0;
+}
+
+.is-active .conversation__comment,
+.show-reply-preview > .conversation__comment-list .conversation__comment {
   position: relative;
   &:before {
     content: ' ';
-    border-left: $conversation-border-width solid $conversation-border-color;
+    border-left: 2px solid $conversation-border-color;
```
kyleharper: This is double the previous value should we *2 the var rather than setting a fixed value? Or just alter the var to be `2px`?

AmeeMorris: Its because that var also controls the outer border of the conversation, this is just the little thread between the two. Before they could share the same value, but some feedback was to change it to 2px, obv I dont want to make all the borders 2px, just this guy.

AmeeMorris: > Its because that var also controls the outer border of the conversation, this is just the little thread between the two. Before they could share the same value, but some feedback was to change it to 2px, obv I dont want to make all the borders 2px, just this guy.

(I can 2* it, but it just seems excessive to me, esp since they may want to be handled differently)

PR #371, opened 2018-12-10
fix overdue due date styling

AmeeMorris:
Make sure its always red and the right font-weight


AmeeMorris:
[Issue: Due date colours](https://trello.com/c/2cRFxf05/3923-issue-due-date-colours)

PR #372, opened 2019-01-07
Multiline support for EditableTextWrapper

AmeeMorris:
And adding it to Field.

Also adding a prop to limit activating the edit state to just the pencil button (heard that idea mentioned a few times). Better prop name suggestions welcomed ;)


AmeeMorris:
[Issue: Typing a long field name in the structure editor doesn't wrap the text](https://trello.com/c/JKutr03C/3741-issue-typing-a-long-field-name-in-the-structure-editor-doesnt-wrap-the-text)

PR #374, opened 2019-01-14
Feature/form modal

petemcfarlane:
This adds a form modal - a form within a modal 😮 

There's a submit handler, you can set the state of the form to `formIsSubmitting` to decide whether to show the progress buttons spinner or not, or allow any more attempts of the form to be submitted (useful to prevent duplicate requests being made).


petemcfarlane:
@kyleharper @AmeeMorris Thanks for the feedback, but do you not think we need to add a test for FormModal too? I've added the failing test I have written in the last commit, maybe one of you can see what's wrong with it but I can't get it to pass. That's where I got stuck a week ago!

AmeeMorris:
@petemcfarlane I've fixed the tests. The modal doesn't actually render unless it has a `show` bool prop passed to it. Usually a ModalTrigger controls that for you

PR #375, opened 2019-01-22
FormModal should use onHide prop, which is passed from withModalTrigger

petemcfarlane:
Ok, I didn't realise we had the onHide prop which is usually passed from the parent `withModalTrigger`... seems to be quite hidden, but I guess I'm just not used to it. So if we want to close the form modal by clicking the "cancel" button, we need to pass it the parent - parent - onHide prop! This will in tern, pass the `show` prop down all the layers and set it to `false`. I've also removed the `cancelHandler` prop, I don't think we'll usually need this?

### DIFF
--- a/lib/FormModal/README.md
+++ b/lib/FormModal/README.md
@@ -3,15 +3,15 @@ A form within a modal
 
 ### Props
 
-| Name             | Type    | Default    | Required | Description                                       |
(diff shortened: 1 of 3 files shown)


petemcfarlane:
[[Security] Generating an API Key should require re-entering a password](https://trello.com/c/KwhNWPTR/3980-security-generating-an-api-key-should-require-re-entering-a-password)

PR #376, opened 2019-01-22
Feature/finder components

AmeeMorris:
Adding groundwork for some new finder navigation components - I don't want to add a story for these just yet since there's more to add and may well change.

Also lovely new icons.

Also also, I've gotten rid of the lint warning about the console log in icon since it was annoying me!


AmeeMorris:
[Initial finder panel styles & clean up](https://trello.com/c/50z0NoB2/4023-initial-finder-panel-styles-clean-up)

PR #380, opened 2019-02-01
Feature/flex collections

AmeeMorris:
A version of CollectionsTable that isnt actually a table.

I've kept the original, but not sure if there's a usecase for both?


AmeeMorris:
[Flex Collections & Item Drag guides](https://trello.com/c/8s92zY5t/4045-flex-collections-item-drag-guides)

PR #383, opened 2019-02-13
Add onBlur to Tabs.Form

AmeeMorris:
A new UI pattern we use is to submit small forms onBlur, this PR adds a `submitOnBlur` prop to Tabs.Form which allows us to do just that.

### DIFF
--- a/lib/Tabs/TabsForm.js
+++ b/lib/Tabs/TabsForm.js
@@ -5,12 +5,20 @@ import Form from '../Form';
 import Icon from '../Icon';
 import Button from '../Button';
 
-const TabForm = ({ onSubmit, onCancel, value, placeholder, onInputChange }) => {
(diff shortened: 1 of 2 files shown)


AmeeMorris:
[I create a new tab and then click save - however I don't click away from the tab. The new tab is not saved](https://trello.com/c/mqT3NJxm/4086-i-create-a-new-tab-and-then-click-save-however-i-dont-click-away-from-the-tab-the-new-tab-is-not-saved)

AmeeMorris:
[It’s not intuitive that you need to press Enter to create a new tab](https://trello.com/c/h2h7MfcK/4087-its-not-intuitive-that-you-need-to-press-enter-to-create-a-new-tab)

AmeeMorris:
actually we dont need this!

PR #384, opened 2019-02-13
Feature/onblur tabs form

AmeeMorris:
Lets try this again...

Adding the option to submit the Tabs.Form on blur. The previous closed PR just had the onBlur event on the input, so would submit the form when clicking the cancel button, or the padding around the input. Now its on the Form itself and we double check that we're not still clicking inside.

### DIFF
--- a/lib/Form/index.js
+++ b/lib/Form/index.js
@@ -22,10 +22,17 @@ class Form extends Component {
   };
 
   render() {
-    const { children, className } = this.props;
+    const {
+      children,
+      className,
+      onSubmit,
+      onCancel,
+      escToClose,
+      ...rest
+    } = this.props;
 
     return (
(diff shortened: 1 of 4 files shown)


AmeeMorris:
[It’s not intuitive that you need to press Enter to create a new tab](https://trello.com/c/h2h7MfcK/4087-its-not-intuitive-that-you-need-to-press-enter-to-create-a-new-tab)

AmeeMorris:
[I create a new tab and then click save - however I don't click away from the tab. The new tab is not saved](https://trello.com/c/mqT3NJxm/4086-i-create-a-new-tab-and-then-click-save-however-i-dont-click-away-from-the-tab-the-new-tab-is-not-saved)

PR #385, opened 2019-02-14
Allow a shortcut to only have one child element

petemcfarlane:
Use concat to always cast the children as an array


petemcfarlane:
[Shortcut listing for bullet indents using the "Tab" button.](https://trello.com/c/VP1Otj2m/4102-shortcut-listing-for-bullet-indents-using-the-tab-button)

AmeeMorris:
@petemcfarlane I'm just getting a proptype error `Failed prop type: Invalid prop `children` of type `object` supplied to `Shortcut`, expected an array. in Shortcut`. Sounds like a job for `PropTypes.oneOfType([])`

petemcfarlane:
Thanks @kyleharper @AmeeMorris, fixed now, I think

PR #387, opened 2019-02-15
PageInformation improvements

AmeeMorris:
Minor things, just adding a title tag and making sure that the PageInformations editable state doesn't go weirdly narrow


AmeeMorris:
[Item names are jannkkky](https://trello.com/c/k8pjKBSw/4089-item-names-are-jannkkky)

lib/PageInformation/index.js
```diff
@@ -16,12 +16,12 @@ const PageInformation = ({
         onChange={rename}
         title={`Rename ${contextName}`}
       >
-        <h1 className="page-information__title" alt={title}>
+        <h1 className="page-information__title" alt={title} title={title}>
```
kyleharper: Do we need the `alt` attribute?

PR #388, opened 2019-02-19
Fix overflow issues

kyleharper:
Our list component has a regression when we moved to `List` and `ItemRow`. When a title is too long the actions are pushed out of the container.


kyleharper:
[Long item names hides dropdown button](https://trello.com/c/sWoc98TN/4111-long-item-names-hides-dropdown-button)

PR #390, opened 2019-02-22
ConfirmationDropdown alterations

kyleharper:
To make the `ConfirmationDropdown` more re-usable I have changed it quite a bit;

- we now use the `Dropdown` components rather than re-creating logic we already have
- we allow consumers to define their own dropdown trigger (via props.children)
- we handle the promise process itself within the component showing a loader!

### DIFF
--- a/lib/Button/index.js
+++ b/lib/Button/index.js
@@ -18,12 +18,15 @@ class Button extends Component {
     disabled: false,
     className: '',
     isSubmit: false,
-    title: ''
+    title: '',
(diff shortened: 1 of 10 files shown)


kyleharper:
@AmeeMorris we only have 1 instance of `ConfirmationDropdown` in use at the moment and it isn't public. Inconsistency in design has some showing and some not.

We're going with no cancel button as you can cancel it by various other means like clicking outside the dropdown.

PR #397, opened 2019-03-04
Feature/new fonts

Cyapow:


styles/settings/_typography.scss
```diff
@@ -1,18 +1,18 @@
 /* ==========================================================================
    Stacks
    ========================================================================== */
-$typo-stack-base: 'Whitney SSm A', 'Whitney SSm B', sans-serif !default;
+$typo-stack-base: 'IBM Plex Sans', sans-serif !default;
```
kyleharper: Where is this being sourced from?

Cyapow: At the moment this comes from the app codebase. Do you think its worth moving the font files into gather-ui? We can then still include them for old accounts using the app

kyleharper: How does the library handle this currently? (check `/.storybook/config.js`).

------------

I don't think we need to build anything for including fonts elsewhere but the library itself should include fonts so when people work on it they're using the correct typeface 👍

Cyapow: It points to a typography URL but we will be self hosting IBM-plex. The reason I say we'd have to include it elsewhere is because not everyone has the new webapp css included on there account yet

Cyapow:
[Apply new fonts app wide (self hosted, feature flagged?)](https://trello.com/c/iRATSepv/4062-apply-new-fonts-app-wide-self-hosted-feature-flagged)

PR #404, opened 2019-03-12
Fix finder item styling

AmeeMorris:
Make sure they're all BEM-y

### DIFF
--- a/lib/FinderNavigation/styles.scss
+++ b/lib/FinderNavigation/styles.scss
@@ -32,7 +32,7 @@ $finder-indent-level-support: 10 !default;
 
   .button:hover path,
   .dropdown-gc.is-active path,
-  &.is-active .finder__item-link path {
+  &.is-active .finder__item--link path {
     fill: $primary-blue;
   }
 }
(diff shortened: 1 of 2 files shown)


lib/FinderNavigation/styles.scss
```diff
@@ -32,7 +32,7 @@ $finder-indent-level-support: 10 !default;
 
   .button:hover path,
   .dropdown-gc.is-active path,
-  &.is-active .finder__item-link path {
+  &.is-active .finder__item--link path {
```
kyleharper: I thought `.finder__item--link` was an element so `.finder__item-link` fits with that.

It seems to be applied to a link meaning that it is an element and not a modifier of an existing element within the component.

I recommend we retain `finder__item-link` and change our usage in the app to use this class over changing it in here to break BEM naming conventions.

AmeeMorris: Ah, i thought it was the other way round!

AmeeMorris:
fixed on both @kyleharper

PR #411, opened 2019-03-25
ItemRow & List changes

AmeeMorris:
Adding a comment count to the ItemRow, and a subtitle to List.

This also fixes a bug with List where it would error out if a child was unmounted

Also a new icon.


lib/ItemRow/index.js
```diff
@@ -20,6 +28,12 @@ const ItemRow = ({ children, label, indicator, participants, stacked }) => {
       {participants && (
         <span className="item-row__participants">{participants}</span>
       )}
+      {commentCount !== '0' && (
```
altaudio:
```suggestion
      {commentCount >= '0' && (
```

That way we cover ourselves if the comment count comes back as `-1` for whatever weird reason.

AmeeMorris: Sadly its a string rather than an int :(

altaudio: Good point! That won't work then XD

PR #413, opened 2019-03-26
Adds noIndentation prop to finder item

altaudio:
- Adds `noIndentation` prop to `<FinderItem />`
- This prop switches the main classname between 'finder__item' and 'finder__item-no-indentation'
- But when you look at the styles, they both apply the same thing, say WHUT!1!?

Take a look at this snippet: https://github.com/gathercontent/gather-ui/blob/master/lib/FinderNavigation/styles.scss#L184-L201

That bad boy basically means that the more parents with the classname `finder__item` a `<FinderItemContent />` component has, the more indented to the right it is (to signify folder structure)

By adding a prop to change the classname, but keep the same styles, we can stop a folder that is not shown from indenting it's children (used in "Archived Items")

Relates to: https://github.com/gathercontent/app/pull/1585

### DIFF
--- a/lib/FinderNavigation/FinderItem.js
+++ b/lib/FinderNavigation/FinderItem.js
@@ -8,13 +8,20 @@ const FinderItem = ({
   hoverSettings,
   selected,
   children,
+  noIndentation,
   ...rest
 }) => {
-  const classNames = cx(`finder__item ${className}`, {
-    'is-active': active,
-    'finder__item--hover-settings': hoverSettings,
-    'finder__item--selected': selected
-  });
+  const classNames = cx(
+    `${
+      noIndentation ? 'finder__item-no-indentation' : 'finder__item'
+    } ${className}`,
+    {
+      'is-active': active,
+      'finder__item--hover-settings': hoverSettings,
+      'finder__item--selected': selected
+    }
+  );
+
   return (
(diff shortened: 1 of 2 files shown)


lib/FinderNavigation/FinderItem.js
```diff
@@ -8,13 +8,20 @@ const FinderItem = ({
...
   ...rest
 }) => {
-  const classNames = cx(`finder__item ${className}`, {
-    'is-active': active,
-    'finder__item--hover-settings': hoverSettings,
-    'finder__item--selected': selected
-  });
+  const classNames = cx(
+    `${
+      noIndentation ? 'finder__item-no-indentation' : 'finder__item'
```
AmeeMorris: Stylistically I think I'd prefer this ternary split out to a const above, just to keep the classnames set up cleaner.

kyleharper: Following BEM what normally happens is `finder__item` sets a base and then you apply modifiers on top of that. Ideally `finder__item` would always be set and then you would add `finder__item-no-indentation` on top like so;

```
cx('finder__item', {
  noIndentation: 'finder__item-no-indentation;
})
```

Was there a reason why we didn't do that for this case?

PR #417, opened 2019-03-27
Fix issues with EditableTextWrappers in CollectionTables

AmeeMorris:
Some scss changes to fix the above ^ 

If the EditableTextWrapper contained a long bit of text the edit button would be pushed out of sight, also the input would be tiny compared to the cell width.


AmeeMorris:
[Content hub items table wont shrink](https://trello.com/c/qTwxWMXf/4274-content-hub-items-table-wont-shrink)

PR #424, opened 2019-04-05
Dropdown.Trigger useSelect prop

kyleharper:
We added the latest `Dropdown` components to be able to compose dropdowns at will in many forms. However, at times we still want to achieve the same look that is provided by `DropdownMenu`, especially the trigger.

This PR adds the ability to add a menu modifier to `<Dropdown.Trigger />` .

--------

I've called the prop `useSelect` as it is a pipe dream of mine to utilise native selects a little more in our `Dropdown` components for accessibility and to increase our progressive nature.

### DIFF
--- a/lib/Dropdown/DropdownTrigger.js
+++ b/lib/Dropdown/DropdownTrigger.js
@@ -1,6 +1,7 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
-import { Button } from '../index';
+import cx from 'classnames';
+import { Button, Icon } from '../index';
 import { GATHER_UI_DROPDOWN } from './index';
 
 class DropdownTrigger extends Component {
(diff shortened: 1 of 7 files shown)


lib/Dropdown/DropdownTrigger.js
```diff
@@ -16,27 +17,44 @@ class DropdownTrigger extends Component {
...
-        className="dropdown__trigger-wrapper"
-      >
-        <Button {...this.props} clickHandler={this.handleClick}>
-          {children}
-        </Button>
-      </div>
-    ) : (
+    const { children, useButton, useSelect, ...rest } = this.props;
+    const { showContent } = this.context[GATHER_UI_DROPDOWN];
+    const createRef = trigger => {
```
altaudio: Is there any difference here to using `React.createRef()`?

kyleharper: Aye I believe so. `React.createRef` creates an object which includes additional ref properties in it like `this.ref.current`. I just want to pass a callback which assigns the ref to the instance I am working with 👍

It was by accident that they were called the same and maybe the end up being the same thing but I prefer callbacks personally.

PR #442, opened 2019-04-18
Firefox style fix for PricingPlan component

kyleharper:
A very small fix for the `PricingPlan` styles. I've also removed the `PricingText` component because we don't handle internationalisation so our test env blows up.


kyleharper:
[Browser test upgrade page](https://trello.com/c/V6N2ESI9/43-browser-test-upgrade-page)

PR #443, opened 2019-04-19
Add current plan state for pricing plan

kyleharper:
We've added a current prop to the pricing plan. It highlights the plan and adds a box shadow 👍


kyleharper:
[Design "active" plan state](https://trello.com/c/3MkE4U8F/55-design-active-plan-state)

PR #444, opened 2019-04-22
Stop CollectionsTableRows from wiggling when selected

AmeeMorris:
EZPZ


AmeeMorris:
[Elements inside rows nudge inwards when selected](https://trello.com/c/CBrXF5eF/84-elements-inside-rows-nudge-inwards-when-selected)

PR #446, opened 2019-04-22
Make ConfirmationDropdown confirm button blue

AmeeMorris:
Oof you might need to tackle this beast of a pr in stages


AmeeMorris:
[Confirmation text button for bulk actions should be blue](https://trello.com/c/TRVpE8YZ/90-confirmation-text-button-for-bulk-actions-should-be-blue)

PR #447, opened 2019-04-23
Fix/content hub/increase due dates row font size

altaudio:
- Adds `triggerButtonSizeMedium` prop to `<DueDatePicker>` to set the font size of the date picker dropdown trigger


altaudio:
[Font Size for Due Dates Row should be 16px](https://trello.com/c/OG78vhl0/132-font-size-for-due-dates-row-should-be-16px)

lib/Button/index.js
```diff
@@ -32,7 +34,8 @@ class Button extends Component {
     disableOnClick: PropTypes.bool,
     className: PropTypes.string,
     title: PropTypes.string,
-    isSubmit: PropTypes.bool
+    isSubmit: PropTypes.bool,
+    medium: PropTypes.bool
```
AmeeMorris: instead of as a dedicated prop, why not just add it as a "type" like the rest of the other button styles?

kyleharper: +1

PR #450, opened 2019-04-24
Improve the UX for resolving

kyleharper:
Resolved conversations weren't interactive but the text was still blue 🤔 

This fixes that and makes the UI much clearer 👍


kyleharper:
[Resolved/Unresolved states](https://trello.com/c/wQicW8wQ/69-resolved-unresolved-states)

PR #451, opened 2019-04-25
Split out UserList to its own component

AmeeMorris:
This grew bigger than expected... a fair chunk is just moving / renaming

The main chunk of this work is splitting up what used to be UserList into two components
1. UserList - this is now just the list of users and the search box
2. UserListDropdown - this is what used to be UserList, but sometimes we don't want a dropdown now, so its its own thing (and just imports the new UserList)

I've added a disabled styles and a secondaryAction prop to ConfirmationDropdown, incase we want to render a different action rather than cancel.

Part of the work for splitting out UserList was using some Dropdown.Actions without their usual Dropdown parents, so i've added some defaults for the context that would usually be included.

A direction prop for DropdownTrigger to control which way the arrow icon displays.

maxRows prop to ExpandingTextArea, which should be self explanatory

I've also updated enzyme to 3.5.0 and ran `yarn upgrade enzyme enzyme-adapter-react-16` since the version we were on didn't support ref forwarding


lib/ConfirmationDropdown/ConfirmationDropdownContent.js
```diff
@@ -51,18 +52,19 @@ const ConfirmationDropdownContent = (
...
-          clickHandler={() => {
-            onHide();
-            onCancel();
-            setShowContent(false);
-          }}
-          className="confirmation-dropdown__cancel"
-        >
-          Cancel
-        </Button>
+        {secondaryAction || (
```
kyleharper: I thought the cancel button was consistent across all confirmation dropdowns. Secondary actions appeared in another place instead. Has this changed?

lib/Dropdown/DropdownAction.js
```diff
@@ -15,11 +15,15 @@ const DropdownAction = (
...
     selected,
     overflow,
-    disabled
+    disabled,
+    ...props
   },
   context
 ) => {
-  const { toggleShowContent, showContent } = context[GATHER_UI_DROPDOWN];
+  const {
```
kyleharper: How come this change was required? The context should always exist. Otherwise, the action cannot function as expected.

AmeeMorris: In UserList we don't render a wrapping Dropdown, only the groups and actions

lib/ConfirmationDropdown/ConfirmationDropdownContent.js
AmeeMorris: See https://projects.invisionapp.com/share/W5PL9F9SX4M#/screens/339318859 how we have a secondary action instead of a cancel button, hopefully that clears things up

PR #453, opened 2019-04-25
Make sure we clear the selected users on toggle

AmeeMorris:
Just realised the component should do this

There's always one little thing I end up missing out!


kyleharper:
Should we maybe add a test to cover this?

PR #461, opened 2019-05-02
fix button backgrounds on safari

AmeeMorris:
As the title says, this should stop this:

Screen Shot 2019-05-02 at 10 24 14


AmeeMorris:
[Are the buttons supposed to look like this?](https://trello.com/c/hZGHcfoF/152-are-the-buttons-supposed-to-look-like-this)

PR #472, opened 2019-05-23
Feature/InputWithButton (Copy To Clipboard)

altaudio:
May-23-2019 12-29-40

- Creates a component that displays a text input with a button appended to it.
- The button onClick callback passes the input element as an argument, so you can manipulate it (eg. copy to clipboard).

### DIFF
--- a/lib/Button/styles.scss
+++ b/lib/Button/styles.scss
@@ -366,3 +366,11 @@ $btn-border-radius: 3px !default;
 .button--medium {
   font-size: $typo-size-base;
 }
+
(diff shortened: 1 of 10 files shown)


DavidHollins6:
I have a feeling (Not certain, even from some quick reading) that this input will need some sort of label, and I _think_ just popping an `aria-labelledby` attribute on it and pointing it towards the button's id should be sufficient.

altaudio:
@DavidHollins6 I will have to check both of those, thanks for flagging 👍

PR #473, opened 2019-05-23
CheckToggle will now also look at its props to see whether it is checked

DavidHollins6:
The checked prop was only used in the constructor, meaning that when the prop changes the checked state doesn't reflect it.

### DIFF
--- a/lib/CheckToggle/README.md
+++ b/lib/CheckToggle/README.md
@@ -3,26 +3,6 @@ Represents a toggle, can be true or false.
 
 ## Usage
 
-
-    className: PropTypes.string,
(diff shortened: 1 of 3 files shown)


altaudio:
I wonder if there's a slightly different way of going about this.

Could we potentially give the `CheckToggle` component a `value` prop, which could be set to `true/false`. If this prop is set then the input becomes controlled.

We could then also give it a `defaultValue` prop, which is set in the constructor.

I'll let you be the judge, not sure if either way is better than the other. Also it may mean changing the `CheckToggle` useage everywhere in the App.

Either way, could you add the new prop to the README for the component?

PR #479, opened 2019-05-31
Fix ImageLoader animations

kyleharper:
`<ImageLoader />` used to animate between loading and showing. It was broken between apply feedback so I've cleared up the conditional logic for styles and made the styles apply to the container.


kyleharper:
[Gather-ui ImageLoader animation fixes](https://trello.com/c/nu3d66EW/134-gather-ui-imageloader-animation-fixes)

PR #481, opened 2019-06-04
Small Fixes for the Shareable Link Generation UI

altaudio:
### CheckToggle
Adds:
- `disabled` prop (disables toggle)
- `labelSizeLarge` prop (Displays the label at a large size)
- `marginSizeLarge` (Adds a large margin to top and bottom of the toggle)
- `spaceBetween` prop (toggle and label spread to fill container (create space between)

### FormGroup
Creates `FormGroup` component to group elements within a form (make them inline)

### FormInput
Adds:
- `paddingSmall` prop (Gives the input small padding)
- `fullWidth ` prop (Gives the input 100% width)

### Form
Adds:
- `disabled` prop (Disables the form.)


assets/icons/link-white.svg
```diff
@@ -0,0 +1,5 @@
+<svg width="12" height="12" viewBox="0 0 12 12" fill="none" xmlns="http://www.w3.org/2000/svg">
```
AmeeMorris: Is this just the same link icon but in white? If so instead of adding a new icon, why not just make a new utility class:
```
.icon-white path {
     fill: white;
}
```

if its a different icon all together then just ignore me!

lib/CheckToggle/index.js
```diff
@@ -48,21 +56,36 @@ class CheckToggle extends Component {
       id,
       className,
       displaySmall,
-      displayChecked
+      displayChecked,
+      spaceBetween,
```
AmeeMorris: another utility class idea...
instead of adding this as a prop for CheckToggle would it make sense to have a `.space-between` class we can just add where ever?

lib/CheckToggle/styles.scss
```diff
@@ -57,8 +70,8 @@ $borderColour: #F8FBFC;
 /*  New switch */
 .toggle-switch + .toggle-switch__label {
   padding: 2px;
-  width: 53px;
-  height: 33px;
+  width: 34px;
```
AmeeMorris: are we ok to change the width and height for these everywhere?

lib/Form/FormGroup.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const FormGroup = ({ children }) => (
```
AmeeMorris: 👍

styles/helpers/_text.scss
```diff
@@ -16,6 +16,10 @@
   }
 }
 
+.neutral-base-text {
```
AmeeMorris: 👍

lib/CheckToggle/styles.scss
altaudio: 👍 Checked everywhere that the CheckToggle is used and non were using the default size (the default was MASSIVE 😱)

altaudio: 👍 Checked this and we aren't using the default size anywhere. It was MASSIVE 😱

PR #483, opened 2019-06-10
Adds PR Template

altaudio:
- Modifies app PR template to suit gather-ui

### DIFF
--- a/pull_request_template.md
+++ b/pull_request_template.md
@@ -1,0 +1,19 @@
+## [PR Title](link_to_trello_card)
+### 👀 Overview
+_A short description of what the changes in the PR do._
+### 💬 Description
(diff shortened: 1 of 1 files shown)


pull_request_template.md
```diff
@@ -0,0 +1,19 @@
+## [PR Title](link_to_trello_card)
```
kyleharper: Is this any different to the title on the PR when creating it in github?

kyleharper: What Trello cards do we show externally?

altaudio: Good point on the title, don't think it is really necessary. 

It was kind of just to attach a link to the related Trello issue to, but we can probably instead add a "Related Links" title where we can put designs, Trello cards etc.

PR #485, opened 2019-06-11
MentionsForm changes

AmeeMorris:
A few things here:

1. A new `@` icon
2. An icon choice for `UserSearchDropdown`
3. Functionality to have default mentions in `MentionForm` 

Added a comment to some squirrelly code in `MentionsFormInput`


lib/MentionsForm/MentionsFormInput.js
```diff
@@ -33,8 +37,20 @@ class MentionsFormInput extends Component {
...
   componentDidMount() {
     this.setState({ inputHeight: this.getInputHeight() }); // eslint-disable-line react/no-did-mount-set-state
+    this.addDefaultUsers();
   }
 
+  addDefaultUsers = async () => {
+    const { defaultUsers } = this.props;
+    if (defaultUsers.length) {
+      defaultUsers.map(user =>
+        setTimeout(() => {
```
AmeeMorris: Open to other suggestions to get around this... 

I came across a bit of a strange race condition with this:
Component would mount -> map through and set our default users -> UpdateInputValue triggers -> map finishes before state can ever update here https://github.com/gathercontent/gather-ui/pull/485/files#diff-4d3f51f4b10ea07b6defa87007b6bc22R90 -> each default user overrides the previous because the input value hasnt yet updated from the previous default users.

AmeeMorris:
[Make workflow notify form match comment @ing](https://trello.com/c/RprcTVwO/241-make-workflow-notify-form-match-comment-ing)

lib/MentionsForm/MentionsFormInput.js
altaudio: This is a really tough one!

The only thing I can think of is to see whether the `react-mentions` has the option to add multiple mentions in one action. This way we would only update the input and the state once.

Either that, or `react-mention` may have the option to set the default users?

The other option would be to make the input uncontrolled, but that would be a MASSIVE refactor.

lib/MentionsForm/index.js
```diff
@@ -82,10 +86,10 @@ class MentionsForm extends Component {
     this.setState({ inputValue: '' });
   };
 
-  updateInputValue = e => {
-    this.props.onInputChange(e.target.value);
-    this.setState({ inputValue: e.target.value });
-  };
+  updateInputValue = e =>
+    this.setState({ inputValue: e.target.value }, () =>
+      this.props.onInputChange(e.target.value)
```
altaudio: Why do we need to do this in the `setState` callback?

AmeeMorris: I think to ensure that we only call `onInputChange` after the value has actually changed

AmeeMorris:
@altaudio 

> The only thing I can think of is to see whether the react-mentions has the option to add multiple mentions in one action. This way we would only update the input and the state once. 

Yeah looked into this and couldn't find anything online :( 

> Either that, or react-mention may have the option to set the default users?

Or this :( 

> The other option would be to make the input uncontrolled, but that would be a MASSIVE refactor.

Eeek yeah not sure about this right now :D

altaudio:
> @altaudio
> 
> > The only thing I can think of is to see whether the react-mentions has the option to add multiple mentions in one action. This way we would only update the input and the state once.
> 
> Yeah looked into this and couldn't find anything online :(
> 
> > Either that, or react-mention may have the option to set the default users?
> 
> Or this :(
> 
> > The other option would be to make the input uncontrolled, but that would be a MASSIVE refactor.
> 
> Eeek yeah not sure about this right now :D

@AmeeMorris In that case I think we have to let it slide and add it to the debt board :+1:

lib/MentionsForm/index.js
altaudio: Why do we need to do that?

AmeeMorris: Its probably not essential, its just 'safer'. If you're triggering something after a value has changed you're making sure that thats actually the case. Since we're just passing the event value its even less needed though. I think I just added it in as a long shot help with my race condition business.

Just curious, is there a downside to doing this?

altaudio: Only downside I can see is extra code and that may not be doing anything and potential time spent for the next dev trying to figure out why it's like this (not usual pattern I don't think).

I will leave to you though, I'm happy you have explained the reasoning behind it :+1:

AmeeMorris: Ah ok cool, I've taken it out since there's no real reason for it to be there... For some reason I had it in my mind that it was liked that before I touched it in this PR

PR #487, opened 2019-06-13
Rename FileCard to Figure

DavidHollins6:
### 👀 Overview
When working on Inline Images for a text field, we noticed that the FileCard had most of the functionality we needed for an image. However, semantically it seems like it should be a figure, and the name `FileCard` restricts it's usability.

### 💬 Description
The wrapper `div` has been changed to a `figure`, the `div` displaying the image is now an `img` and the `div` displaying the label is now a `figcaption` element. I have renamed every instance of `FileCard` to now be `Figure`.

I have added an `altText` prop as well for future use.

### 🚪 Start Points
`Figure/index.js`

### 🚪 Related PRs (optional)
I'm just beginning work on refactoring app to use Figure, I will attach that PR when it's done.

 ### ✅ Checklist
- [x] Tests written
- [x] Storybook tested
- [ ] Added to documentation


lib/Figure/FigurePreview.js
```diff
@@ -12,53 +12,54 @@ const FileCardPreview = ({
...
     return (
-      <div className="file-card__title">
-        <Icon name="loader" className="file-card__loader" />
+      <div className="figure__title">
+        <Icon name="loader" className="figure__loader" />
       </div>
     );
   }
-
+  // TODO: Add alt-text prop
```
kyleharper: Is this being done before merging in yeah?

DavidHollins6: Ah this was actually done, just forgot to remove the TODO

DavidHollins6: Turns it it wasn't, it is now!

DavidHollins6:
> Only 1 piece of feedback. Also, do we set the image width to fill figures?

Yep

PR #488, opened 2019-06-13
MentionForm feedback

AmeeMorris:
Little change so we have the option to display the amount of mentions on the mention form

### DIFF
--- a/lib/MentionsForm/MentionsFormActions.js
+++ b/lib/MentionsForm/MentionsFormActions.js
@@ -2,6 +2,7 @@ import React from 'react';
 import cx from 'classnames';
 import PropTypes from 'prop-types';
 import Button from '../Button';
+import { pluarisePerson } from '../helpers';
 
 const MentionsFormActions = props => {
   const additionalClasses = cx({
@@ -17,12 +18,8 @@ const MentionsFormActions = props => {
     ? ['collapse', 'link-default']
     : ['primary'];
 
-  const peopleText =
-    props.mentionCount === 1
-      ? `${props.mentionCount} person`
-      : `${props.mentionCount} people`;
   const submitText = props.displayMentionCount
-    ? `${props.submitText} & notify ${peopleText}`
+    ? `${props.submitText} & notify ${pluarisePerson(props.mentionCount)}`
     : props.submitText;
 
   return (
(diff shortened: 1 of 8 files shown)


AmeeMorris:
[Make workflow notify form match comment @ing](https://trello.com/c/RprcTVwO/241-make-workflow-notify-form-match-comment-ing)

lib/MentionsForm/MentionsFormActions.js
```diff
@@ -17,6 +17,14 @@ const MentionsFormActions = props => {
     ? ['collapse', 'link-default']
     : ['primary'];
 
+  const peopleText =
+    props.mentionCount === 1
```
altaudio: We could have a helper function that either gives us back `person/people` based on a number input

lib/MentionsForm/index.js
```diff
@@ -87,8 +89,12 @@ class MentionsForm extends Component {
   };
 
   updateInputValue = e => {
-    this.setState({ inputValue: e.target.value });
-    this.props.onInputChange(e.target.value);
+    const { value } = e.target;
+    const mentionCount = (value.match(pattern) || []).concat(
```
altaudio: I imagine there is probably a way to combine the two regexes here with an OR operator, so we only need to do the `value.match` once?

lib/MentionsForm/index.js
```diff
@@ -87,8 +89,12 @@ class MentionsForm extends Component {
...
+      value.match(/(\B@\w+)+/gi) || []
```
altaudio: We could _maybe_ create a const file for regex patterns that are re-useable. I'll let you decide on if that's a good idea or not, I'm on the fence.

tests/MentionsForm/MentionsForm.test.js
```diff
@@ -117,6 +117,14 @@ describe('Mentions Form', () => {
...
 
   test('updates the input value', () => {
+    expect(wrapper.state('mentionCount')).toEqual(0);
+    wrapper
+      .instance()
+      .updateInputValue({ target: { value: '@waffle @[plop]' } });
+    expect(wrapper.state('mentionCount')).toEqual(2);
+  });
+
+  test('updates the mentionCount', () => {
```
altaudio: Is this the description for the above test?

lib/MentionsForm/MentionsFormActions.js
AmeeMorris: Yeah this is a tricky one, the helper exists in app, but i wasn't sure wether to add it here too. I think it makes sense to make a helpers file though

lib/MentionsForm/index.js
AmeeMorris: There's a consts file, so I can just plop it in there for now

PR #490, opened 2019-06-20
Updates/choose editor font updates

altaudio:
### 👀 Overview
Minor updates that go along with the PR for changing the editor font in accessibility settings
### 💬 Description
- Adds a `buttonClassName` prop to the `<DropdownMenu />` so that its trigger button can styled (eg. with a different font).
- Allows a `className` to be passed as part of a item object passed to `<DropdownMenu />`, so each item can be styled individually (eg. with a different font).
- Adds a `blackText` helper class
### 🚪 Start Points
- `lib/DropdownMenu/index.js`
- `styles/helpers/_text.scss`
### 👫 Related PRs (optional)
https://github.com/gathercontent/app/pull/1839

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook


lib/DropdownMenu/index.js
```diff
@@ -34,12 +35,14 @@ class DropdownMenu extends Component {
     direction: 'down',
     shouldDisplay: true,
     fullWidth: false,
+    buttonClassName: null,
```
AmeeMorris: i think this might have to be changed to `''`, I swear I've seen it try to use 'null' as a class before (i could be remembering a really boring dream though)

AmeeMorris:
Hmm I just realised that this is using DropdownMenu, gonna add some comments about this on your other PR. bare with!

PR #491, opened 2019-06-20
Fix/figure image stretch

DavidHollins6:
### 👀 Overview
This is to fix a bug where file previews would show a stretched image instead of it zoomed in to cover.

This came as a result of a refactoring to a `figure`, and I changed the `div` with `background-image` to an `img` tag... causing the stretching!

### 💬 Description
To keep everything semantic and accessible, we kept the `img` tag and set it to visually hidden. Which is a class I've added, anything that is `visually-hidden` will be picked up by screen readers but not show on the page. I have then made the `div` visible but `aria-hidden`, so that does not come up on the screen reader as it's not very semantic.

### 🔗 Links
https://trello.com/c/sghpxJ8e/165-asset-thumbnails-are-distorted-on-items-score-8

### ✅ Checklist
- [x] Browser tested

### DIFF
--- a/lib/Figure/FigurePreview.js
+++ b/lib/Figure/FigurePreview.js
@@ -49,7 +49,7 @@ const FigurePreview = ({
       <div
         className="figure__thumbnail"
         title={caption}
-        style={{ backgroundImage: `url("${previewSrc}` }}
+        style={{ backgroundImage: `url("${previewSrc}")` }}
         alt={altText}
         aria-hidden="true"
       />
--- a/lib/Figure/FigurePreview.js
+++ b/lib/Figure/FigurePreview.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { Fragment } from 'react';
 import PropTypes from 'prop-types';
 import { Icon } from '../index';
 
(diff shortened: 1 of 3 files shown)


styles/generic/_global.scss
```diff
@@ -16,3 +16,11 @@ body {
         @include applyFontSmoothing();
     }
 }
+
+.visually-hidden {
```
DavidHollins6: This was taken from the tinterwebs but is a common pattern

lib/Figure/FigurePreview.js
```diff
@@ -39,12 +39,21 @@ const FigurePreview = ({
...
-      className="figure__thumbnail"
-    />
+    <Fragment>
+      <img
+        title={caption}
+        src={previewSrc}
+        alt={altText}
+        className="visually-hidden"
+      />
+      <div
```
kyleharper: Is the image visually hidden all the time? Is this component ever used with the image being it's native size?

styles/generic/_global.scss
```diff
@@ -16,3 +16,11 @@ body {
...
+    position: absolute !important;
+    height: 1px; width: 1px;
+    overflow: hidden;
+    clip: rect(1px 1px 1px 1px); /* IE6, IE7 */
```
kyleharper: IE6 + 7 😱 - we only support IE11+

lib/Figure/FigurePreview.js
DavidHollins6: Yeah the `img` tag is always visually hidden. I am unsure if this component was ever used for the image's native size. But I am not sure it matters because it is using a div like it used to, so that would be a separate bug from before.

PR #499, opened 2019-07-02
Add error states to FormInputs

AmeeMorris:
### 👀 Overview
This makes the inputs look all error-y and optionally display an error message depending on props.
### 📹 GIF (optional)
Screen Shot 2019-07-02 at 12 14 18
### 🚪 Start Points
`lib/Form/FormInput.js`
### 👫 Related PRs (optional)
coming soon...

### ✅ Checklist
- [x] Tests written
- [ ] Browser tested
- [x] Added to documentation
- [ ] Added to storybook

### DIFF
--- a/lib/Form/FormInput.js
+++ b/lib/Form/FormInput.js
@@ -77,10 +77,13 @@ class FormInput extends Component {
           className={classNames}
           placeholder={placeholder}
           autoFocus={this.props.focusOnMount} // eslint-disable-line jsx-a11y/no-autofocus
+          aria-invalid={hasError}
           {...rest}
         />
         {hasError && errorMessage && (
-          <span className="form-input__error-message">{errorMessage}</span>
+          <span role="alert" className="form-input__error-message">
+            {errorMessage}
+          </span>
         )}
       </Fragment>
     );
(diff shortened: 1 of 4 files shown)


lib/Form/FormInput.js
```diff
@@ -50,26 +54,35 @@ class FormInput extends Component {
...
+          type={type}
+          value={inputValue}
+          onChange={this.handleOnChange}
+          className={classNames}
+          placeholder={placeholder}
+          autoFocus={this.props.focusOnMount} // eslint-disable-line jsx-a11y/no-autofocus
+          {...rest}
+        />
+        {hasError && errorMessage && (
+          <span className="form-input__error-message">{errorMessage}</span>
```
DavidHollins6: When this validation message appears, users with screen readers won't be notified, so they won't know there is an error to be fixed! To make this happen, you can add the attribute `role="alert"` which will screen read whatever is inside the span when it appears.

lib/Form/FormInput.js
```diff
@@ -50,26 +54,35 @@ class FormInput extends Component {
...
-      />
+      <Fragment>
+        <input
+          type={type}
+          value={inputValue}
+          onChange={this.handleOnChange}
+          className={classNames}
+          placeholder={placeholder}
+          autoFocus={this.props.focusOnMount} // eslint-disable-line jsx-a11y/no-autofocus
+          {...rest}
```
DavidHollins6: When a screen reader tabs through this input, it has no way of knowing whether it is valid or not, an easy fix to this is to use the attribute `aria-invalid` and passing it true or false depending on whether it is valid

PR #506, opened 2019-07-10
Add some default FormInput styles

AmeeMorris:
### 👀 Overview
A couple of other form specific styles, and I've tweaked InputWithButton to have a small and large version.
### 💬 Description
 I could go way further down this rabbit hole, but i think its probably best to have a cross guild chat to resolve most of the basic element core styling issues that we have!

### ✅ Checklist
- [x] Tests written
- [ ] Browser tested
- [x] Added to documentation
- [x] Added to storybook

### DIFF
--- a/lib/Form/styles.scss
+++ b/lib/Form/styles.scss
@@ -94,16 +94,24 @@ $form-checkbox-label-size: $typo-size-lead !default;
 .form-radiobutton--invisible {
   visibility: hidden;
 }
-
+form .dropdown-gc .button--outline,
 .form__input {
   border: 1px solid lighten($neutral-base, 20%);
-  border-radius: 3px;
+  border-radius: 4px;
   font-size: $typo-size-base;
   padding: $layout-spacing-base/2;
   transition: $animation-time-micro border-color $animation-curve;
-
-  &:focus {
+  color: $neutral-dark;
+  font-size: $typo-size-base;
+  line-height: $typo-line-height-base;
+  &:hover {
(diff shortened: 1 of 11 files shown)


lib/Form/styles.scss
```diff
@@ -94,16 +94,24 @@ $form-checkbox-label-size: $typo-size-lead !default;
 .form-radiobutton--invisible {
   visibility: hidden;
 }
-
+form .dropdown-gc .button--outline,
```
kyleharper: Should this be `.form`? Also, why do we need to target the dropdown module?

lib/InputWithButton/index.js
```diff
@@ -23,13 +23,18 @@ class InputWithButton extends Component {
...
+      disabled,
+      paddingSmall
     } = this.props;
 
     const { hasBeenClicked } = this.state;
 
     return (
-      <div className="input-with-button">
+      <div
+        className={cx('input-with-button', {
```
kyleharper: Is it worth assigning this in a `classNames` const?

lib/Form/styles.scss
AmeeMorris: our form component does not have that class...

& this is why: 
Screen Shot 2019-07-10 at 16 30 53 now the dropdowns look like the other inputs

lib/InputWithButton/index.js
AmeeMorris: Since its just one condition I dont really see a problem with it, but happy to change it

PR #508, opened 2019-07-10
Add an initialValue prop to SearchInput

petemcfarlane:
### 👀 Overview
Allows us to set an initial value to a `SearchInput`, with an `initialValue` prop.

### 💬 Description
This is useful for when we want to profile the search input based on the url filters.

### 🚪 Start Points
I spent quite a while trying to workout why the component wasn't working in the storybook, then I realised we have a `SearchInput` component and a `Search\SearchInput` component - do we need 2?! Also, the `SearchInput` const is called `SearchInputField` inside the index.js file, so it's even more confusing for IDEs!!! 

### 👫 Related PRs (optional)

### ✅ Checklist
- [x] Tests written
- [x] Browser tested
- [ ] Added to documentation
- [x] Added to storybook

### DIFF
--- a/tests/SearchInput.js
+++ b/tests/SearchInput.js
@@ -6,7 +6,6 @@ describe('SearchInput', () => {
 
   test('should render an input component', () => {
     props = {
-      value: 'Botão',
       type: 'primary',
       onChangeHandler() {}
     };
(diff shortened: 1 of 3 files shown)


petemcfarlane:
Also, I can't see evidence that the test even ran in the CI...

tests/SearchInput.js
```diff
@@ -18,6 +17,19 @@ describe('SearchInput', () => {
...
   });
 
+  test('can have an initial value', () => {
+    props = {
+      initialValue: 'foo bar',
+      type: 'primary',
+      onChangeHandler() {}
+    };
+
+    const Element = shallow(<SearchInput {...props} />);
```
DavidHollins6: For future notice, we usually call the return value a `wrapper`.

PR #513, opened 2019-07-15
Refactor SearchInput to no longer be controlled

DavidHollins6:
### 👀 Overview
A lot of work has been done around filtering in the app, as a result there has been a need to control the Search Input, either to clear it or update it from the app.
### 💬 Description
I have removed the local state and exposed it's API. It is now going to be controlled through the props instead.

**This will be a minor version update.**

### ✅ Checklist
- [x] Tests written
- [x] Browser tested
- [ ] Added to documentation
- [x] Added to storybook

### DIFF
--- a/lib/SearchInput/index.js
+++ b/lib/SearchInput/index.js
@@ -1,79 +1,54 @@
-import React, { Component } from 'react';
+import React from 'react';
 import PropTypes from 'prop-types';
 import { FontAwesomeIcon } from '../index';
 
-class SearchInputField extends Component {
-  static defaultProps = {
-    placeholder: 'Search...',
-    hideIcon: false,
-    initialValue: ''
-  };
+const SearchInputField = ({
+  placeholder,
+  onChangeHandler,
+  onClearHandler,
+  hideIcon,
+  value
+}) => {
+  const hasSearch = value ? 'is-visible' : '';
 
-  static propTypes = {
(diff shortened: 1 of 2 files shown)


DavidHollins6:
> Nice, but do we not need to update the test / storybook also?

So all the tests passed when I ran `yarn test`, and they pass in the CI. But when I run them through the webstorm test runner the `SearchInput` ones run and fail...

This is similar to the issue we were having last week, I'm close to renaming this whole component & tests

petemcfarlane:
@DavidHollins6 that's disturbing, I wonder how long that's been happening. Do the others know about that also? CC @altaudio @kyleharper @AmeeMorris 

@amee It does introduce breaking changes, hence the minor version, but I think @DavidHollins6 will fix those in the app with the release 👍

kyleharper:
@DavidHollins6 the UI is used;

- account dashboard
- old content hub release archive items (turn your flag off to see)
- new content hub

The new content hub you have sorted (with the work you mentioned) but what about the others? 🤔

DavidHollins6:
> @DavidHollins6 the UI is used;
> 
> * account dashboard
> * old content hub release archive items (turn your flag off to see)
> * new content hub
> 
> The new content hub you have sorted (with the work you mentioned) but what about the others? 🤔

This PR should have all of that in: https://github.com/gathercontent/app/pull/1925

PR #514, opened 2019-07-16
Fix padding for noBorder FormInputs

AmeeMorris:
### 👀 Overview
A real baby pr. Padding styles for `.form__input--noborder` were being overridden by the default focus styles


AmeeMorris:
[Reduce indentation on finder panel](https://trello.com/c/G89w1N6G/367-reduce-indentation-on-finder-panel)

PR #521, opened 2019-07-23
 Pill Component

DavidHollins6:
### 💬 Description
A Pill component has been created to be used in the filter redesign. 

### 🔗 Links
https://app.goabstract.com/share/0deb63c9-cb4b-4077-9c41-a9d312b3f17b?collectionLayerId=6b7fbfe7-6a55-48d1-a30d-a187c7a4134c&present=true

### 🚪 Start Points
`lib/Pill/index.js`

### ✅ Checklist
- [x] Tests written
- [x] Browser tested
- [x] Added to documentation
- [x] Added to storybook

### DIFF
--- a/lib/Pill/index.js
+++ b/lib/Pill/index.js
@@ -1,0 +1,33 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Icon from '../Icon';
+
+const Pill = props => {
+  return (
+    <span className={`pill ${props.className}`}>
+      {props.children}
+      {props.onClearClick ? (
+        <button
+          type="button"
+          onClick={props.onClearClick}
+          className="button--clear button--icon-only-padding"
+        >
+          <Icon name="cross" />
+        </button>
+      ) : null}
+    </span>
+  );
+};
+
+Pill.propTypes = {
+  children: PropTypes.node.isRequired,
+  onClearClick: PropTypes.func,
+  className: PropTypes.string
+};
+
+Pill.defaultProps = {
+  onClearClick: null,
+  className: ''
+};
+
+export default Pill;
--- a/lib/Pill/index.js
+++ b/lib/Pill/index.js
@@ -1,32 +1,20 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import Icon from '../Icon';
 
-const Pill = props => {
+const Pill = ({ className, children, ...rest }) => {
   return (
-    <span className={`pill ${props.className}`}>
-      {props.children}
-      {props.onClearClick ? (
-        <button
-          type="button"
-          onClick={props.onClearClick}
-          className="button--clear button--icon-only-padding"
-        >
-          <Icon name="cross" />
-        </button>
-      ) : null}
+    <span className={`pill ${className}`} {...rest}>
+      {children}
     </span>
   );
 };
 
 Pill.propTypes = {
   children: PropTypes.node.isRequired,
-  onClearClick: PropTypes.func,
   className: PropTypes.string
 };
 
 Pill.defaultProps = {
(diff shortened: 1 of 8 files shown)


lib/Pill/index.js
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Icon from '../Icon';
+
+const Pill = props => {
+  return (
+    <span className={`pill ${props.className}`}>
```
kyleharper: Should we be deconstruction the required props and then spread the rest onto here?

--------

This would allow us to do a thing like add DOM attr when required or even test attributes for react testing lib.

lib/Pill/index.js
```diff
@@ -0,0 +1,33 @@
...
+      {props.children}
+      {props.onClearClick ? (
+        <button
+          type="button"
+          onClick={props.onClearClick}
+          className="button--clear button--icon-only-padding"
```
kyleharper: Any reason why you're using a `button` with the button classes on it? Seems like you wanna use the `Button` proxy component?

lib/Pill/index.js
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Icon from '../Icon';
+
+const Pill = props => {
+  return (
+    <span className={`pill ${props.className}`}>
+      {props.children}
+      {props.onClearClick ? (
```
kyleharper: Couldn't this be composed via `props.children`? Imagine if we want to use this space for another type of action in the future. What do you think?

kyleharper: This could then use any icon or any re-usable pattern within the library rather than a strictly defined one.

stories/components/Pill.js
```diff
@@ -0,0 +1,16 @@
...
+import { storiesOf } from '@storybook/react';
+import { action } from '@storybook/addon-actions';
+import Pill from '../../lib/Pill';
+import Icon from '../../lib/Icon';
+import StoryItem from '../styleguide/StoryItem';
+
+storiesOf('Components', module).add('Pill', () => (
+  <div>
+    <StoryItem title="Pill">
+      <Pill onClearClick={action('clear')}>
```
kyleharper: How does this look with multiple pills? Do we need a wrapper?

DavidHollins6: A wrapper will be needed when there's multiple pills to separate them. I didn't feel it was right to include styles to separate them inside the component.

lib/Pill/index.js
DavidHollins6: Completely agree.

lib/Pill/index.js
DavidHollins6: Agree, that's done

lib/Pill/index.js
DavidHollins6: The `Button` proxy came with a lot of extra classes that didn't work well with `button--icon-only`. It turns out whenever we use the class `button--icon-only` or `button--icon-only-padding`, it is on an `a` and not a `Button`

stories/components/Pill.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import Pill from '../../lib/Pill';
+import Icon from '../../lib/Icon';
+import StoryItem from '../styleguide/StoryItem';
+
+storiesOf('Components', module).add('Pill', () => (
+  <div>
+    <StoryItem title="Pill">
+      <Pill>
```
kyleharper: You had a button in this before to show clearing. That is still valid for the stories as you'll need styling to handle component X + Y working together.

PR #523, opened 2019-07-24
Added storybook addons

kyleharper:
This PR adds a number of add-ons to help us improve our usage of storybook moving forward. It includes;

- knobs
- a11y
- actions
- notes

I've added a new component to as an example.

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook


lib/HierarchyCollection/HierarchyCollectionStory.js
```diff
@@ -0,0 +1,45 @@
...
+            {createKnobCollection(3)}
+
+            <HierarchyCollection>{createKnobCollection(4)}</HierarchyCollection>
+          </HierarchyCollection>
+        </HierarchyCollection>
+      </HierarchyCollection>
+    </HierarchyCollection>
+  ),
+  {
+    notes: { markdown: notes }
```
AmeeMorris: Thoughts on moving this and the README.md to a `/lib/HierarchyCollection/docs/`?

DavidHollins6: I like this idea, It's still next to the implementation but separated in some way

kyleharper: @AmeeMorris I guess my only challenge would be that we've never created more than 1 doc and ideally we don't want to encourage over documenting at this point.

I think it's a little overkill especially if we're going to get the docs to auto-generate too using the info addon.

PR #524, opened 2019-07-24
Added new icon - clockWithoutPadding

DavidHollins6:
### 👀 Overview
The existing clock icon has padding around it, changing it's size to be 30x30. This is causing issues like the one below: 

Without icon:
Screenshot 2019-07-24 at 13 48 54

With icon: 
Screenshot 2019-07-24 at 13 49 38

### 💬 Description
Instead of faffing around with the sizes, peaks has kindly created a new clock icon that doesn't include the padding. Unfortunately we can't replace the current one with that, so I have created a new one. With the idea being some day we replace it.

### ✅ Checklist
- [x] Browser tested
- [x] Added to storybook


lib/Icon/index.js
```diff
@@ -164,7 +165,8 @@ const getIcon = name => {
     quoteStart,
     embed,
     people: peopleSVG,
-    item: itemSVG
+    item: itemSVG,
+    clockWithoutPadding: clockWithoutPaddingSVG
```
kyleharper: `clock` & `clockWithoutPadding` - seems like a hack to me. Why can't we have 1 clock icon? Updating this in the future (if we go down this route) will turn into a nightmare to resolve.

kyleharper: We need to decide at this point, do icons from our design source have (or don't include) padding. And then we work from there.

Then maybe we could make sure that all the icon's you want to use follow that pattern and it should be easy to resolve. (include existing usages of said icon(s).

DavidHollins6: > `clock` & `clockWithoutPadding` - seems like a hack to me. Why can't we have 1 clock icon? Updating this in the future (if we go down this route) will turn into a nightmare to resolve.

I had made the assumption that the clock was used in many places, but it turns out it's not used at all from what I can find. So I will just replace the existing clock.



> We need to decide at this point, do icons from our design source have (or don't include) padding. And then we work from there.
> 
> Then maybe we could make sure that all the icon's you want to use follow that pattern and it should be easy to resolve. (include existing usages of said icon(s).

I agree that we need to make a decision. Maybe we add it to the frontend input and have it in the next guild meeting?

PR #525, opened 2019-07-26
Hierarchy view

kyleharper:
## 💬 Description
This is quite a beefy PR but for good reason. Nothing complex but more of a shift of thinking for us as a team.

`gather-content-ui` has always been a component library only for us internally but we're experimenting on building larger UIs to illustrate how we could piece together entire chunks of UIs for the app rather than only doing that in the app.

There are many benefits;

- no business logic or speed issues (Storybook is our faster env for development)
- we can isolate UI challenge and tackle them in the correct place
- we can do design reviews in a faster env
- we can hook into addons like Knobs to allow people to customise their experience for testing
- and much more!

I have also included some much-needed packages to allow us to build faster e.g.;

- `uuid` for generating uuids for loop keps
- `faker` for generating fake content

I have also started to move library-based stories that I have touched into their component directories so things can start living in a single area rather than scaling the entire codebase to find what you are looking for 👍 

## 🚪 Start Points
I would check the "web app" based story and take it from there. These stories are just larger formats of the original library-based stories, they're a little more complicated but only because they're larger (IMO).

You can make these even simpler if you don't include Knobs and dynamic settings to be applied but you then lose the power.

### ✅ Checklist
- [x] Tests written
- [x] Browser tested
- [ ] Added to documentation
- [x] Added to storybook


.eslintrc.json
```diff
@@ -13,6 +13,7 @@
   },
   "plugins": ["jsx-a11y", "prettier"],
   "rules": {
+    "import/prefer-default-export": [0],
```
kyleharper: Always requiring default exports...yuk!

lib/Avatar/AvatarGroup/index.js
```diff
@@ -1,7 +1,7 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import cx from 'classnames';
-import ArrowIcon from '../../assets/icons/arrow-right.svg';
+import ArrowIcon from '../../../assets/icons/arrow-right.svg';
```
kyleharper: I moved `AvatarGroup` into `Avatar` because...why not!

lib/Form/RadioButton/Group.js
```diff
@@ -55,8 +55,8 @@ export default class Group extends Component {
     const targetId = e.target.id;
     const selected = this.state.choices.filter(c => c.id === targetId);
     this.setState(prevState => ({
-        selected: prevState.choices.filter(c => c.id === targetId)
-      }));
+      selected: prevState.choices.filter(c => c.id === targetId)
```
kyleharper: automated format fix (nothing has changed)

lib/ItemRow/styles.scss
```diff
@@ -47,12 +47,16 @@ $item-row-stacked-margin: $typo-size-slight/3;
   }
 }
 
+.item-row__participants {
```
kyleharper: I need to test stacked item rows to make sure this doesn't break anything!

package.json
```diff
@@ -49,7 +49,7 @@
     "test:watch": "jest --notify --watch",
     "test:ci": "jest --runInBand",
     "tdd": "yarn run test:watch",
-    "format:js": "prettier --write \"{lib,tests}/**/*.js\"",
+    "format:js": "prettier --write \"{lib,tests,stories/webapp}/**/*.js\"",
```
kyleharper: Should we format all stories? Maybe we should, I'll change this but in another PR so I don't make this MASSSSIIIVEEE!

styles/helpers/_layout.scss
```diff
@@ -1,3 +1,26 @@
+/* ==========================================================================
+   Margins & Padding
+   ========================================================================== */
+
+$props: ('margin', 'padding');
+$subProps: ('left', 'right', 'top', 'bottom');
+
+@each $p in $props {
```
kyleharper: After talking with @AmeeMorris we've agreed to give helpers more space in the library. This adds margin and padding helpers. I need to add to this though;

- [ ] add '' to `subProps` so we get all round helpers
- [ ] what about breakpoints?

styles/settings/_layout.scss
```diff
@@ -12,6 +12,11 @@ $shadow-deep: 0 $layout-spacing-base/4 $layout-spacing-base/2 $neutral-light;
 $shadow-shallow: 0px 0px 6px 0px transparentize($neutral-light, 0.7);
 $shadow-base: 0 0 6px 0 transparentize($neutral-light, 0.29);
 
+/* ==========================================================================
+   Bordered
+   ========================================================================== */
+$border-radius-base: 4px !default;
```
kyleharper: We define `4px` in a few placed so 👍

kyleharper:
[Create <HierarchyList /> in gather UI](https://trello.com/c/JvPs1hSf/243-create-hierarchylist-in-gather-ui)

lib/FolderRow/FolderRow.js
```diff
@@ -1,37 +1,71 @@
...
 
   return (
-    <div className={classNames} {...props}>
-      <div className="folder-row__name">
-        <div className="text-overflow-ellipsis">{children}</div>
-      </div>
+    <Fragment>
+      <div className={classNames} {...props}>
+        <div className="folder-row__name">
+          <div className="text-overflow-ellipsis">{name}</div>
```
DavidHollins6: Is there a better html element to describe this?

lib/FolderRow/FolderRow.js
```diff
@@ -1,37 +1,71 @@
...
+        </div>
 
-      <div className="folder-row__meta">
-        {actions && <div className="folder-row__actions">{actions}</div>}
-        {metaText && <div className="folder-row__meta-text">{metaText}</div>}
+        <div className="folder-row__meta">
+          {actions && <div className="folder-row__actions">{actions}</div>}
+          {metaText && <div className="folder-row__meta-text">{metaText}</div>}
```
DavidHollins6: Same here regarding `div`, if `metaText` is just text, I'm not sure `div` is the correct element.

package.json
DavidHollins6: 👍

styles/helpers/_layout.scss
DavidHollins6: This is cool!

altaudio:
@kyleharper Loving the new concept, I think building larger UI in gather ui is 💯 

This PR also seems to be building some specific UI (from the title), but there's no detail of it in the description or the attached Trello cards. If this is the case, could we see the designs/UI in action? That would maximise the "context transfer" efficiency.

kyleharper:
@altaudio I think grabbing me and us having a chat is best. I don't agree with adding too much context to a public repo or Trello cards that only we can access.

Some overall context will help though so ping me. I'd also recommend just reviewing the code and looking for smells at the point.

kyleharper:
@AmeeMorris @altaudio find design attached. I'll be merging this morning, please review by then 👍 

design

lib/FolderRow/FolderRow.js
AmeeMorris: Was there a consensus on what to use after?

lib/FolderRow/FolderRow.js
```diff
@@ -1,37 +1,71 @@
...
+          {showToggle && (
+            <div className="folder-row__toggle h-margin-left-half">
+              <Button types={['icon-only']} onClick={() => setToggle(!toggle)}>
+                <Icon name={toggle ? 'caretUp' : 'caret'} />
```
AmeeMorris: Not for now, but I feel like this could be useful to split out into a little helper, its a pattern thats pretty common

lib/FolderRow/FolderRow.js
```diff
@@ -1,37 +1,73 @@
...
+  actions,
+  metaText,
+  open,
+  className,
+  showToggle,
+  ...props
+}) => {
+  const [toggle, setToggle] = useState(open);
+
+  useEffect(() => {
```
altaudio: 👍

lib/FolderRow/FolderRowStory.js
```diff
@@ -0,0 +1,53 @@
+import React, { Fragment } from 'react';
```
altaudio: Feels like we should have a `stories` directory, even if it just has an `index.js` file. 

I feel like keeping the components and stories at the same level is a bit messy.

package.json
altaudio: Yeah, prolly lint as well. Although I wouldn't want to be the person to fix all the lint errors!

lib/FolderRow/FolderRowStory.js
kyleharper: No more `index.js` files, they make the codebase very difficult to search.

However, you could have multiple so I agree there 👍

lib/FolderRow/FolderRow.js
kyleharper: `p` tags for this 👍

kyleharper:
@altaudio @DavidHollins6 feedback applied, please approve 👍

lib/FolderRow/FolderRowStory.js
altaudio: Good point!

PR #531, opened 2019-07-31
User Search Heading

DavidHollins6:
### 👀 Overview
The user search heading should only be displayed if one has been provided.
### 💬 Description
I have also added some knobs to the UserSearch story.

### ✅ Checklist
- [ ] Tests written
- [x] Browser tested
- [ ] Added to documentation
- [x] Added to storybook


DavidHollins6:
> but if searchHeading is now optional, maybe make the prop not required?

Good point!

PR #532, opened 2019-07-31
Item row refactor

kyleharper:
## 👀 Overview
**Breaking change**: `ItemRow` has been refactored to follow the compound component pattern.

## 💬 Description
In future work, we're adding complexity to the item row (adding additional item data). The previous format of the component took fixed prop values and render components based on those props.

We've faced multiple challenges when our UI components do too many specific things and this is following that same problem so we're fixing it before that happens.

I'll need to alter our consuming app before we deploy this as it is a breaking change!

## 🚪 Start Points
Visually nothing has changed, from #525 however we can now add more data cells without needing to manage specific states and styling.

The commits are pretty well staged to if you wanted to take that trip (your call).

### ✅ Checklist
- [x] Tests written
- [x] Browser tested
- [x] Added to documentation
- [x] Added to storybook


lib/Avatar/stories/AvatarGroupMock.js
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
+import { number } from '@storybook/addon-knobs';
+import faker from 'faker';
+import uuid from 'uuid/v1';
+import AvatarGroup from '../AvatarGroup';
+import { AvatarWithPopover, ParticipantInfo } from '../../index';
+
+const AvatarGroupMock = ({ children, defaultValue }) => {
```
kyleharper: In the work for #525 I added this function but now the item row story now uses it so I thought it would be a great idea to move this into a kind of mocking component.

Meaning whenever we want an AvatarGroup in stories you can use this mock rather than detailing our the same implementation every time.

lib/ItemRow/ItemRowName.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { node, oneOfType, string } from 'prop-types';
+
+const ItemRowName = ({ children, ...props }) => {
+  return (
+    <div className="item-row__name" {...props}>
+      {children}
```
kyleharper: This can be text but ideally, the consuming parent will define the correct text node e.g. `p`. Maybe I should add that to the story?

lib/List/styles.scss
```diff
@@ -18,6 +18,12 @@ $list-border: 1px solid $color-border-base;
   .is-active & svg path {
     fill: $color-text-base;
   }
+
+  &.is-active {
```
kyleharper: This was in the item row stylesheet but had very little to do with it. I moved it into here 👍

stories/webapp/hierarchy/HierarchyFolderRow.js
```diff
@@ -10,18 +10,16 @@ export const HierarchyFolderRow = ({ childCount, children, open }) => (
     name={faker.commerce.department()}
     open={open}
     showToggle
+    style={{ minWidth: '320px' }}
```
kyleharper: This min-width allows us to handle hierarchy levels that go way beyond the screen width. It doesn't show a whole role of data but at least it doesn't break.

You can test this in the storybook.

tests/ItemRow/ItemRow.spec.js
```diff
@@ -1,59 +1,16 @@
...
-    );
-  });
-
-  test('renders an indicator', () => {
-    expect(wrapper.find('.item-row__indicator').contains('indicator')).toEqual(
-      true
-    );
-  });
-
-  test('renders an label', () => {
```
kyleharper: Tests no longer required...everything is handled via the compound pattern and render children 👍

lib/ItemRow/ItemRowName.js
```diff
@@ -0,0 +1,22 @@
...
+  return (
+    <div className="item-row__name" {...props}>
+      {status}
+      {children}
+    </div>
+  );
+};
+
+ItemRowName.propTypes = {
+  children: oneOfType([node, string]).isRequired,
```
altaudio: The `node` PropType includes strings already

lib/ItemRow/ItemRowAside.js
```diff
@@ -0,0 +1,16 @@
...
+const ItemRowAside = ({ children, ...rest }) => {
+  return (
+    <div className="item-row__aside" {...rest}>
+      {children}
+    </div>
+  );
+};
+
+ItemRowAside.propTypes = {
+  children: oneOfType([node, arrayOf(node)]).isRequired
```
altaudio: `node` PropTypes includes arrays

lib/Avatar/stories/createAvatarGroup.js
```diff
@@ -0,0 +1,39 @@
...
+const createAvatarGroup = (defaultValue = 3) => {
+  const count = Math.floor(
+    Math.random() * number('Max number of assignees', defaultValue) || 0
+  );
+
+  return (
+    <AvatarGroup maximum={3}>
+      {[...Array(count).keys()].map(() => {
+        const name = faker.name.findName();
+        const initials = `${[...name][0]}${[...name][1]}`;
```
altaudio: Can we just `name.subString(0, 2)`

lib/Avatar/stories/createAvatarGroup.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import { number } from '@storybook/addon-knobs';
+import faker from 'faker';
+import uuid from 'uuid/v1';
+import AvatarGroup from '../AvatarGroup';
+import { AvatarWithPopover, ParticipantInfo } from '../../index';
+
+const createAvatarGroup = (defaultValue = 3) => {
```
altaudio: Love this ❤️

lib/ItemRow/index.js
```diff
@@ -19,13 +19,10 @@ const ItemRow = ({ children, stacked, className, bordered, ...rest }) => {
...
 
 ItemRow.propTypes = {
-  children: PropTypes.oneOfType([
-    PropTypes.node,
-    PropTypes.arrayOf(PropTypes.node)
-  ]).isRequired,
-  stacked: PropTypes.bool,
-  bordered: PropTypes.bool,
-  className: PropTypes.string
+  children: oneOfType([node, arrayOf(node)]).isRequired,
```
altaudio: Same comment on node and arrays

stories/webapp/hierarchy/HierarchyCollection.js
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import PropTypes from 'prop-types';
+import { number, bool } from 'prop-types';
```
altaudio: We should have a prize for anyone who can write a script to run over our whole codebase to refactor all the PropTypes in this way 🏆

lib/ItemRow/README.md
```diff
@@ -1,22 +1,32 @@
...
-| children            | Node          | N/A       | Yes      | A node which renders inside a row. Mostly likely a text and a link.                    |
-| indicator           | Node          | null      | No       | A StatusIndicator component.                                                           |
-| label               | String        | null      | No       | A label which renders next to the child contents.                                      |
-| commentCount        | String        | "0"      | No       | Displays the passed comment count                                      |
+| children            | Node          | N/A       | Yes      | 
+| stacked             | Bool          | N/A       | Yes      | Adds a stacked modifier to apply stacked style. |
+| bordered            | Bool          | N/A       | Yes      | Adds a bordered modifier to apply padding and boarders.                    |
+| *                   | N/a           | N/A       | No       | Spreads the remaining props onto the root component.                    |
```
altaudio: 👍 Like this. Maybe we should specify what the root 
element is?

lib/Avatar/stories/AvatarGroupMock.js
altaudio: Love this 💟

lib/ItemRow/ItemRowData.js
```diff
@@ -0,0 +1,16 @@
...
+const ItemRowData = ({ children, ...rest }) => {
+  return (
+    <div className="item-row__data" {...rest}>
+      {children}
+    </div>
+  );
+};
+
+ItemRowData.propTypes = {
+  children: oneOfType([node, arrayOf(node)]).isRequired
```
altaudio: Same comment on the `node` PropType

lib/ItemRow/ItemRowName.js
```diff
@@ -0,0 +1,16 @@
...
+const ItemRowName = ({ children, ...props }) => {
+  return (
+    <div className="item-row__name" {...props}>
+      {children}
+    </div>
+  );
+};
+
+ItemRowName.propTypes = {
+  children: oneOfType([node, string]).isRequired
```
altaudio: `node` PropType includes strings

lib/ItemRow/stories/ItemRowStory.js
```diff
@@ -0,0 +1,75 @@
...
+  () => {
+    const stacked = boolean('Stacked', false);
+    const bordered = boolean('Bordered', true);
+    const commentCount = text('Comment count', 0);
+    const templateName = text('Template name', '');
+
+    return (
+      <ItemRow bordered={bordered} stacked={stacked}>
+        <ItemRow.Name>
+          {!stacked &&
```
altaudio: What's the `stacked` value?

styles/helpers/_layout.scss
```diff
@@ -14,6 +14,10 @@ $subProps: ('', 'left', 'right', 'top', 'bottom');
     #{$property}: $layout-spacing-base/2;
   }
 
+  #{$selector}-quarter {
```
altaudio: Like these 👍

lib/Avatar/stories/AvatarGroupMock.js
AmeeMorris: +1 great idea!

lib/ItemRow/ItemRowAside.js
AmeeMorris: didnt know this!

lib/ItemRow/ItemRowName.js
AmeeMorris: or this!

lib/ItemRow/README.md
kyleharper: I don't think we need to over document it. Check the code if you wanna know details like that (IMO).

lib/ItemRow/stories/ItemRowStory.js
kyleharper: It's set in knobs above, it's a boolean and adds a modifier.

The problem is that it requires a little more than just the CSS modifier to be valid, it requires a slight composition re-jig on the component side.

Hense this code.

kyleharper:
@altaudio thanks for the feedback dude, the prop type one was really epic ❤️ 

If you could -re-review this then that would be great. I need to also get https://github.com/gathercontent/gather-ui/pull/534 so I can test them both along with the consuming app.

lib/ItemRow/stories/ItemRowStory.js
altaudio: So what does it do in human terms? eg. "You would turn the stacked modifier on if you wanted to see..."?

lib/ItemRow/ItemRowAside.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { node } from 'prop-types';
+
+const ItemRowAside = ({ children, ...rest }) => {
+  return (
+    <div className="item-row__aside" {...rest}>
```
DavidHollins6: Could this be an `aside` element?

lib/ItemRow/ItemRowName.js
DavidHollins6: 👍

lib/ItemRow/ItemRowAside.js
kyleharper: I did originally but the aside isn't in anything substantial to warrant it. For example `article` or whatever.

lib/ItemRow/stories/ItemRowStory.js
kyleharper: You'd turn it on if you wanted the elements to stack so;

- name
- aside
-- data
-- data

By default they are inline 😄

kyleharper: If you open storybook up you can see it visually, it's pretty obvious.

PR #533, opened 2019-08-01
Update/styled components example

altaudio:
### 👀 Overview
Refactors the `<InputWithButton />` component to use `styled-components`, as a basis for discussion.

Review this PR at your leisure, definitely non-urgent.
#### What the heck is `styled-components`?
[styled-components](https://www.styled-components.com/docs) is a library that lets you style React components directly, without having to use CSS files.

#### Styling a standard HTML element
The following code creates a component, which is a styled version of the `<input />` HTML element. We use the `styled.[element]` syntax to style a standard HTML element.
```
// components/core/Input.js

import styled from 'styled-components';

const Input = styled.input`
  border: solid 1px red;
  color: pink;
`
export default Input;

```
The exported value is a React component, which we can import and use elsewhere:
```
// components/Form.js

import Input from 'components/core/Input;'

export default () => (
  <Form>
    <Input onChange={({ target: { value }}) => console.log(value)} />
  </Form>
)
```
#### Styling a styled component
If we wanted to add some specific styling to `<Input />`, for example it needed some extra margin on top, we can style the component using the `styled(Component)` syntax: 
```
// components/Form.js

import Input from 'components/core/Input;'
import styled from 'styled-components';

const InputWithMargin = styled(Input)`
  margin-top: 50px;
`
export default () => (
  <Form>
    <InputWithMargin onChange={({ target: { value }}) => console.log(value)} />
  </Form>
)
```
`<InputWithMargin />` is a new component with the styling of the original `<Input />` component, plus the top margin. We can style any component that accepts a `className` prop in this way.
#### Props
Any props we pass to a styled component can also be used in the CSS 😱:
```
// components/core/Input.js

import styled from 'styled-components';

const Input = styled.input`
  border: solid 1px red;
  color: ${props => props.disabled ? 'grey' : 'pink'};
  pointer-events: ${props => props.disabled ? 'none' : 'all'};
`

export default Input;
```
Here we apply styles conditionally based on the `disabled` prop value, that we pass into the form:
```
// components/Form.js

import Input from 'components/core/Input;'

export default ({ disabled }) => (
  <Form>
    <Input disabled={disabled} onChange={({ target: { value }}) => console.log(value)} />
  </Form>
)
```
As our styles are now just string literals, we can use the full power of JS in our CSS. Want to create a `getColorFromStatus` function to use in your CSS? EzPz. Want to do some mad calculations using a random NPM library? Go for it!
#### ThemeProvider
`styled-components` also exports a `<ThemeProvider>` component, which you pass a `theme` object. This theme object contains all of your style values, eg:
```
{
  colors: {
    white: '#FF',
    neutral: {
      dark: 'rgb(41, 51, 61)',
      base: 'rgb(133, 153, 173)',
      light: 'rgb(224, 229, 234)',
      lightest: 'rgb(249, 250, 251)'
    }
  },
  fonts: {
    sizes: {
      xlarge: '1.5rem',
      large: '1.25rem',
      lead: '1.125rem',
      base: '1rem',
      slight: '.875rem',
      small: '.75rem',
      micro: '.6875rem'
    },
}
```
Then we wrap our entire app in the `<ThemeProvider />`, with our theme object passed as a prop:
```
import { ThemeProvider } from 'styled-components';
import { theme } from '../styles/theme';
....
export default () => (
  <ThemeProvider theme={theme}>
    <App />
  </ThemeProvider>
)
```
Now, every styled component can access the theme object through it's props:
```
// components/core/Input.js

import styled from 'styled-components';

const Input = styled.input`
  border: solid 1px red;
  color: ${props => props.disabled ? props.theme.colors.neutral.dark : props.theme.colors.neutral.light};
  pointer-events: ${props => props.disabled ? 'none' : 'all'};
`
```
#### Why is this better than standard CSS files?
It's not necessarily better, just different, and definitely a matter of taste. Here are the reasons I like styled components (this is a straight cp from their docs, but they wrote it better than I could have):

- Automatic critical CSS: styled-components keeps track of which components are rendered on a page and injects their styles and nothing else, fully automatically. Combined with code splitting, this means your users load the least amount of code necessary.
- No class name bugs: styled-components generates unique class names for your styles. You never have to worry about duplication, overlap or misspellings.
- Easier deletion of CSS: it can be hard to know whether a class name is used somewhere in your codebase. styled-components makes it obvious, as every bit of styling is tied to a specific component. If the component is unused (which tooling can detect) and gets deleted, all its styles get deleted with it.
- Simple dynamic styling: adapting the styling of a component based on its props or a global theme is simple and intuitive without having to manually manage dozens of classes.
- Painless maintenance: you never have to hunt across different files to find the styling affecting your component, so maintenance is a piece of cake no matter how big your codebase is.
Automatic vendor prefixing: write your CSS to the current standard and let styled-components handle the rest.

You get all of these benefits while still writing the CSS you know and love, just bound to individual components.

`styled-components` can do [so much more](https://www.styled-components.com/docs/), but those are the basics.
#### So what does _this PR_ do?
- Creates a `<Input />` component with default styling.
- Refactors the `<InputWithButton />` component.
- Adds a `theme` file.
- Exports the `<ThemeProvider />` from `lib/index.js` to be used in the app. This is a fix, so that wrapping the `<App />` in the `<ThemeProvider />` passes the `theme` object down to components imported from `gather-content-ui`. Fix from [this issue](https://github.com/styled-components/styled-components/issues/2379)
### 👫 Related PRs (optional)
https://github.com/gathercontent/app/pull/1997

#### How are the resulting styles outputted?

For each styled component, the library creates a stylesheet containing the components styles, and passes the corresponding `className` to the component.

The library then adds the stylesheets to the `<head/>` of the DOM at _runtime_ (eg. it only adds stylesheets for components that are on the page 😱)

Check out these two bits of the docs for more info
- https://www.styled-components.com/docs/advanced#existing-css
- https://www.styled-components.com/docs/advanced#issues-with-specificity

#### Do you have any good examples of styled components in large libraries?

Sure thing!
##### [Grommet (used by Uber, Netflix etc.)](https://github.com/grommet/grommet)
This library is a great example of SC at scale. There are others out there, but this one seemed the most useful to me. Below are some good jumping off points!
- [Button Component](https://github.com/grommet/grommet/blob/master/src/js/components/Button/StyledButton.js)
- [Their themes](https://github.com/grommet/grommet/tree/master/src/js/themes) (as they are just objects, they can switch them instantly to change between light/dark themes)
- [More complex elements of a calendar component](https://github.com/grommet/grommet/blob/master/src/js/components/Calendar/StyledCalendar.js)

#### What the heck is this weird `styled''` syntax?
- This blog post explains [more than you'll ever need to know!](https://mxstbr.blog/2016/11/styled-components-magic-explained/)
### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook


lib/InputWithButton/index.js
```diff
@@ -1,8 +1,32 @@
...
 import Button from '../Button';
 
+const Container = styled.div`
+  display: flex;
+  flex-direction: row;
+`;
+
+const LeftHandInput = styled(Input)`
+  width: 220px;
+  border-radius: 4px 0px 0px 4px;
```
kyleharper: So where should we define `consts`? 😏 

Ideally, we should mostly map what we have now I think.

kyleharper: Nvm, just saw the `theme` stuff 👍

lib/InputWithButton/index.js
```diff
@@ -1,8 +1,32 @@
...
+`;
+
+const RightHandButton = styled(Button)`
```
kyleharper: I was kind of hoping to see some deleted styles, to get an idea of the refactors from `scss` to this. Could we do that?

lib/styles/theme.js
```diff
@@ -0,0 +1,29 @@
...
+      small: '.75rem',
+      micro: '.6875rem'
+    },
+    lineHeights: {
+      base: '1.25rem',
+      lead: '1.5rem'
+    }
+  },
+  layout: {
+    spacingBase: 20
```
kyleharper: We've been wanting this for a while in `scss` bu never added it - lets bake in some modifiers by default like `half` or whatever, what do you think?

stories/components/InputWithButton.js
```diff
@@ -1,44 +1,60 @@
...
-          value="https://www.google.com/search?safe=off&source=hp&ei=-mfmXMLbCdChkwWAvpnIDg&q=hello&oq=hello"
-          onClick={value => {}}
-          buttonText="Click Me"
-          buttonTextAfterClick="I've been clicked!"
-          paddingSmall
-        />
-      </div>
-    </StoryItem>
-  </div>
+  <ThemeProvider theme={theme}>
```
kyleharper: Hmm, I wonder if we can include by default somewhere so it's always there. Any thoughts?

lib/InputWithButton/index.js
```diff
@@ -1,8 +1,32 @@
...
+  ${props =>
+    props.paddingSmall
```
kyleharper: How do we handle multiple modifiers?

```
styles(Button)`
  ${props =>
    props.paddingSmall
      ? '' : '',
    props.paddingLarge
      ? '' : ''
   }
`

Does the function take multiple arguments?
```

lib/InputWithButton/index.js
kyleharper: I realised you wrote it in the PR but I kind of wanted a super fresh opinion without any pre-knowledge, thanks for writing the description etc.

altaudio:
@altaudio To add info on the outputted styles. (Are they inline, abstracted etc.?)

altaudio:
Comments from @DavidHollins6 
> The styled component s are really cool! I am just struggling to imagine what it would look like when scaled up
>
> Do you have any good examples of styled components in large libraries?

lib/InputWithButton/index.js
altaudio: We have one deleted `scss` file here https://github.com/gathercontent/gather-ui/pull/533/files#diff-07096a4f84efc8ff7ff11c7cf5b8b024

lib/styles/theme.js
altaudio: Great idea, I'll add those in :+1:

stories/components/InputWithButton.js
altaudio: Good shout, just found this storybook plugin!

Adding here as a note https://github.com/echoulen/storybook-addon-styled-component-theme

lib/InputWithButton/index.js
altaudio: Whatever is returned from the function is added as a value in the string literal (just as in normal JS), eg.

```
styled(Button)`
  padding: ${props => props.paddingSmall ? '5' : '10'}px; 
`
```
is essentially just 
```
styled(Button)`
  padding: ${props => {
      return props.paddingSmall ? '5' : '10'
   }px; 
`
```
and you can call the function as many times as you like in the string literal
```
styled(Button)`
  padding: ${props => props.paddingSmall ? '5' : '10'}px; 
  width: ${props => props.fullWidth ? '100%' : '500px'}; 
`
```
I don't think this is what you are asking though, so ping me back!

PR #534, opened 2019-08-01
AvatarGroup toggle refactor

kyleharper:
## 👀 Overview
Improve the toggling UI for `AvatarGroup`

## 💬 Description
Currently, the `AvatarGroup` component renders an [inline toggle](https://cl.ly/7f5e63e49fc3) when the avatar count exceeds the count set in props. This causes various issues in the UI because when it opens it can move other UI or even break out of its container, which isn't ideal.

The solution is to move from this toggle and toward the `Dropdown` pattern. 

------------

There are a few bonuses in this PR (related to the work anyways);

- `AvatarStory` now uses knobs

- `autoPosition` in `Dropdown` didn't detect width based positions, it also didn't handle if the new position set by the logic itself exceeded the browser, now it does. This code change is a little smelly but that's due to the underlining pattern itself, not so much the code I think.

- `AvatarGroupMock` now allows more props, makes it more helpful

------------

Another side note is that I am going to merge this into #276 as I want both to go live at the end time. I am aware that I will need to test #276 when this has been merged in.

## 📹 GIF (optional)

- https://cl.ly/6ed1bbee9977

## 🚪 Start Points
I'll leave comments where required but just take a look at the code in general. `AvatarGroup` is the one with the largest refactor.

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook


lib/Avatar/AvatarGroup/index.js
```diff
@@ -1,93 +1,68 @@
...
- * </AvatarGroup>
- */
-class AvatarGroup extends Component {
-  constructor() {
-    super();
+const AvatarGroup = ({ children, maximum, className, small }) => {
+  const total = (children && children.length) || 0;
+  const plusLabel = `+${total - maximum}`;
 
-    this.state = { showExtra: false };
```
kyleharper: Now that the dropdown has been added we don't need to manually toggle.

lib/Avatar/AvatarGroup/index.js
```diff
@@ -1,93 +1,68 @@
...
+      })}
 
-          return (
-            <span className="avatar-group__item" style={styles}>
-              {React.cloneElement(child)}
-            </span>
-          );
-        })}
+      {total > maximum && (
+        <Dropdown id="avatar-group-dropdown" autoPosition>
```
kyleharper: This is where the real change in this component happens.

lib/Avatar/AvatarGroup/styles.scss
```diff
@@ -2,7 +2,7 @@
    Config
    ========================================================================== */
 $avatar-group-item-spacing: .115em;
-$avatar-goup-size-small: 24px !default;
+$avatar-group-size-small: 24px !default;
```
kyleharper: Spelling mistake.

lib/Avatar/stories/AvatarStory.js
```diff
@@ -0,0 +1,117 @@
...
+  AvatarWithPopover,
+  ParticipantInfo,
+  AvatarInformation,
+  Button
+} from '../../index';
+import StoryItem from '../../../stories/styleguide/StoryItem';
+import notes from '../README.md';
+import { AvatarGroupMock } from './AvatarGroupMock';
+
+storiesOf('Components', module).add(
```
kyleharper: The story now uses knobs and for the better 👍

lib/Dropdown/DropdownContent.js
```diff
@@ -4,32 +4,76 @@ import PropTypes from 'prop-types';
 import { GATHER_UI_DROPDOWN } from './consts';
 
 class DropdownContent extends Component {
+  getTopPosition = (pos, bounds, contentBounds) => {
```
kyleharper: This is the dropdown auto position changes. The process goes as follows;

- get what the top and left positions will be and run them through `getTopPosition` and `getLeftPosition`
- if they exceed then alter the position
- if the altered position goes off-screen the other way then alter the position
- if the position is always good, just remain as default

lib/Dropdown/index.js
```diff
@@ -100,7 +100,10 @@ class Dropdown extends Component {
         bounds: {
           top: bounds.top,
           left: bounds.left,
-          width: bounds.width
+          right: bounds.right,
```
kyleharper: We needed these for the auto-positioning logic

styles/helpers/_layout.scss
```diff
@@ -28,17 +28,11 @@ $subProps: ('', 'left', 'right', 'top', 'bottom');
    ========================================================================== */
 
 $props: ('display');
-$subProps: ('none', 'block', 'inline-block', 'flex', 'inline-flex', 'inline');
+$subProps: (none, block, inline-block, flex, inline-flex, inline);
```
kyleharper: These helpers never worked because there were strings. `none` etc aren't strings.

lib/Dropdown/DropdownContent.js
```diff
@@ -4,32 +4,76 @@ import PropTypes from 'prop-types';
...
-            (bounds.top + contentBounds.height - window.innerHeight) -
-            40,
-          left
-        };
-      }
-      return { top: bounds.top - 20, left };
+
+      const topPos =
+        bounds.top !== 'auto'
+          ? bounds.top + contentBounds.height + 40
```
altaudio: Add this to a const

lib/Avatar/stories/AvatarGroupMock.js
```diff
@@ -5,27 +5,40 @@ import uuid from 'uuid/v1';
...
+  children,
+  defaultCount,
+  minCount,
+  avatarProps,
+  avatarGroupProps,
+  overlayProps
+}) => {
   const count = Math.floor(
-    Math.random() * number('Max number of assignees', defaultValue) || 0
+    Math.random() * number('Max number of assignees', defaultCount || 0)
```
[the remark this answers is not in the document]
kyleharper: Hmm, it's the default max count for the knobs control. So I'll rename it as `defaultMaxCount`.

lib/Avatar/AvatarGroup/index.js
```diff
@@ -1,93 +1,68 @@
...
+const AvatarGroup = ({ children, maximum, className, small }) => {
+  const total = (children && children.length) || 0;
+  const plusLabel = `+${total - maximum}`;
 
-    this.state = { showExtra: false };
-    this.showMore = this.showMore.bind(this);
-  }
+  let display = children;
+  let remaining = 0;
+  let index = children.length;
```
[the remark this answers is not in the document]
kyleharper: Well this is the length so I'll check to see if it makes more sense to just name it `avatarLength` which is better IMO.

lib/Avatar/AvatarGroup/index.js
```diff
@@ -1,93 +1,68 @@
...
+            <Avatar initials={plusLabel} />
+          </Dropdown.Trigger>
 
-        {this.state.showExtra &&
-          React.Children.map(remaining, child => {
-            const styles = { zIndex: index };
-            index -= 1;
+          <Dropdown.Content>
+            {React.Children.map(remaining, child => {
+              const styles = { zIndex: index };
```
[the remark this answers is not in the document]
kyleharper: They overlap in the opposite direction to native node stacking so we fix this with z-index. Here's a GIF of my turning it on and off: https://cl.ly/54ed5054c335

kyleharper:
Thanks for feedback @altaudio - can you re-review? 👍

lib/Avatar/AvatarGroup/index.js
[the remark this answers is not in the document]
altaudio: But aren't you applying a z-index to these components at this point? 
Screenshot 2019-08-02 at 12 51 36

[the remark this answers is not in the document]
kyleharper: Good shout @altaudio - in the original version they were inline so they needed the z-index. Now they do not 👍

kyleharper:
All ready @altaudio

PR #535, opened 2019-08-05
More semantic user search

DavidHollins6:
### 👀 Overview
User search could do with being a little more semantic. This makes testing it with react-testing-library easier.

### 💬 Description
The list is now a `ul` instead of just a Fragment, and I have wrapped every `button` with `li` to signify that it's a list item.

### ✅ Checklist
- [ ] Tests written
- [x] Browser tested
- [ ] Added to documentation
- [x] Added to storybook

### DIFF
--- a/lib/UserSearch/UserSearchList.js
+++ b/lib/UserSearch/UserSearchList.js
@@ -22,7 +22,7 @@ const UserSearchList = ({
     );
   }
   return (
-    <ul className="user-search-list">
+    <ul className="user-search-list h-padding-clear h-margin-clear:q">
       {users.map(user => (
         <li className="user-search-list__user">
           <Dropdown.Action
--- a/lib/UserSearch/UserSearchList.js
+++ b/lib/UserSearch/UserSearchList.js
@@ -22,7 +22,7 @@ const UserSearchList = ({
     );
   }
   return (
-    <ul className="user-search-list h-padding-clear h-margin-clear:q">
+    <ul className="user-search-list h-padding-clear h-margin-clear">
       {users.map(user => (
         <li className="user-search-list__user">
           <Dropdown.Action
--- a/lib/UserSearch/UserSearchList.js
+++ b/lib/UserSearch/UserSearchList.js
@@ -1,4 +1,4 @@
-import React, { Fragment } from 'react';
+import React from 'react';
 import PropTypes from 'prop-types';
 import Dropdown from '../Dropdown';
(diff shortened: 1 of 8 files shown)


lib/UserSearch/UserSearchList.js
```diff
@@ -22,35 +22,36 @@ const UserSearchList = ({
     );
   }
   return (
-    <Fragment>
+    <ul className="user-search-list h-padding-clear h-margin-clear:q">
```
altaudio: Do you think it's worth adding the `<ul>` and `<li>` items to the `<Dropdown>` and `<Dropdown.Action>` components respectively?

altaudio: So we don't have to wrap them every time we use them in the app?

kyleharper: What's the `:q` on the end of this class? 🤔

kyleharper: @altaudio are you suggesting that dropdowns are only ever used for lists? Because if so that isn't the case, `ActionGroup` and `Action` would only be the case where it would warrant the fixed markup IMO.

lib/UserSearch/styles.scss
```diff
@@ -29,14 +29,14 @@
   .toggle-wrapper {
     margin-left: auto;
     flex-shrink: 0;
-    padding-right: $layout-spacing-base * .75;
+    padding-right: $layout-spacing-base * 0.75;
```
kyleharper: I wonder why we `* 0.75` in the first place 🤔 

Is this padding used anywhere else in `gather-ui`? If not I recommend we fix this to make it more consistent with every other UI component 👍

lib/UserSearch/styles.scss
```diff
@@ -64,27 +64,32 @@
...
+    &:last-child {
+      border-bottom: 1px solid $neutral-light;
+    }
+
+    .user-search-list__user-name {
+      display: flex;
+      align-items: center;
+
+      .icon {
+        margin-left: $layout-spacing-base/4;
```
kyleharper: I've been handling icon spacing with helpers. It removes the need for these little bits of `scss` every time we add icons to any component 😄 

What are your thoughts are using a helper here instead of this?

lib/UserSearch/UserSearchList.js
altaudio: Sure thing, thanks for the response 👍

PR #537, opened 2019-08-06
Add Drag and Drop UI components

kyleharper:
## 👀 Overview
This PR adds UI components which enabled us to establish re-usable drag and drop patterns.

## 💬 Description
In the past, we've added Drag and Drop to our consuming app directly because we believed it was better suited to not include that domain into this library. However, with the new hierarchy view, we cannot build it without drag and drop.

As well as this we had a number of substantial challenges when building our Drag and Drop experience in the app, to my knowledge the biggest challenge was;

- generating a custom drag preview (which altered based on selection context & warnings)

We thought as well as creating the predictable `Draggable` & `Droppable` components I would also resolve the challenge above by adding a `DragPreview` component and showing how we can disable the default preview per drag instance.

We also thought that we'd show just how we can handle custom `canDrop` logic which is shown in the GIF on this PR.

**Surely this isn't just using components?**

You're correct, this PR also contains a provider which enables us to have a top-level parent to manage our drag and drop experience.

**Is this tested?**

No. I tried to get tests working but even using their tutorial doesn't work. I will investigate this further but please don't hold this back because of that. We will get this covered, just not right now.

## 📹 GIF (optional)
https://cl.ly/11ab65720e16

## 🚪 Start Points
I've commented on the changes.

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,5 +1,13 @@
 module.exports = {
   verbose: true,
   testMatch: ['**/tests/**/*.js?(x)'],
-  testPathIgnorePatterns: ['tests/setup.js']
+  testPathIgnorePatterns: ['tests/setup.js'],
+  moduleNameMapper: {
+    "^dnd-core$": "dnd-core/dist/cjs",
+    "^react-dnd$": "react-dnd/dist/cjs",
+    "^react-dnd-html5-backend$": "react-dnd-html5-backend/dist/cjs",
+    "^react-dnd-touch-backend$": "react-dnd-touch-backend/dist/cjs",
+    "^react-dnd-test-backend$": "react-dnd-test-backend/dist/cjs",
+    "^react-dnd-test-utils$": "react-dnd-test-utils/dist/cjs"
+  }
 };
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,7 +1,7 @@
 module.exports = {
   verbose: true,
   testMatch: ['**/tests/**/*.js?(x)'],
-  testPathIgnorePatterns: ['tests/setup.js'],
+  testPathIgnorePatterns: ['tests/setup.js', 'dist'],
   moduleNameMapper: {
     "^dnd-core$": "dnd-core/dist/cjs",
     "^react-dnd$": "react-dnd/dist/cjs",
--- a/lib/DnD/DndProvider.js
+++ b/lib/DnD/DndProvider.js
@@ -1,0 +1,37 @@
+import React, { createContext, useState } from 'react';
+import { DndProvider as DragAndDropProvider } from 'react-dnd';
+import { node, shape } from 'prop-types';
+import { DragPreview } from './DragPreview';
+
+export const DndContext = createContext({});
+
+function DndProvider({ children, backend }) {
+  const [isDragging, setIsDragging] = useState('');
+  const [preview, setPreview] = useState(null);
+  const [failurePreview, setFailurePreview] = useState(null);
+
+  const sharedState = {
+    isDragging,
+    setIsDragging,
+    preview,
+    setPreview,
+    setFailurePreview
+  };
+
+  return (
+    <DragAndDropProvider backend={backend}>
+      <DndContext.Provider value={sharedState}>
+        {isDragging && <DragPreview>{failurePreview || preview}</DragPreview>}
+
+        {children}
+      </DndContext.Provider>
+    </DragAndDropProvider>
+  );
+}
+
+DndProvider.propTypes = {
+  children: node.isRequired,
+  backend: shape.isRequired
+};
+
+export { DndProvider };
--- a/lib/DnD/DndProvider.js
+++ b/lib/DnD/DndProvider.js
@@ -1,12 +1,12 @@
 import React, { createContext, useState } from 'react';
 import { DndProvider as DragAndDropProvider } from 'react-dnd';
-import { node, shape } from 'prop-types';
+import { func, node } from 'prop-types';
 import { DragPreview } from './DragPreview';
 
 export const DndContext = createContext({});
 
 function DndProvider({ children, backend }) {
-  const [isDragging, setIsDragging] = useState('');
+  const [isDragging, setIsDragging] = useState(false);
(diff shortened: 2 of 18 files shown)


jest.config.js
```diff
@@ -1,5 +1,13 @@
 module.exports = {
   verbose: true,
   testMatch: ['**/tests/**/*.js?(x)'],
-  testPathIgnorePatterns: ['tests/setup.js']
+  testPathIgnorePatterns: ['tests/setup.js', 'dist'],
```
kyleharper: When we add tests to the `lib` we need to exclude `dist` because they're built.

jest.config.js
```diff
@@ -1,5 +1,13 @@
...
+  moduleNameMapper: {
```
kyleharper: This fixes some module issues with the tests. It comes from the `react-dnd` documentation about testing.

lib/DnD/DndProvider.js
```diff
@@ -0,0 +1,37 @@
...
+    setIsDragging,
+    preview,
+    setPreview,
+    setFailurePreview
+  };
+
+  return (
+    <DragAndDropProvider backend={backend}>
+      <DndContext.Provider value={sharedState}>
+        {isDragging && <DragPreview>{failurePreview || preview}</DragPreview>}
```
kyleharper: This is what allows us to have a single custom drag preview for the whole provider. When we start dragging you'll see that `Draggable` sets the preview.

The failure preview is there so we can set that and when removed it'll revert to the previously set preview.

lib/DnD/DragPreview.js
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import { oneOfType, node, string } from 'prop-types';
+import { useDragLayer } from 'react-dnd';
+import { getItemStyles } from './helpers/getItemStyles';
+
+function DragPreview({ children }) {
+  const { initialOffset, currentOffset } = useDragLayer(monitor => ({
```
kyleharper: Our `DragPreview` uses the drag layer aspect of `react-dnd`, we just need to set the offsets.

This was actually retrieved from an example used on their site, I just used the idea in a different way to achieve our custom preview 😄

lib/DnD/Draggable.js
```diff
@@ -0,0 +1,45 @@
...
+    collect: monitor => ({
+      itemType: monitor.getItemType(),
+      isDragging: !!monitor.isDragging()
+    })
+  });
+
+  return (
+    <>
+      {preview && (
+        <DragPreviewImage
```
kyleharper: This disables the default drag preview.

It took me some time to figure it out but it turns out you can only set an image to replace the default preview. However, if you set it to an empty image then the preview is disabled (I saw we were doing this in the app 👍 )

lib/DnD/stories/DnDStory.js
```diff
@@ -0,0 +1,19 @@
...
+import { DndDropNotification } from './DndDropNotification';
+import { ItemDndMock } from './ItemDndMock';
+import { DndProvider } from '../DndProvider';
+import { FolderDndMock } from './FolderDndMock';
+
+const stories = storiesOf('Components', module);
+
+stories.add('Drag and Drop', () => (
+  <DndProvider backend={HTML5Backend}>
+    <FolderDndMock name={faker.commerce.productName()} showToggle open>
```
kyleharper: I thought we might want to create more examples in the future so I added an item and folder mock.

package.json
```diff
@@ -76,6 +76,7 @@
   "license": "MIT",
   "dependencies": {
     "@babel/polyfill": "^7.0.0",
+    "@storybook/addon-actions": "^5.1.9",
```
kyleharper: moved due to eslint

lib/DnD/Droppable.js
```diff
@@ -0,0 +1,30 @@
+import { useDrop } from 'react-dnd';
+import { func, node } from 'prop-types';
+
+function Droppable({ children, acceptDragTypes, onDrop, canDropChecker }) {
+  const [collected, defineDropRef] = useDrop({
+    accept: acceptDragTypes,
+    drop: onDrop,
+    canDrop: canDropChecker,
+    collect: monitor => ({
```
kyleharper: I quite like being specific with what we collect. If you ever needed a different implementation then you can always use the hook in your component 👍

kyleharper: I'll assume that we'll add / alter this as we use it in the hierarchy view.

lib/DnD/helpers/getItemStyles.js
```diff
@@ -0,0 +1,18 @@
+function getItemStyles(initialOffset, currentOffset) {
+  if (!initialOffset || !currentOffset) {
+    return {
+      display: 'none'
+    };
+  }
+
+  const { x, y } = currentOffset;
+
+  const transform = `translate(${x}px, ${y}px)`;
```
AmeeMorris: I had to tweak this a few times depending on what we were previewing, we may need to make this configurable in the future

lib/DnD/Draggable.js
AmeeMorris: Ha yeah, this is a little weird quirk

lib/DnD/Droppable.js
AmeeMorris: We will most likely need hover in here too, we use that heavily across the app

lib/DnD/DragPreview.js
```diff
@@ -0,0 +1,36 @@
...
+        height: '100%'
+      }}
+    >
+      <div style={getItemStyles(initialOffset, currentOffset)}>{children}</div>
+    </div>
+  );
+}
+
+DragPreview.propTypes = {
+  children: oneOfType([node, string]).isRequired
```
altaudio: `node` type includes string

kyleharper: @altaudio this has been altered, thanks for the feedback 👍

PR #540, opened 2019-08-09
Adding key to user search

DavidHollins6:
### 👀 Overview
User search map needs a key! This also removes a warning in the filter tests.


lib/UserSearch/UserSearchList.js
```diff
@@ -24,7 +24,7 @@ const UserSearchList = ({
   return (
     <ul className="user-search-list h-padding-clear h-margin-clear">
       {users.map(user => (
-        <li className="user-search-list__user">
+        <li className="user-search-list__user" key={user.id}>
```
altaudio: Do you think it's worth doing something like
```suggestion
        <li className="user-search-list__user" key={`user-search-list__user-${user.id}`}>
```

I'm thinking that if we render another list of users on the page for a different feature, someone may unknowingly do `key={user.id}`. Then we have two component with the same key.

kyleharper: @altaudio keys don't work in that way;

https://reactjs.org/docs/lists-and-keys.html#keys-must-only-be-unique-among-siblings

Keys only need to be unique among their siblings not globally.

kyleharper: Unless I've understood keys incorrect or those docs are throwing me off. Let me know if I've misunderstood.

altaudio: No that looks correct! Thanks for that, I didn't know that 👍

PR #543, opened 2019-08-13
Add renaming to Item and Folder rows!

kyleharper:
Add renaming pattern to folder and item rows.

## 💬 Description
We're missing renaming on item and folder rows and this PR adds it in. It also resolves an issue we've been having where the editing state doesn't span the entire width of the container it is in.

Some little extra added wins are in this PR too like width help classes!

## 📹 GIF (optional)
https://cl.ly/38bef1ebbbd2

## 🚪 Start Points
The code 😄 

## ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook


lib/EditableTextWrapper/styles.scss
```diff
@@ -42,6 +43,8 @@
   padding: 0;
   margin: 0;
   width: 100%;
+  background: none;
```
kyleharper: currently uses the default background colour.

lib/EditableTextWrapper/styles.scss
```diff
@@ -8,6 +8,7 @@
 .editable-text__wrapper {
   display: flex;
   margin: 0 0 0.25em;
+  width: 100%;
```
AmeeMorris: Make sure this doesn't cause any issues on the app!

There's a few places where you've got the 100% width set like this instead of a helper class. Just curious to know where you draw the line of using a helper vs using css? 

Personally I think helpers are more of a external (i.e. not used in ui) tool, since in my head they're mostly overrides that I want to use on UI components that im using. If this used the `h-width-100` helper, I'm not sure I would be able to override it with a `h-width-80` helper elsewhere.

stories/webapp/hierarchy/HierarchyFolderRow.js
```diff
@@ -1,20 +1,34 @@
...
+  const name = faker.commerce.department();
+
+  return (
+    <FolderRow
+      className="h-margin-bottom-half"
+      metaText={`${childCount} items`}
+      name={
+        <EditableTextWrapper
+          value={name}
+          className="h-margin-clear h-width-100"
```
AmeeMorris: Isnt this already 100% width https://github.com/gathercontent/gather-ui/pull/543/files#diff-09384d7eb14c24294d2ea796c6ea29deR11 ?

styles/helpers/_layout.scss
```diff
@@ -36,6 +36,19 @@ $subProps: (none, block, inline-block, flex, inline-flex, inline);
   }
 };
 
+/* ==========================================================================
+   Widths
+   ========================================================================== */
+
+$props: ('width');
+$subProps: ('5', '10', '15', '20', '25', '30', '35', '40', '45', '50', '55', '60', '65', '70', '75', '80', '85', '90', '95', '100');
```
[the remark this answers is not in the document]
kyleharper: Yeah true I'll limit it down some more. Maybe 10's will do 👍

stories/webapp/hierarchy/HierarchyFolderRow.js
kyleharper: 💯

lib/EditableTextWrapper/styles.scss
kyleharper: It's a good question.

I've overused them here I think. I'll stick to how we've used them before and add some fixed styles. I think because I added them I overly used them 😄

styles/helpers/_layout.scss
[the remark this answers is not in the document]
kyleharper: I actually just stook with `100` in the end 👍

PR #545, opened 2019-08-13
Add windowing components

kyleharper:
## 👀 Overview
Add windowing UI components for rendering a list in a performant manner.

## 💬 Description
When rendering a large number of items in a list we need to apply what is called virtualisation or windowing to tackle performance issues.

React mention the topic in the docs: https://reactjs.org/docs/optimizing-performance.html#virtualize-long-lists

---

I and @DavidHollins6 have done quite a bit of work to make sure we get a nice clean graph during the scroll events. The following graph shows the performance graph.

graph

---

The tests are pretty good too.

## 🔗 Links

- https://reactjs.org/docs/optimizing-performance.html#virtualize-long-lists

## 🚪 Start Points

Comments are a pretty good start, the topic is complex and large.

## ✅ Checklist
- [x] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook


lib/Windowing/InViewList.js
```diff
@@ -0,0 +1,67 @@
+import React, { useContext, useEffect, useState } from 'react';
+import { func, node, number } from 'prop-types';
+import { WindowingContext } from './WindowingProvider';
+
+function InViewList({ items, offset, children, itemHeight }) {
+  const [renderIndexes, setRenderIndexes] = useState({});
```
kyleharper: `renderIndexes` is the start and end index for what is "in view". E.g. `[0, 10]` will say to render the first 10 items in the overall collection.

lib/Windowing/InViewList.js
```diff
@@ -0,0 +1,67 @@
...
+    index + offset >= items.length ? items.length : index + offset;
+
+  const getRenderIndexes = () => {
+    const startBoundary = scrollTop;
+    const endBoundary = startBoundary + height;
+
+    const startIndex = Math.round(startBoundary / itemHeight);
+    const endIndex = Math.round(endBoundary / itemHeight);
+
+    const startWithOffset = subtractFromStartIndex(startIndex);
```
kyleharper: `subtractFromStartIndex` handles the "pre-mounting" aspect of the render index setting. This makes sure we mount the previous X items.

The other way is handled to so `addToEndIndex` adds the same offset to make sure we don't get blank rows appearing as you scroll.

lib/Windowing/InViewList.js
```diff
@@ -0,0 +1,67 @@
...
+    const endWithOffset = addToEndIndex(endIndex);
+
+    setRenderIndexes({
+      start: startWithOffset,
+      end: endWithOffset
+    });
+  };
+
+  useEffect(() => {
+    getRenderIndexes();
```
kyleharper: updates the render indexes every time the `height` or `scroll` is updated which in the provider is updates on `resize` and `scroll`.

lib/Windowing/InViewList.js
DavidHollins6: 👍 This function looks really clean now

lib/Windowing/tests/WindowingProvider.spec.js
```diff
@@ -0,0 +1,73 @@
...
+  fireEvent,
+  waitForElement,
+  queryByTestId
+} from '@testing-library/react';
+import { React } from '../../../tests/setup';
+import { WindowingProvider, InViewList, InViewListItem } from '../..';
+
+jest.mock('debounce', () => fn => fn);
+
+describe('WindowingProvider', () => {
```
DavidHollins6: These tests are great. The fact we can specify `scrollTop: 200` and know exactly what will be rendered and test is so powerful

lib/Windowing/tests/WindowingProvider.spec.js
```diff
@@ -0,0 +1,73 @@
...
+
+  it('renders 15 ({ start: 20, end: 35 }) in-view items', async () => {
+    const { getByTestId, container } = render(createWrapper());
+    await waitForElement(() => [
+      getByTestId('test-id-1'),
+      getByTestId('test-id-15')
+    ]);
+
+    fireEvent.scroll(getByTestId('windowing-scroller'), {
+      target: { scrollTop: 200 }
```
AmeeMorris: hey this is cool!

lib/Windowing/WindowingProvider.js
```diff
@@ -0,0 +1,50 @@
...
+        setScrollTop(eventNode.scrollTop);
+      }, debounceTimer);
+
+      elem.addEventListener('scroll', () => debouncedSetScrollTop(elem));
+      elem.addEventListener('resize', () => debouncedSetScrollTop(elem));
+    }
+  }, []);
+
+  return (
+    <WindowingContext.Provider value={{ height, scrollTop }}>
```
AmeeMorris: hmm is this more a getting an elements height & scrollTop rather than a windowing provider? (i get that you digest those for windowing, but this isnt doing any windowing logic).

I wonder if it could be more useful as a broader getElementBoundsProvider which returns the element bounds and positioning?

kyleharper: Good point, with windowing I think it's better to keep it all bundled together for now.

We actually have a `RefValues` like component in our consuming app...maybe it needs to be brought over to this repo and seen if we can use it here.

lib/Windowing/InViewList.js
```diff
@@ -0,0 +1,67 @@
...
+
+function InViewList({ items, offset, children, itemHeight }) {
+  const [renderIndexes, setRenderIndexes] = useState({});
+  const { scrollTop, height } = useContext(WindowingContext);
+
+  const subtractFromStartIndex = index =>
+    index - offset <= 0 ? 0 : index - offset;
+
+  const addToEndIndex = index =>
+    index + offset >= items.length ? items.length : index + offset;
```
altaudio: This is the same as 
```suggestion
    Math.min(items.length, index + offset);
```
which you could probably just stick on line 23

lib/Windowing/InViewList.js
```diff
@@ -0,0 +1,67 @@
+import React, { useContext, useEffect, useState } from 'react';
+import { arrayOf, func, number, shape } from 'prop-types';
+import { WindowingContext } from './WindowingProvider';
+
+function InViewList({ items, offset, children, itemHeight }) {
+  const [renderIndexes, setRenderIndexes] = useState({});
+  const { scrollTop, height } = useContext(WindowingContext);
+
+  const subtractFromStartIndex = index =>
+    index - offset <= 0 ? 0 : index - offset;
```
altaudio: This is the same as 
```suggestion
    Math.max(index - offset, 0);
```
which you could probably just stick on line 22

lib/Windowing/InViewListItem.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import { node, number, shape } from 'prop-types';
+
+function InViewListItem({ children, renderIndexes, index, baseStyle }) {
+  const top = (renderIndexes.start + index) * 50;
```
altaudio: Should this be?
```suggestion
  const top = (renderIndexes.start + index) * itemHeight;
```

This is the only RFC comment in this review.

lib/Windowing/WindowingProvider.js
altaudio: Agreed!

PR #546, opened 2019-08-13
Move SelectedObjectsProvider to gather-ui

DavidHollins6:
### 👀 Overview
As part of the Hierarchy work there is a requirement for being able to select Items/Folders so a user can possibly perform bulk actions (At some point).
### 💬 Description
It is the same Provider as is in `app`, however the `setSelected` function has been renamed to `updateSelected`, as it makes more sense and was conflicting with the hooks function.

It also uses hooks now, instead of a class component. The tests now use react-testing-library, an element is passed in as a child to the consumer, the tests interact with the provider through that element. I tried to get close to how a provider would be used in the real world, and this made the most sense. 

This card will not be adding it to storybook, it will be added when the implementation of the Provider happens.

### ✅ Checklist
- [x] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook


DavidHollins6:
> This is great!
> 
> My only feedback is to create a `lib` story showing how we use this 👍

There's now a story!

PR #555, opened 2019-08-16
Feature/select custom hook

DavidHollins6:
### 👀 Overview
This is a new custom hook to handle selection of objects! It consumes the context for you and provides you with some nice values to use. 

There are also tests for this custom hook, using a newly installed package `@testing-library/react-hooks`, that is part of the testing-library family!

WARNING: This PR does not include all the usage of this hook, that's going in another PR that's a bit of a mess at the moment. This PR is strictly to show you all a cool custom hook that's been made. 

### 💬 Description
The hook will provide you with the following bools:

`isSelected`, which tells you whether the given id is selected.
`isDisabled`, same as above but whether it is disabled.
`isHovered`, again, but whether the id is intended to be selected.

It will also create some nice functions to be used:

`handleClick`, this will handle all of the selecting for you!
`handleMouseEnter` & `handleMouseLeave`, is for setting the intended to be selected (Hovering).

### 🚪 Start Points
`useObjectSelector.js`

### ✅ Checklist
- [x] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [x] Added to storybook

### DIFF
--- a/lib/SelectedObjectsProvider/index.js
+++ b/lib/SelectedObjectsProvider/index.js
@@ -6,7 +6,7 @@ const SelectedObjectsContext = React.createContext({});
 
 const SelectedObjectsProvider = ({ children }) => {
   const [selected, setSelected] = useState([]);
-  const [lastInteracted, setLastInteracted] = useState(null);
+  const [intendedToSelect, setIntendedToSelect] = useState([]);
   const [currentSelectedType, setCurrentSelectedType] = useState(null);
 
   const updateSelected = (id, type) => {
@@ -17,7 +17,6 @@ const SelectedObjectsProvider = ({ children }) => {
     } else {
       newSelected = selected.filter(existingId => existingId !== id);
     }
-    setLastInteracted(id);
     setSelected(newSelected);
 
     if (!newSelected.length) {
@@ -29,22 +28,19 @@ const SelectedObjectsProvider = ({ children }) => {
 
   const clear = () => {
     setSelected([]);
-    setLastInteracted(null);
     setCurrentSelectedType(null);
   };
 
-  const selectMultiple = (ids, type, id = null) => {
+  const selectMultiple = (ids, type) => {
     setSelected([...new Set(selected.concat(ids))]);
-    setLastInteracted(id);
     setCurrentSelectedType(type);
   };
 
-  const deselectMultiple = (ids, id = null) => {
+  const deselectMultiple = ids => {
     const newSelected = selected.filter(
       selectedId => !ids.includes(selectedId)
     );
     setSelected(newSelected);
-    setLastInteracted(id);
 
     if (!newSelected.length) {
       setCurrentSelectedType(null);
@@ -53,12 +49,13 @@ const SelectedObjectsProvider = ({ children }) => {
 
   const sharedState = {
     selected,
-    lastInteracted,
     updateSelected,
     clear,
     selectMultiple,
     deselectMultiple,
-    currentSelectedType
+    currentSelectedType,
+    setIntendedToSelect,
+    intendedToSelect
   };
   return (
     <SelectedObjectsContext.Provider value={sharedState}>
--- a/lib/SelectedObjectsProvider/useObjectSelector.js
+++ b/lib/SelectedObjectsProvider/useObjectSelector.js
@@ -1,0 +1,33 @@
+import { useContext } from 'react';
+import { SelectedObjectsContext } from './index';
+
+export function useObjectSelector(
+  id,
+  type,
+  idsToSelect,
+  disabledCallback = () => false
+) {
+  const {
+    selected,
+    selectMultiple,
+    deselectMultiple,
+    updateSelected,
+    currentSelectedType
+  } = useContext(SelectedObjectsContext);
+
+  const isSelected = selected.indexOf(id) !== -1;
+  const isDisabled = disabledCallback(currentSelectedType, isSelected);
+
+  const handleClick = idsToSelect
+    ? () =>
+        isSelected
+          ? deselectMultiple([...new Set([id, ...idsToSelect])], type)
+          : selectMultiple([...new Set([id, ...idsToSelect])], type)
+    : () => updateSelected(id);
+
+  return {
+    isSelected,
+    handleClick: isDisabled ? () => {} : handleClick,
+    isDisabled
+  };
+}
--- a/lib/SelectedObjectsProvider/useObjectSelector.js
+++ b/lib/SelectedObjectsProvider/useObjectSelector.js
@@ -12,22 +12,38 @@ export function useObjectSelector(
     selectMultiple,
     deselectMultiple,
     updateSelected,
-    currentSelectedType
+    currentSelectedType,
+    intendedToSelect,
+    setIntendedToSelect
   } = useContext(SelectedObjectsContext);
 
   const isSelected = selected.indexOf(id) !== -1;
   const isDisabled = disabledCallback(currentSelectedType, isSelected);
+  const isHovered = intendedToSelect.indexOf(id) !== -1;
 
-  const handleClick = idsToSelect
-    ? () =>
-        isSelected
-          ? deselectMultiple([...new Set([id, ...idsToSelect])], type)
-          : selectMultiple([...new Set([id, ...idsToSelect])], type)
-    : () => updateSelected(id);
+  const getIds = (additionalIds = []) => [
+    ...new Set([id, ...idsToSelect, ...additionalIds])
+  ];
+
+  const handleClick =
+    idsToSelect.length > 1
+      ? () =>
+          isSelected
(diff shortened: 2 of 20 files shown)


stories/webapp/hierarchy/HierarchyFolderRow.js
```diff
@@ -49,14 +48,7 @@ function HierarchyFolderRow({ data, nameForm, children, open }) {
...
       showToggle
-      onClick={() => {
-        if (!isDisabled) {
-          return isLevelSelected
-            ? deselectMultiple([data.id, ...childIds], 'folder')
-            : selectMultiple([data.id, ...childIds], 'folder');
-        }
-        return null;
-      }}
+      onClick={handleClick}
```
AmeeMorris: I feel like handleClick is a bit too vague of a name, just calling it along the lines of updateSelected would be better

lib/SelectedObjectsProvider/useObjectSelector.js
```diff
@@ -0,0 +1,49 @@
...
+    ...new Set([id, ...idsToSelect, ...additionalIds])
+  ];
+
+  const handleClick =
+    idsToSelect.length > 1
+      ? () =>
+          isSelected
+            ? deselectMultiple(getIds(), type)
+            : selectMultiple(getIds(), type)
+      : () => updateSelected(idsToSelect[0]);
```
AmeeMorris: should this just be `() => updateSelected(id)`?

lib/SelectedObjectsProvider/useObjectSelector.js
```diff
@@ -0,0 +1,49 @@
+import { useContext } from 'react';
+import { SelectedObjectsContext } from './index';
+
+export function useObjectSelector(
+  id,
+  type,
+  idsToSelect,
+  disabledCallback = () => false
```
AmeeMorris: disabledCallback makes it seem like it fires when a object is disabled. Maybe `isDisabledChecker` or something might be better?

lib/SelectedObjectsProvider/index.js
```diff
@@ -6,7 +6,7 @@ const SelectedObjectsContext = React.createContext({});
 
 const SelectedObjectsProvider = ({ children }) => {
   const [selected, setSelected] = useState([]);
-  const [lastInteracted, setLastInteracted] = useState(null);
```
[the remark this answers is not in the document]
DavidHollins6: We decided as they weren't needed yet then we should not include them.

lib/SelectedObjectsProvider/index.js
```diff
@@ -53,12 +49,13 @@ const SelectedObjectsProvider = ({ children }) => {
...
   const sharedState = {
     selected,
-    lastInteracted,
     updateSelected,
     clear,
     selectMultiple,
     deselectMultiple,
-    currentSelectedType
+    currentSelectedType,
+    setIntendedToSelect,
```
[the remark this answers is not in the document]
DavidHollins6: Yep, these are the id's that would be selected if the person clicked with the mouse.

stories/webapp/hierarchy/HierarchyFolderRow.js
DavidHollins6: 👍

lib/SelectedObjectsProvider/useObjectSelector.js
DavidHollins6: We wanted to seperate the first argument, `id`, with the ones being selected, `idsToSelect`. This is so we're certain that it will only selected the id's that are in `idsToSelect` and nothing more.

lib/SelectedObjectsProvider/useObjectSelector.js
DavidHollins6: 👍

lib/SelectedObjectsProvider/useObjectSelector.js
AmeeMorris: Forgive me if im missing something! is individual selection not meant to work for `stories/webapp/hierarchy/HierarchyFolderRow.js` and only for its parent folder? , cos you're not passing over any idsToSelect (just an empty array) so that updateSelected will have nothing to pass up

kyleharper: Correct me if I am incorrect @DavidHollins6 but didn't we decide to just use `idsToSelect` rather than the complexity of figuring out "why do we only use the first key in this scenario"?

lib/SelectedObjectsProvider/index.js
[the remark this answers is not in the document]
kyleharper: Yeah, currently we have no hover state @AmeeMorris - which isn't great.

Not only that but in the hierarchy view, we have instances where a single selection can select multiple elements to show to make this easier to understand (UX wise) we're adding a hover state.

We may move to CSS hovers in the future but there will be structural complications there as these rows are bound to include additional wrappers in the near future.

E.g. windowing will do this.

lib/SelectedObjectsProvider/index.js
```diff
@@ -6,7 +6,7 @@ const SelectedObjectsContext = React.createContext({});
 
 const SelectedObjectsProvider = ({ children }) => {
```
kyleharper: Something that is starting to bug me with this name is why did we add "objects" to the name? 😂 

Why not something like `SelectionProvider`?

tests/SelectedObjectsProvider/useObjectSelector.spec.js
```diff
@@ -0,0 +1,86 @@
+import React from 'react';
+import { renderHook, act } from '@testing-library/react-hooks';
+import { useObjectSelector } from '../../lib/SelectedObjectsProvider/useObjectSelector';
+import { SelectedObjectsProvider } from '../../lib/SelectedObjectsProvider';
+
+describe('useObjectSelector', () => {
+  it('should select the given ids', () => {
+    const id = 1;
+    // eslint-disable-next-line react/prop-types
```
kyleharper: 👍

lib/SelectedObjectsProvider/useObjectSelector.js
```diff
@@ -0,0 +1,49 @@
...
+    intendedToSelect,
+    setIntendedToSelect
+  } = useContext(SelectedObjectsContext);
+
+  const isSelected = selected.indexOf(id) !== -1;
+  const isDisabled = disabledCallback(currentSelectedType, isSelected);
+  const isHovered = intendedToSelect.indexOf(id) !== -1;
+
+  const getIds = (additionalIds = []) => [
+    ...new Set([id, ...idsToSelect, ...additionalIds])
```
altaudio: > We wanted to seperate the first argument, id, with the ones being selected, idsToSelect. This is so we're certain that it will only selected the id's that are in idsToSelect and nothing more.

This is a cool idea that gives you a nice bit of flexibility!

Should this line therefore be?
```suggestion
    ...new Set([...idsToSelect, ...additionalIds])
```
as it currently selects the `id` and `idsToSelect` on line 32 and 33?

lib/SelectedObjectsProvider/useObjectSelector.js
```diff
@@ -0,0 +1,49 @@
...
+    currentSelectedType,
+    intendedToSelect,
+    setIntendedToSelect
+  } = useContext(SelectedObjectsContext);
+
+  const isSelected = selected.indexOf(id) !== -1;
+  const isDisabled = disabledCallback(currentSelectedType, isSelected);
+  const isHovered = intendedToSelect.indexOf(id) !== -1;
+
+  const getIds = (additionalIds = []) => [
```
altaudio: I'd be tempted to get rid of this arg if it's not used, just to reduce complexity.

lib/SelectedObjectsProvider/useObjectSelector.js
altaudio: You could users our `unique` helper here, from `helpers/arrays`.

lib/SelectedObjectsProvider/useObjectSelector.js
```diff
@@ -0,0 +1,49 @@
...
+      ? () =>
+          isSelected
+            ? deselectMultiple(getIds(), type)
+            : selectMultiple(getIds(), type)
+      : () => updateSelected(idsToSelect[0]);
+
+  const handleMouseEnter = () => setIntendedToSelect(idsToSelect);
+
+  const handleMouseLeave = () =>
+    setIntendedToSelect(intendedToSelect.filter(s => !idsToSelect.includes(s)));
```
altaudio: Unfortunately includes isn't supported by IE (damn you IE 😾).

you could check the transpiled code to see if we have a polyfill for this, but I'm pretty sure last time I checked we didn't.

If we do, you could also replace the uses of `array.indexOf(id) !== -1`, which is our workaround for not being able to use `includes`.

PR #556, opened 2019-08-16
Feature/hierarchy item hover

DavidHollins6:
### 👀 Overview
Here we are implementing the custom hook for hovering over objects. 

There is also some refactoring to make selecting a bit more intuitive, along with heights of each row have been fixed.


lib/FolderRow/FolderRow.js
```diff
@@ -27,10 +29,15 @@ const FolderRow = ({
     'folder-row--highlight': highlight
   });
 
+  const folderNameClassNames = cx('folder-row__name', {
+    'h-width-100': isEditing
```
kyleharper: 👍 so the input spans correctly but not until someone is editing so hover isn't affected.

styles/helpers/_selected.scss
```diff
@@ -2,3 +2,13 @@
   border: 1px solid $primary-blue;
   box-shadow: inset 0 0 0 1px $primary-blue;
 }
```
kyleharper: Add section comment blocks here 👍

PR #559, opened 2019-08-23
Add DIR prop to fields

kyleharper:
## 👀 Overview
Showing fields in RTL format should switch the field label position.

## 💬 Description
When we display content in RTL the field labels don't move which makes the experience very difficult for those using it. This PR adds a `dir` prop and by default sets it to ltr. 

It also adds the dir attr to the surrounding div which we do in the consuming app so we'll be able to remove that when this is released.

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook


lib/Field/stories/FieldStory.js
```diff
@@ -0,0 +1,93 @@
...
+    clickHandler: action('action clicked')
+  },
+  {
+    text: 'test action 2',
+    clickHandler: action('action 2 clicked')
+  }
+];
+
+storiesOf('Components', module).add('Field', () => {
+  const dir = text('Text direction', 'ltr');
```
DavidHollins6: It could be cool if we used radios here instead of letting the user type in the dir.

https://www.npmjs.com/package/@storybook/addon-knobs#radio-buttons

kyleharper: done 👍

lib/Field/index.js
```diff
@@ -9,7 +9,8 @@ import ExpandingTextArea from '../ExpandingTextArea';
 import EditableTextWrapper from '../EditableTextWrapper';
 
 const Field = props => {
-  const additionalClasses = cx({
+  const classNames = cx('field', {
```
altaudio:
```suggestion
  const classNames = cx(`field field--${props.dir}`, {
```
as `props.dir` will always be true?

kyleharper:
@altaudio all done with the feedback, please review.

PR #565, opened 2019-09-02
Windowing with nested objects

DavidHollins6:
### 👀 Overview
Windowing became a lot more complex when we had to handle nested lists of children and parents. This PR now handles it!

I advise that you review this in your IDE and have a play with it in storybook. The diff provided by git does not really give the whole story of what's going on here. The story can be found inside `Windowing`, called `Nested`.

### 💬 Description
How it works:

Each parent has holds two values (`offset`, `length`) that are called their `acceptedRange`. This is basically the range of their children in terms of the greater structure.

Whenever a scroll happens, the parent will check whether any of their children's indexes (the `acceptedRange`) are inside the indexes that are supposed to be rendered (`renderIndexes`), if that is the case then the parent will render the children that intersect both of these ranges of indexes. 

### 🚪 Start Points
`InViewList`

### ✅ Checklist
- [x] Tests written
- [x] Browser tested
- [ ] Added to documentation
- [x] Added to storybook


lib/Windowing/WindowingProvider.js
```diff
@@ -10,16 +10,16 @@ const WindowingProvider = ({
   debounceTimer,
   itemHeight,
   itemsLength,
-  offset,
+  buffer,
```
AmeeMorris: nice renaming, makes it much clearer

lib/Windowing/InViewList.js
```diff
@@ -1,17 +1,58 @@
...
+
+  if (endOfIndexesToRender > endOfAcceptedRangeOfIndexes) {
+    return totalLength;
+  }
+
+  if (endOfIndexesToRender < endOfAcceptedRangeOfIndexes) {
+    return endOfIndexesToRender - renderIndexes.offset;
+  }
+
+  return -1;
```
altaudio: Have you guys seen an edge case where we need this final return?

If not, I would say get rid of it and have line 23 as the final return, without the if statement. Realistically a number can only be bigger, smaller or the same as another number. There aren't any other edge cases.

lib/Windowing/InViewList.js
```diff
@@ -1,17 +1,58 @@
...
+  if (endOfIndexesToRender === endOfAcceptedRangeOfIndexes) {
+    return endOfIndexesToRender;
+  }
+
+  if (endOfIndexesToRender > endOfAcceptedRangeOfIndexes) {
+    return totalLength;
+  }
+
+  if (endOfIndexesToRender < endOfAcceptedRangeOfIndexes) {
+    return endOfIndexesToRender - renderIndexes.offset;
```
altaudio: Maybe
```suggestion
    return renderIndexes.length;
```

DavidHollins6: Good point!

DavidHollins6:
Hello!

Just been working through @altaudio comments. It has ended up being a really nice refactor of that whole function. I have re-requested a review because it was quite a bit change.

PR #570, opened 2019-09-11
Adding Create Item action to hierarchy

DavidHollins6:
### 👀 Overview
A new action has been added to the folder row to create an item in that folder.
### 💬 Description
It works very similar to creating a folder, it just creates a new item instead!
### 🚪 Start Points
`HierarchyCollection`

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [x] Added to storybook

### DIFF
--- a/stories/webapp/hierarchy/ItemRow/NewHierarchyItemRow.js
+++ b/stories/webapp/hierarchy/ItemRow/NewHierarchyItemRow.js
@@ -1,0 +1,55 @@
+import React, { useState } from 'react';
+import { func, string } from 'prop-types';
+import { Icon, ItemRow } from 'lib';
+import { HierarchyRowForm } from '../HierarchyRowForm';
+import { HierarchyItemRow } from './HierarchyItemRow';
+
+const NewHierarchyItemRow = ({ id, removeItem }) => {
+  const [itemName, setItemName] = useState(null);
+  const [isEditing, setIsEditing] = useState(true);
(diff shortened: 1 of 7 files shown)


stories/webapp/hierarchy/ItemRow/NewHierarchyItemRow.js
```diff
@@ -0,0 +1,55 @@
...
+import { func, string } from 'prop-types';
+import { Icon, ItemRow } from 'lib';
+import { HierarchyRowForm } from '../HierarchyRowForm';
+import { HierarchyItemRow } from './HierarchyItemRow';
+
+const NewHierarchyItemRow = ({ id, removeItem }) => {
+  const [itemName, setItemName] = useState(null);
+  const [isEditing, setIsEditing] = useState(true);
+  const [isSaving, setIsSaving] = useState(false);
+  const [hasSaved, setHasSaved] = useState(false);
```
AmeeMorris: Im sure most of this would be replaced when it comes time to hook up to an api, but maybe a status state might be a bit cleaner? so:
`const [currentStatus, setStatus] = useState('');` then the status can be set to `IS_SAVING`, `IS_EDITING`, `HAS_SAVED`. Since it doesnt look like any two need to be true at any given time. Or perhaps just ` `, `IS_SAVING`, `HAS_SAVED` to get rid of having both isSaving and hasSaved states.

Not something you have to action, just felt a bit weird to me!

PR #572, opened 2019-09-17
Adding buttonLabel to EditableTextWrapper

DavidHollins6:
### 👀 Overview
This adds a prop to have some text inside the pencil button.

EDIT:

A label has also been added for the input that `EditableTextWrapper` renders. All inputs should have labels!
### 💬 Description
**Any** button that is just an icon should also have some text inside, but set to `visually-hidden`. This is so screen readers know what the button does.

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [x] Added to storybook

### DIFF
--- a/stories/components/EditableTextWrapper.js
+++ b/stories/components/EditableTextWrapper.js
@@ -4,36 +4,48 @@ import { action } from '@storybook/addon-actions';
 import EditableTextWrapper from '../../lib/EditableTextWrapper';
 import StoryItem from '../styleguide/StoryItem';
 
-const longText = 'Original text thats quite long and may well go over a single line, so its much safer to set the multiline prop, that way we render an ExpandingTextArea component instead of a normal text input';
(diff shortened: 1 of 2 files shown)


stories/components/EditableTextWrapper.js
```diff
@@ -4,36 +4,48 @@ import { action } from '@storybook/addon-actions';
...
-        title="EditableTextWrapper"
-        description="Wraps text in an editable wrapper"
+storiesOf('Components', module).add('EditableTextWrapper', () => (
+  <div>
+    <StoryItem
+      title="EditableTextWrapper"
+      description="Wraps text in an editable wrapper"
+    >
+      <EditableTextWrapper
+        buttonLabel="Editable text"
```
DavidHollins6: This is the only change in this file. Prettier had it's way with it.

PR #575, opened 2019-09-18
Accept an imported svg in Logo

AmeeMorris:
### 👀 Overview
As well as a url string, I wanted to be able to pass an imported SVG to display in the Logo component. This lets me do just that.

Also adding a max-height to the AvatarGroup dropdown, since if theres a load of Avatars it can easily go off screen. 

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook


AmeeMorris:
[Definitely think we need a but more of a brand presence on the shareable link UI](https://trello.com/c/tmue7BJu/560-definitely-think-we-need-a-but-more-of-a-brand-presence-on-the-shareable-link-ui)

PR #577, opened 2019-09-20
Nested windowing v2

DavidHollins6:
### 👀 Overview
As you may have heard, nested windowing is dead! 

The original implementation of windowing for nested lists was not performant. We were trying to render multiple lists which would over-calculate which items to render. After a few lists, it became sluggish and overcomplicated.

The old way has been thrown out and given us a new perspective. We have a single list of items, and render them all in a linear fashion. 

During this re-write, the `Windowing` components have gone through a big refactor. We now use the compound component pattern, with each component having their own responsibility.

I recommend checking out the branch and looking at the stories to review this, as the diff's here will be very difficult to parse. I have also created a thorough README to explain how windowing works and what it's components do, this can be found here: https://github.com/gathercontent/gather-ui/tree/nested-windowing-v2/lib/Windowing

### 🚪 Start Points
the branch `nested-windowing-v2`

### ✅ Checklist
- [x] Tests written
- [ ] Browser tested
- [x] Added to documentation
- [x] Added to storybook


lib/Windowing/README.MD
```diff
@@ -0,0 +1,124 @@
+# Windowing
+
+A selection of components to give you virtualisation.
+
+## How does it work?
+
+A `div` is rendered that is high enough to hold every item, it is simply calculated as follows:
```
kyleharper: Some content change suggestions;

- use "tall" rather than "high"
- remove the word simply 😄

lib/Windowing/README.MD
```diff
@@ -0,0 +1,124 @@
...
+
+Each `item` has to have the same height, making all of this maths a lot easier.
+
+This `div` is wrapped in a container, with a set height, the `containerHeight`. This container is our window to look
+at the long list of items, it restricts the number of items we can see at a time though!
+
+When a scrolling event is triggered on the long list of items we retrieve a value called `scrollTop`, this represents
+the number of pixels from the top of the list that we have scrolled.
+
+With all of these values we can calculate which items can be seen through the window. Here is a quick example of the
```
kyleharper: I wonder if this description helps? I wrote windowing and it confuses me a little...maybe we could simplify it by explaining key points e.g;

"How does it work? It calculates the number of components that are in view so you can render a subset of UI within a container. It does this by using a number of properties like `scrollTop`, `containerHeight` and the height of each component (this is currently fixed)."

The code explains the fact that we render a div etc...Thoughts?

lib/Windowing/README.MD
```diff
@@ -0,0 +1,124 @@
...
+`itemHeight`: 10px  
+`listHeight`: 400px  
+`containerHeight`: 100px  
+`scrollTop`: 50px
+
+`numberOfItemsVisible`: 10  
+`inViewWindowingIds`: [5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15]
+
+We can set a buffer to ensure the user experience is much nicer. A buffer is the amount of items we render outside of
+the window. This ensures that there won't be any "loading" items on the page, it will all happen in the buffer. Here is an example:
```
kyleharper: Content suggestion...change "loading" for unmounted or something.

lib/Windowing/README.MD
```diff
@@ -0,0 +1,124 @@
...
+This `div` is wrapped in a container, with a set height, the `containerHeight`. This container is our window to look
+at the long list of items, it restricts the number of items we can see at a time though!
+
+When a scrolling event is triggered on the long list of items we retrieve a value called `scrollTop`, this represents
+the number of pixels from the top of the list that we have scrolled.
+
+With all of these values we can calculate which items can be seen through the window. Here is a quick example of the
+window being at the top of the list.
+
+#### Example 1 - Beginning of list
```
kyleharper: It might be worth explaining that these examples are here to show the output data-wise? I mean, `usage` and stories show examples so I'm not too sure if these are readable to not.

kyleharper: Maybe it's just the way it's written by you could write it up like;

- Let's go through an example...here are my properties "{ show object }"
- Ok, without any scrolling here are the items to show...here is why.
- Ok, now lets scroll...here's the new IDS

No need to go into the actually implementation detail like - "here is why - let's go over the math" but I feel like it could be explained a little better than

-------

With a fresh pair of eyes do these examples help more than stories?

lib/Windowing/README.MD
```diff
@@ -0,0 +1,124 @@
...
+`containerHeight`: 100px  
+`buffer`: 5  
+`scrollTop`: 50px
+
+`numberOfItemsVisible`: 20 (10 + 5 (buffer top) + 5 (buffer bottom))  
+`inViewWindowingIds`: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20]
+
+## Usage
+
+Windowing was designed to be flexible, any of these compound components can be switched out for a custom one, all of the
```
kyleharper: Maybe it's worth simplifying this..."windowing uses context...go to town if you need anything specific for your solution...otherwise look at the stories.".

lib/Windowing/README.MD
```diff
@@ -0,0 +1,124 @@
...
+| containerHeight | number (px) | 0       | No       | The pixel height of the window.                                             |
+| debounceTimer   | number (ms) | 15      | No       | The debounce time in milliseconds for the scrolling update.                 |
+| buffer          | number      | 10      | No       | The number of items in the buffer.                                          |
+| allIds          | array       | N/A     | Yes      | An array holding the id of every item that windowing could possibly render. |
+
+##### Provides
+
+| Name                  | Type        | Description                                                            |
+| --------------------- | ----------- | ---------------------------------------------------------------------- |
+| scrollTop             | number      | The number of pixels that have been scrolled                           |
```
kyleharper: Might be worth checking some of these...do we need to pass them all into context?

lib/Windowing/README.MD
```diff
@@ -0,0 +1,124 @@
...
+| children | node | N/A     | Yes      | The children. |
+
+#### `<Windowing.Item />`
+
+The `Item` calculates it's `top` value based off it's index in the `allWindowingIds` and the items height.
+
+| Name                    | Type   | Default | Required | Description                                            |
+| ----------------------- | ------ | ------- | -------- | ------------------------------------------------------ |
+| children                | node   | N/A     | Yes      | The children.                                          |
+| relativeToRenderIndexes | bool   | true    | No       | DEPRECATED                                             |
```
kyleharper: This was never truly a thing right? Can it be deleted and we just push the repo to a minor version?

lib/Windowing/README.MD
DavidHollins6: There's a couple of props or so that aren't needed anymore, they were going to be removed as part of the clean up of the old version.

lib/Windowing/Windowing.js
```diff
@@ -1,25 +1,26 @@
...
+  const [inViewWindowingIds, setInViewWindowingIds] = useState([]);
 
   const subtractFromStartIndex = index => Math.max(index - buffer, 0);
 
-  const addToEndIndex = index => Math.min(itemsLength, index + buffer);
+  const addToEndIndex = index =>
+    Math.min(allWindowingIds.length, index + buffer);
 
   const getRenderIndexes = () => {
     const startBoundary = scrollTop;
```
altaudio: I feel like we are going a bit of a round about way to calculate the `length` on line 37. I don't think we need to calculate the `endWithBuffer` - this is a hangover from when we used the `endWithBuffer` value in a previous implementation.

To get `length` you can do `(height / itemHeight) + (buffer * 2)`. I think this method makes more sense, are we are no longer calculating the `endWithBuffer` that we don't need.

I'm pretty sure this works, but worth checking!

PR #592, opened 2019-10-04
Various styling fixes for hierarchy

DavidHollins6:
### 💬 Description
This includes multiple styling fixes that came as feedback to the hierarchy. They are originally done inside storybook and will be ported to the app soon.

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [x] Added to storybook

### DIFF
--- a/lib/Hierarchy/styles.scss
+++ b/lib/Hierarchy/styles.scss
@@ -1,0 +1,7 @@
+.hierarchy-line {
+  content: '';
+  position: absolute;
+  bottom: 0;
+  width: 2px;
+  background: $neutral-light;
+}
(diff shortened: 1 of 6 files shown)


lib/Hierarchy/styles.scss
```diff
@@ -0,0 +1,7 @@
+.hierarchy-line {
+  content: '';
+  position: absolute;
+  bottom: 0;
+  width: 2px;
+  background: $neutral-light;
+}
```
DavidHollins6: Added this in ready for the other lines work, I just didn't want to do another PR for this one class. Also suggestions for the name of the class is welcome.

PR #598, opened 2019-10-14
Moving overrides into ui, topbar full height helper

AmeeMorris:
### 💬 Description
Pretty much as the title says!


styles/helpers/_layout.scss
```diff
@@ -101,3 +101,14 @@ $subProps: (stretch, center, flex-start, flex-end, baseline, initial, inherit);
 .space-between {
   justify-content: space-between;
 }
+
+/* ==========================================================================
+   Heights
+   ========================================================================== */
+
+.full-height-minus-topbar {
```
altaudio: OOooh like this one 👍

DavidHollins6: This would be great for the hierarchy!

PR #611, opened 2019-10-25
Creating a class for a dragged preview

DavidHollins6:
### 💬 Description
We need some styles to apply to the drag preview, since they now are in separate components. I started trying to use the existing classes but there were too many issues with the width.

### DIFF
--- a/lib/FinderNavigation/styles.scss
+++ b/lib/FinderNavigation/styles.scss
@@ -54,6 +54,16 @@ $finder-indent-level-support: 10 !default;
   width: 100%;
 }
 
+.finder__item--drag-wrapper {
+  background-color: white;
+  display: inline-flex;
+  align-items: center;
+  height: 35px;
+  line-height: 35px;
+  border-radius: $layout-spacing-base / 4;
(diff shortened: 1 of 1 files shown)


lib/FinderNavigation/styles.scss
```diff
@@ -54,6 +54,16 @@ $finder-indent-level-support: 10 !default;
   width: 100%;
 }
 
+.finder__item--drag-wrapper {
```
kyleharper: I'm not too sure this name encapsulates what this is really doing. Unless I am reading it wrong. `finder__item` is the entire row including sub children (looking at the story).

So does this add a drag wrapper style around the whole row and sub rows? Or is it just for the individual row?

------

Maybe if it was added to the story it'd be easier to understand...which then brings to the next piece of feedback;

Is this standalone class used within being bound to a prop? If so it'll be difficult to communicate how we use it in a components namespace I think. For example;

```
<FinderItem.Content isDraggable /> // has a class of .finder__item--draggable within the component
```

I believe this communicates the intent much easier and the story can reflect that, thoughts?

PR #624, opened 2019-11-19
Changing hierarchy row borders

DavidHollins6:
### 💬 Description
Preface - Everything in the hierarchy is based on height. The changes in this PR might seem odd, but they were done in a way to not affect the heights of anything!

Now to the actual PR. Borders around rows in the hierarchy used to be as followed: 

`border: 1px solid grey`

Then when we wanted to apply a 2px border for a selected style we would do the following:

`border: 1px solid blue`
`box shadow: 0 0 1px blue`

This was done to preserve the height and prevent the contents from shifting around when border thickness changed.

This option is no longer viable because there'ss a fancy new border type for - `When my parent has been selected, give me a dashed blue border`. This wasn't possible because you couldn't have a dashed box shadow. Thus we're at this solution, which is an absolute pseudo element giving a border!

I'm sorry...

### 🔗 Links
https://trello-attachments.s3.amazonaws.com/5ca62742c7569644c3fefeba/5dd26b92cad7fe5d03c0fc77/b3039403f6cf344c56cdb64760d668aa/image.png
### 🚪 Start Points
`styles/tools/_mixins.scss`

### DIFF
--- a/lib/FolderRow/styles.scss
+++ b/lib/FolderRow/styles.scss
@@ -22,10 +22,9 @@ $folder-row-spacing: $layout-spacing-base/2 !default;
 .folder-row__inner {
   padding: $folder-row-spacing;
   background: $neutral-lightest;
-  border-radius: 4px;
-  border: 1px solid $neutral-light;
   position: relative;
   margin: 0 0 $layout-spacing-base/2;
+  @include before-border(1px, solid, $neutral-light, 4px);
 
   + .folder-row {
     margin-top: $folder-row-spacing;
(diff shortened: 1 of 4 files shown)


AmeeMorris:
Could we not take away some padding in order to have a 2px border without changing the height instead?

DavidHollins6:
> Could we not take away some padding in order to have a 2px border without changing the height instead?

I had tried that. But because the border thickness changes from 1px to 2px, the contents inside get juggled around!

PR #633, opened 2019-12-09
Adds title tags to SVGs

altaudio:
### 💬 Description
Adds some title tags to SVGs, for use in testing.

Notes
- Removes `svg-url-loader` package as it isn't used. (storybook uses `raw-loader`, gather-ui uses `babel-plugin-inline-react-svg`)

#### TODO
- export icon name needs changing to export1 in app (clashes with export reserved word)
### 🔗 Links
### 📹 GIF (optional)
### 🚪 Start Points
### 👫 Related PRs (optional)

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook


AmeeMorris:
Also if this is for tests I've been setting the Icon's text prop, and the hideText prop to true and finding/targeting that

PR #640, opened 2019-12-30
FiX/Dropdown Auto Position

altaudio:
#### Bug
We had a bug reported where the the `<DueDatePicker />` dropdown would position itself off the bottom of the screen. This dropdown had the prop `autoPosition=true`, so it was supposed to position itself on the screen.

[Video, it really did happen!](https://cl.ly/f3e3c0d8165e)
#### Cause
The `<DropdownContent />` component calculates whether its contents overflows the main window, by saying:
>  "My content starts at ***position X***, and has a ***height of Y*** pixels, therefore I know that if ***X + Y > inner window height***, then my content is off the bottom of the screen, and I need to reposition it!".

To get this Y height value, the component calls `getBoundingClientRect()` on its wrapping div element, inside the function `getStyle()`.

In the previous implementation, `getStyle()` was called inside the render method. This meant that in some cases (namely the one for which this bug was reported), `getStyle()` was called *before* all the `<DropdownContent>`'s children had rendered.

This meant that it would give an incorrect Y height value, in the case above the height was given as `2px` rather than `300px`. The `getStyle()` function therefore said, "the Y height is only two pixels, I can easily fit that on the screen!".

The component would did not fix itself when the `<DropdownContent>`'s children eventually rendered, because it was a change in its children, and wouldn't trigger a re-render.
#### Fix
The fix was to move the call to `getStyle()` out of the render method, and into `componentDidUpdate`. Calling the function ***after*** the render instead of ***during*** the render gives enough time for `<DropdownContent>`'s children to be rendered, ensuring the correct Y height value, before calculating where the dropdown should be positioned.

### 🔗 Links 
### 📹 GIF (optional)
http://g.recordit.co/JxKdW1aoxc.gif
### 🚪 Start Points
https://github.com/gathercontent/gather-ui/pull/640/files#diff-e639efd056e0519f9e78c73125c3463a
### 👫 Related PRs (optional)

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook

### DIFF
--- a/lib/Dropdown/DropdownContent.js
+++ b/lib/Dropdown/DropdownContent.js
@@ -1,4 +1,5 @@
 import React, { Component } from 'react';
+import { isEqual } from 'lodash';
 import cx from 'classnames';
 import PropTypes from 'prop-types';
 import { GATHER_UI_DROPDOWN } from './consts';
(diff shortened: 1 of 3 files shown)


altaudio:
[[12] Due Date Dropdown Positions Off Screen](https://trello.com/c/CDoXjoIV/751-12-due-date-dropdown-positions-off-screen)

PR #642, opened 2020-01-17
Hierarchy row styles

DavidHollins6:
### 💬 Description
Each row in the hierarchy should have it's margin shared between top and bottom, instead of just all at the bottom. This makes calculating the drop area easily, as the row is now symmetrical!

6px might seem like a random number, but it is to fill out the whole row in the hierarchy.

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook

### DIFF
--- a/lib/FolderRow/styles.scss
+++ b/lib/FolderRow/styles.scss
@@ -10,6 +10,7 @@ $folder-row-spacing: $layout-spacing-base/2 !default;
 
 .folder-row {
   overflow: hidden;
+  width: 100%;
 }
 
 .folder-row__inner,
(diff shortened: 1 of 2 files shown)


DavidHollins6:
[Fix drop area for hierarchy rows](https://trello.com/c/wjIZtsa4/704-fix-drop-area-for-hierarchy-rows)

lib/FolderRow/styles.scss
```diff
@@ -23,7 +24,7 @@ $folder-row-spacing: $layout-spacing-base/2 !default;
   padding: $folder-row-spacing;
   background: $neutral-lightest;
   position: relative;
-  margin: 0 0 $layout-spacing-base/2;
+  margin: 6px 0 6px;
```
kyleharper: 🔴 can we move this to a var and then use it for the margin in item row? When we come to update it in the future it'll always catch us out when we update one and not the other.

DavidHollins6: Love that idea

PR #649, opened 2020-01-31
Refactoring SelectionBar

DavidHollins6:
### 💬 Description
`SelectionBar` was a bit static as to how it looked, there wasn't really any wiggle room around what can be displayed on it. This refactor fixes that!

It now uses a group of compound components, which are explained here:

- `SelectionBar` : The overall wrapping component.
- `SelectionBar.Information`: Holds the information that can be placed at the left hand side of the bar. E.g. Item Counts, Select All button
- `SelectionBar.Actions`: Wrapper for each `SelectionBar.Action` component.
- `SelectionBar.Cancel`: A cancel button that will be displayed in `SelectionBar` if a callback prop has been provided.
- `SelectionBar.Counter`: That nice little pill + text that will display the count of something.

The existing enzyme tests have been removed and been replaced with a couple of stronger integration tests.
### 🚪 Start Points
The best place to start would be `SelectionBar`, and branch away from there.

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook


lib/SelectionBar/index.js
```diff
@@ -1,80 +1,47 @@
...
-  autoHide: PropTypes.bool
+  autoHide: PropTypes.bool,
+  onCancel: PropTypes.func
 };
 
 SelectionBar.defaultProps = {
   fixed: false,
-  autoHide: false
+  autoHide: false,
+  onCancel: undefined
```
altaudio: 💚What do you think to using null here instead of undefined? I think it makes more sense as null is an assigned value, whereas undefined suggests that the variable hasn't been defined, which it actually has.

PR #657, opened 2020-02-07
Reduce font-weight of item names

p34ks:
Item names in the content hub are too heavy since updating the app font-library with the correct weights. Reducing to 400 brings them in line with the designs.

### 💬 Description
The tiniest PR in the history of PRs.
### 🔗 Links
### 📹 GIF (optional)
### 🚪 Start Points
### 👫 Related PRs (optional)

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook


lib/CollectionsTable/styles.scss
```diff
@@ -133,6 +133,7 @@ $collections-table-cell-min-height: 40px !default;
 .collections-table__row-title {
   margin: 0;
   font-size: $typo-size-base;
+  font-weight: 400;
```
kyleharper: `$typo-weight-roman` is the var we use to specify a weight of 400. You can see these settings in `styles/settings/_typography.scss`.

PR #662, opened 2020-02-14
Update/icon title prop

altaudio:
### 💬 Description
Adds `title` as a prop to `<Icon />`.

When passed this prop, a title element will be embedded within the SVG element, eg:
```
<svg>
  <title>Filter by assignee</title>
  <path d="M8 0C3.6 0 0 3.6 0 8C0 12.4 " fill="#8599AD"/>
</svg>
```
this can then be used in our tests to select icons with `getByTitle` AND is read out by screen readers, so icons that we're previously a mystery can now have meaning!

If an icon triggers an action, it is recommended to use a description of the action, rather than of the image, eg.

❌`title="Calendar icon"`
✅`title="Set due dates"`

> There's a bit of crappy DOM manipulation here to get the title in, as we couldn't pass children to the SVG with the [SVG Library we are using](https://github.com/airbnb/babel-plugin-inline-react-svg) :(

#### ✅ Tests 
Removes some tests as they were failing, and weren't providing any value.
### 🔗 Links
### 📹 GIF (optional)
### 🚪 Start Points
### 👫 Related PRs (optional)

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook


lib/Icon/index.js
```diff
@@ -223,7 +234,9 @@ const Icon = props => {
         <span className="icon__notification">{props.notification}</span>
       )}
       <span className="icon__body">
-        <span className="icon__image">{icon}</span>
+        <span ref={svgParentRef} className="icon__image">
+          <IconComponent />
```
AmeeMorris: 🚩cant you just chuck a title prop here? `<IconComponent title={props.title || props.name} />`

kyleharper: 🚩furthermore, can't we add the title into the SVG directly, manually? I feel like this is doing too much to alter the composition of an SVG. We could even just add the title to the surrounding span tags and not need a ref and all the DOM alterations in this PR.

altaudio: > 🚩cant you just chuck a title prop here? `<IconComponent title={props.title || props.name} />`

Good question! Passing the title as the prop here does pass it to the child svg:
```<svg title="Edit item">...</svg>```
and this works for testing, _but_ the attribute is not read out by screen readers, as the `<title>Edit item</title>` element is.

Unfortunately, the [library we use for creating our SVG components](https://github.com/airbnb/babel-plugin-inline-react-svg) does not allow us to create nodes in the child SVG, thus the nasty DOM manipulation in this PR. We could change the library, but that is a much bigger change.

altaudio: > 🚩furthermore, can't we add the title into the SVG directly, manually? I feel like this is doing too much to alter the composition of an SVG. We could even just add the title to the surrounding span tags and not need a ref and all the DOM alterations in this PR.

Again, the `title` attribute can be added to SVGs and span tags, but does not provide any accessibility value, so in this case we would be as well to use a test id.

altaudio: @AmeeMorris @kyleharper A couple of references that got me to this point 👍 

https://css-tricks.com/accessible-svgs/

> HTML title attributes are often perceived as an accessibility (and SEO) bonus, but the opposite is true. For screen reader users the content included inside of the title attribute is typically unnecessary, redundant, and possibly not even used.
https://a11yproject.com/posts/title-attributes/

> A few weeks ago I spent some time using screenreaders like a blind person would. ... . One of the main things I noticed is that the title attribute isn’t read aloud, AT ALL.
https://silktide.com/blog/2013/i-thought-title-text-improved-accessibility-i-was-wrong

AmeeMorris: @altaudio I see, well thats unfortunate! Do we also need the `aria-labelledby` attribute for the strange chrome/ff bug mentioned on this https://css-tricks.com/accessible-svgs/ ?

altaudio: > @altaudio I see, well thats unfortunate! Do we also need the `aria-labelledby` attribute for the strange chrome/ff bug mentioned on this https://css-tricks.com/accessible-svgs/ ?

Good catch - added to the code 👍

PR #665, opened 2020-02-19
Loading overlay text

kyleharper:
The PR adds the ability to add text to the loading overlay component. This will be really handy when we need to show loading percentages for long loading processes.

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook


lib/LoadingOverlay/styles.scss
```diff
@@ -51,6 +51,15 @@ $loading-line-width: 32px;
   }
 }
 
+.loading-overlay__content {
```
altaudio: 💚 I think we have helper classes for these (except `text-align: center;`? Should be easy to add).

Would be nice to get rid of the `p:last-child` selector, so we wouldn't need to change the CSS if we add another `p` tag

lib/LoadingOverlay/index.js
```diff
@@ -1,24 +1,40 @@
...
+      </div>
+    </div>
+  );
 };
 
 LoadingOverlay.propTypes = {
   fixed: PropTypes.bool,
-  hideSVG: PropTypes.bool
+  hideSVG: PropTypes.bool,
+  percentageText: PropTypes.string
```
altaudio: 💚💚💚 What do you think to passing `percentageLoaded` as a number rather than a string, and doing the string conversion/adding of `%` in this component?

Might save a little time having to do the conversion and postfix the % every time we use this component

kyleharper:
Epic feedback @altaudio I've actioned it all 👍

PR #671, opened 2020-03-05
Update/Shareables Upgrade Modal Changes

altaudio:
### 💬 Description
Here are a few changes that were spawned from the adding the Shareable Upgrade Modal:
> https://github.com/gathercontent/app/pull/2650

- Adds `defaultActiveColor` prop to Icon. By default all SVGs with the class `.icon` are set to blue if they have a parent with the class `.is-active`. Setting this new prop to false overrides this behaviour.
- Updates the `h-border-default` helper to accept a `-clear` postfix, meaning it will clear borders.
- Adds `!important` to the "Margin & Padding" SCSS helper values. This was to override some values set in React Bootstrap.
  - I wasn't so sure about this, and am happy to be challenged. I know using `!important` is a bit of a smell, as it makes it impossible to over-write. But I think it is ok for helpers to be impossible to overwrite, as they are always applied at a component level, not globally.
### 🔗 Links
### 📹 GIF (optional)
### 🚪 Start Points
### 👫 Related PRs (optional)
https://github.com/gathercontent/app/pull/2650

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook

### DIFF
--- a/styles/helpers/_layout.scss
+++ b/styles/helpers/_layout.scss
@@ -7,27 +7,27 @@ $subProps: ('', 'left', 'right', 'top', 'bottom');
 
 @include generate-helper($props, $subProps) using ($selector, $property, $style) {
   #{$selector}-double {
-    #{$property}: $layout-spacing-base * 2;
+    #{$property}: $layout-spacing-base * 2 !important;
   }
 
   #{$selector} {
-    #{$property}: $layout-spacing-base;
+    #{$property}: $layout-spacing-base !important;
   }
 
   #{$selector}-half {
-    #{$property}: $layout-spacing-base/2;
+    #{$property}: $layout-spacing-base/2 !important;
   }
 
   #{$selector}-quarter {
-    #{$property}: $layout-spacing-base/4;
+    #{$property}: $layout-spacing-base/4 !important;
   }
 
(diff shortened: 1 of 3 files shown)


altaudio:
[Shareable Links: Add Modal for upgrading](https://trello.com/c/c5iQ5YYB/1100-shareable-links-add-modal-for-upgrading)

styles/helpers/_layout.scss
```diff
@@ -7,27 +7,27 @@ $subProps: ('', 'left', 'right', 'top', 'bottom');
 
 @include generate-helper($props, $subProps) using ($selector, $property, $style) {
   #{$selector}-double {
-    #{$property}: $layout-spacing-base * 2;
+    #{$property}: $layout-spacing-base * 2 !important;
```
AmeeMorris: This makes me sad ☹️ but I've had troubles with these getting overridden too

altaudio: I thought this at first, but I think it will be ok!

Setting `!important` on something like divs globally would cause an issue, but as this class is added at a component level, it can always be removed!

PR #672, opened 2020-03-06
Update/Add justify-content helper class

altaudio:
### 💬 Description
Add `justify-content` helper class.

Same idea as `align-items` helper class.

Also removes the old `space-between` helper class. Usage instances have been removed here and in app (as you will see in the required PR 👇)

Required for https://github.com/gathercontent/app/pull/2655
### 🔗 Links
### 📹 GIF (optional)
### 🚪 Start Points
### 👫 Related PRs (optional)

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook

### DIFF
--- a/lib/CheckToggle/index.js
+++ b/lib/CheckToggle/index.js
@@ -88,6 +88,7 @@ class CheckToggle extends Component {
         {labelLeft && <p className={wrapperLabelClasses}>{labelLeft}</p>}
         <span className="toggle-wrapper__inner">
           <input
+            data-testid={id}
             onChange={this.onClickHandler}
             checked={checked}
             type="checkbox"
--- a/lib/CheckToggle/index.js
+++ b/lib/CheckToggle/index.js
@@ -70,7 +70,7 @@ class CheckToggle extends Component {
     const wrapperClasses = cx(`toggle-wrapper ${className}`, {
       'size-small': displaySmall,
       'is-checked': displayChecked && checked,
-      'space-between': spaceBetween,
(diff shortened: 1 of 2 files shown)


lib/CheckToggle/index.js
```diff
@@ -88,6 +88,7 @@ class CheckToggle extends Component {
         {labelLeft && <p className={wrapperLabelClasses}>{labelLeft}</p>}
         <span className="toggle-wrapper__inner">
           <input
+            data-testid={id}
```
AmeeMorris: Do we need this? Cant we just grab this by the label text?

altaudio:
[Add CTA to the bottom of the Shareable menu](https://trello.com/c/iYXTeWXb/1106-add-cta-to-the-bottom-of-the-shareable-menu)

lib/CheckToggle/index.js
altaudio: Good point, we should be able to!

altaudio: > Do we need this? Cant we just grab this by the label text?

To follow up on this, the `CheckToggle` component basically contains debt that means we cannot target it `byLabelText`.

The current `CheckToggle` component uses a `<label>` element as the actual switch (and hides the native input), and uses `<p>` tags for the labels.

I took a look at the refactor and I think it is going to go our the scope of this PR. One for the future though under lesser time pressures 👍

PR #673, opened 2020-03-10
add a new DismissablePrompt component

AmeeMorris:
### 💬 Description
This is just adding a simple component to allow us to show dismissible prompts around the app. Nothing clever going on, it's mainly so we don't have styling clogging the app!

### ✅ Checklist
- [x] Tests written
- [ ] Browser tested
- [x] Added to documentation
- [x] Added to storybook

### DIFF
--- a/lib/DismissiblePrompt/index.js
+++ b/lib/DismissiblePrompt/index.js
@@ -1,0 +1,29 @@
+import React from 'react';
+import { node, string, func } from 'prop-types';
+import Icon from '../Icon';
+import Button from '../Button';
+
+const DismissiblePrompt = ({ children, onDismiss, className, ...rest }) => (
+  <div {...rest} className={`dismissible-prompt ${className}`}>
+    <div className="dismissible-prompt__children">{children}</div>
+    <Button
+      types={['icon-only', 'collapse']}
+      className="dismissible-prompt__close"
+      onClick={onDismiss}
+    >
+      <Icon name="cross" />
+    </Button>
+  </div>
+);
+
+DismissiblePrompt.propTypes = {
+  children: node.isRequired,
+  className: string,
+  onDismiss: func.isRequired
+};
+
+DismissiblePrompt.defaultProps = {
+  className: ''
+};
+
+export default DismissiblePrompt;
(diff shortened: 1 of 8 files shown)


lib/DismissiblePrompt/index.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import { node, string, func } from 'prop-types';
+import Icon from '../Icon';
+import Button from '../Button';
+
+const DismissiblePrompt = ({ children, onDismiss, className, ...rest }) => (
+  <div {...rest} className={`dismissible-prompt ${className}`}>
```
DavidHollins6: Gonna get accessible-cray here. Could we add these roles + attributes to this div. It just makes screen readers understand it a bit better.

There are bigger accessibility issues are modals/dialogs, but this at least makes things better going forward!

Screenshot 2020-03-10 at 14 49 33


(I stole it from here: https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/dialog.html)

lib/DismissiblePrompt/stories/DismissiblePromptStory.js
```diff
@@ -0,0 +1,15 @@
...
+import DismissiblePrompt from '..';
+import StoryItem from '../../../stories/styleguide/StoryItem';
+
+storiesOf('Components', module).add('DismissiblePrompt', () => (
+  <div>
+    <StoryItem
+      title="DismissiblePrompt"
+      description="A small prompt that can be dissmissed"
+    >
+      <DismissiblePrompt>hello there!</DismissiblePrompt>
```
petemcfarlane: does this require an `onDismiss` prop? I can't see a default one

AmeeMorris: keen eye! i'll add in

lib/DismissiblePrompt/index.js
AmeeMorris: Just for visibility, cleared things up on slack. this guy isnt a modal, just a little box that lives along side content rather than overlaying anything. We spread the rest of the props so if you wanted to use it that way you can always add these in!

PR #679, opened 2020-03-18
Add an optional close button to Notification

AmeeMorris:
### 💬 Description
This adds the option for a close button on the Notification component. I'm kinda surprised we didn't have this in already!

### ✅ Checklist
- [x] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [x] Added to storybook


lib/Notification/index.js
```diff
@@ -28,23 +32,34 @@ const Notification = ({
...
-      {children}
+      <div className="notification__content">{children}</div>
+      {onClose && (
+        <Button
+          onClick={onClose}
+          types={['icon-only']}
+          className="notification__close"
+        >
+          <Icon name="cross" />
+        </Button>
```
DavidHollins6: 🍏 Should either of these things have a `title`? It might make testing with it in the futur easier.

AmeeMorris: Excellent point! will add

PR #681, opened 2020-03-23
Fix memory leaks with Draggables

kyleharper:
### 💬 Description
When writing tests we've noticed that our `Draggable` component can cause memory leaks. The underlining source of the issue is still an unknown but there are complications with `onDrop` and `onEndDrag` which cause us to have memory leaks in our tests.

There's a silver lining. The change to `Draggable` makes sense because when we drop A into B we should handle changes in `onDrop` via `Droppable`. This leaves `onEndDrag` to handle all those times we don't drag something into a dropable.

E.g. we should clear the drag preview when we drag `A` into some whitespace. We should also clear the preview when dropping `A` successfully.

---

I've also added a nice little additional for `Dropdown`s where we can expose the showing boolean so we can log it for usage tracking 👍 

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook

### DIFF
--- a/lib/DnD/Draggable.js
+++ b/lib/DnD/Draggable.js
@@ -14,8 +14,10 @@ function Draggable({ children, preview, item, onBeginDrag, onEndDrag }) {
       onBeginDrag(item, monitor);
     },
     end: (targetItem, monitor) => {
-      onEndDrag(targetItem, monitor);
-      setIsDragging(false);
+      if (!monitor.didDrop()) {
+        onEndDrag(targetItem, monitor);
+        setIsDragging(false);
+      }
     },
     collect: monitor => ({
       itemType: monitor.getItemType(),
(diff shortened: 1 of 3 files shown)


lib/DnD/Draggable.js
```diff
@@ -14,8 +14,10 @@ function Draggable({ children, preview, item, onBeginDrag, onEndDrag }) {
       onBeginDrag(item, monitor);
     },
     end: (targetItem, monitor) => {
-      onEndDrag(targetItem, monitor);
-      setIsDragging(false);
+      if (!monitor.didDrop()) {
+        onEndDrag(targetItem, monitor);
+        setIsDragging(false);
+      }
```
DavidHollins6: I've started implementing the app changes, and I've found myself inserting `setIsDragging(false)` in every `onDrop` in the app. This feels a bit smelly, maybe we could hook it into `Droppable` inside gather-ui instead?

PR #683, opened 2020-03-25
Add a callback for onToggle in Guidelines

kyleharper:
### 💬 Description
Seems like a sensible callback to add!

Very handy for things like tracking etc.

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook

### DIFF
--- a/lib/Guideline/index.js
+++ b/lib/Guideline/index.js
@@ -10,11 +10,13 @@ class Guideline extends Component {
       PropTypes.arrayOf(PropTypes.node),
       PropTypes.string
     ]),
-    title: PropTypes.string.isRequired
+    title: PropTypes.string.isRequired,
+    onToggle: PropTypes.func
   };
 
   static defaultProps = {
-    children: ''
+    children: '',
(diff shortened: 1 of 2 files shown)


lib/Guideline/index.js
```diff
@@ -26,9 +28,13 @@ class Guideline extends Component {
...
 
   toggleShow = () => {
-    this.setState(prevState => ({
-      showContent: !prevState.showContent
-    }));
+    this.setState(
+      prevState => ({
+        showContent: !prevState.showContent
+      }),
+      () => this.props.onToggle(this.state.showContent)
```
altaudio: 💚As `this.state.showContent` is set asynchronously, I don't think we can guarantee that `this.state.showContent` is the state that has just been set.

I'm pretty sure this is edge case, as someone would have to be clicking the toggle pretty fast for it to get out of whack. Does the `setState` callback pass the updates state as its first arg?

PR #684, opened 2020-03-25
Update/Bulk Share Test Changes

altaudio:
### 💬 Description
Couple of changes to GUI that correspond to the tests added for Bulk Share Items https://github.com/gathercontent/app/pull/2747

#### 壱 `<ProgressButton />` `onCallbackResolved` prop
This new prop is called after the `clickHandler` promise passed to progress button has resolved, and is passed the result of the promise.

Previously I had a call to close the modal _inside_ the promise `clickHandler` that was passed to a progress button.

This threw a `Cannot update state on an unmounted component` error as we unmounted the modal (and the `<ProgressButton/>` inside it), then the `<ProgressButton/>` tried to update it's `showLoader` state.

The new change means we can call `hideModal` after the `<ProgressButton/>` has finished setting state.
#### 弍 `<SelectionProvider />` `initialSelected` prop
This new prop was added so that we can set the initial selected values in tests (in the Bulk Share example, we aren't rendering the UI required to set items normally.)
### 🔗 Links
### 📹 GIF (optional)
### 🚪 Start Points
### 👫 Related PRs (optional)
https://github.com/gathercontent/app/pull/2747

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook

### DIFF
--- a/lib/ProgressButton/index.js
+++ b/lib/ProgressButton/index.js
@@ -14,7 +14,8 @@ class ProgressButton extends Component {
     disabled: false,
     autoSpinner: false,
     spinnerText: null,
-    className: ''
+    className: '',
+    onCallbackResolved: null
   };
 
   static propTypes = {
@@ -28,7 +29,8 @@ class ProgressButton extends Component {
     disabled: PropTypes.bool,
     autoSpinner: PropTypes.bool,
     spinnerText: PropTypes.string,
-    className: PropTypes.string
+    className: PropTypes.string,
+    onCallbackResolved: PropTypes.func
   };
 
   constructor(props) {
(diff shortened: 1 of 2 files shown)


altaudio:
[Bulk Share: Tests!](https://trello.com/c/35mj3C6X/100-bulk-share-tests)

lib/ProgressButton/index.js
```diff
@@ -28,7 +29,8 @@ class ProgressButton extends Component {
     disabled: PropTypes.bool,
     autoSpinner: PropTypes.bool,
     spinnerText: PropTypes.string,
-    className: PropTypes.string
+    className: PropTypes.string,
+    onCallbackResolved: PropTypes.func
```
DavidHollins6: Amazing, I was going at add the exact same thing today!

PR #686, opened 2020-03-26
onCompletion callback for ConfirmationDropdown

DavidHollins6:
### 💬 Description
This PR adds a callback prop to `ConfirmationDropdown`. We need this because there needs to be some logic executed when the promise has completely finished and the dropdown is mostly closed.

There was previously a memory leak here, because clearing the selected objects could happen to early, and cause this to unmount before it had finished doing everything. This now ensures that clearing the selected objects happens last.

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook

### DIFF
--- a/lib/ConfirmationDropdown/ConfirmationDropdownContent.js
+++ b/lib/ConfirmationDropdown/ConfirmationDropdownContent.js
@@ -50,13 +50,13 @@ const ConfirmationDropdownContent = (
           <Button
             types={confirmButtonTypes}
             clickHandler={() =>
-              onConfirm()
-                .then(() => {
-                  if (hideOnCompletion) {
-                    setShowContent(false);
-                  }
-                })
-                .then(onCompletion)
+              onConfirm().then(() => {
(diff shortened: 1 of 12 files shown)


DavidHollins6:
[Fix memory leaks in selection bars](https://trello.com/c/FuBdhtDm/97-fix-memory-leaks-in-selection-bars)

lib/ConfirmationDropdown/ConfirmationDropdownContent.js
```diff
@@ -49,11 +50,13 @@ const ConfirmationDropdownContent = (
           <Button
             types={confirmButtonTypes}
             clickHandler={() =>
-              onConfirm().then(() => {
-                if (hideOnCompletion) {
-                  setShowContent(false);
-                }
-              })
+              onConfirm()
+                .then(() => {
```
kyleharper: 🍏 why chain twice?

PR #690, opened 2020-04-03
Feature/tailwind notification

AmeeMorris:
### 💬 Description
This adds some exciting GCDS-y tailwind-y notifications in GUI.

We now have two types of Notification:
- bar (which is a restyled version of our old Notification component) 
- inline. 

As suggested we've now split out all our notification "levels" to their own separate components, but we've kept the ability to use one component and pass a level prop. 

A chunk of the changes here are just changing the path/name for things that relied on old notifications

### 🔗 Links
https://www.figma.com/file/66iSurBAx32DWrJymo9NKL/GCDS-%E2%80%93-Web-UI?node-id=500%3A5389
### 🚪 Start Points
`lib/Notification/bar/index.js`
`lib/Notification/inline/index.js`

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook


tailwind.config.js
```diff
@@ -12,9 +12,13 @@ module.exports = {
     spacing: {
       quarter: '5px',
       half: '10px',
+      threeQuarter: '15px',
       base: '20px',
       double: '40px',
     },
+    borderRadius: {
+      default: '6px',
+    },
```
DavidHollins6, line 21: Are shadows included in this?

lib/Notification/styles.scss
```diff
@@ -1,114 +1,13 @@
-/* ==========================================================================
```
DavidHollins6, line 1: This is so good to see.

tailwind.config.js
AmeeMorris: Ah I missed this config, will add now since our inline boys use a shadow

lib/Notification/styles.scss
```diff
@@ -1,114 +1,13 @@
...
-	font-size: $notification-font-size;
-	padding: $notification-padding;
-	border: $notification-border-width $notification-border-style transparent;
-	border-radius: $border-radius-base;
-
-	&__content {
-		width: 100%;
-	}
-
+	text-shadow: none;
```
altaudio, line 5: 💚 We should look to move this styling to whatever text element is rendered as the child of notification, rather than the notification element handling it as the parent.

Total future change, as I think this would cause a bucket load of work at the moment!

lib/Notification/bar/NotificationBarBase.js
```diff
@@ -0,0 +1,53 @@
...
+  center,
+  onClose,
+  ...rest
+}) => {
+  const classes = cx({
+    'cursor-pointer': clickHandler,
+    'text-center': center,
+    flex: onClose,
+    'justify-end': center && onClose,
+    'notification--has-close': onClose
```
altaudio, line 26: 🚩 Do we need this class anymore? I _think_ that it was removed from the styles, and line 24 does what it used to.

lib/Notification/styles.scss
altaudio: 🚩 Can we not add this with tailwind?

lib/Notification/styles.scss
```diff
@@ -1,114 +1,13 @@
...
-	justify-content: flex-end;
-	.notification__close {
-		margin-left: $layout-spacing-base/2;
+		@apply text-current;
 	}
-	.notification__content {
-		margin: auto;
+	.icon path,
+	.notification__close .icon path {
+		@apply fill-current;
```
altaudio, line 11: 🚩Can we apply these with tailwind to Icon instead of here?

lib/Notification/inline/NotificationInlineBase.js
```diff
@@ -0,0 +1,42 @@
...
+      role="alert"
+      className={`notification text-base font-semi-bold py-half px-threeQuarter border-2 border-solid inline-flex text-neutral-20 bg-white rounded items-center shadow-large border-${borderColor} ${className}`}
+      {...rest}
+    >
+      <Icon
+        name={iconName}
+        className={`text-${borderColor} mr-half`}
+        defaultActiveColor={false}
+      />
+      <div className="notification__content w-full">{children}</div>
```
altaudio, line 29: 🚩 Do we still need `notification__content`?

lib/Notification/bar/NotificationBarBase.js
```diff
@@ -0,0 +1,53 @@
...
+  });
+
+  return (
+    <Alert
+      onClick={clickHandler}
+      className={`notification text-base font-semi-bold p-half border-0 rounded-none ${classes} ${className}`}
+      {...rest}
+    >
+      <div className="notification__content w-full">{children}</div>
```
altaudio, line 35: 🚩 Do we still need `notification__content`?

lib/Notification/bar/barTypes.js
```diff
@@ -0,0 +1,16 @@
+import { func, node, string, bool } from 'prop-types';
+
+export const defaults = {
```
altaudio, line 3: 💚 One other method could be to add these default props and types to `NotificationBarBase`, and then do something like the following in the sub components:
```
// NotificationBarWarning.js
....

NotificationBarWarning.propTypes = NotificationBarBase.propTypes;
```

lib/Notification/bar/index.js
```diff
@@ -0,0 +1,39 @@
...
+import { defaults, types } from './barTypes';
+import NotificationBarDanger from './NotificationBarDanger';
+import NotificationBarInformation from './NotificationBarInformation';
+import NotificationBarPromo from './NotificationBarPromo';
+import NotificationBarWarning from './NotificationBarWarning';
+
+const NotificationBar = ({ level, children, ...rest }) => {
+  const getNotificationBar = () => {
+    switch (level) {
+      case 'warning':
```
altaudio, line 12: 💚 👍 Nice!

You could export these level strings as part of the component, and then use them in the app
```
// bar/index.js
...
NotificationBar.levels = {
  warning: 'warning',
  ...etc
}

// App.js

<NotificationBar level={NotificationBar.levels.warning} />
```

lib/Notification/bar/index.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import { string } from 'prop-types';
+import { defaults, types } from './barTypes';
+import NotificationBarDanger from './NotificationBarDanger';
+import NotificationBarInformation from './NotificationBarInformation';
+import NotificationBarPromo from './NotificationBarPromo';
+import NotificationBarWarning from './NotificationBarWarning';
+
+const NotificationBar = ({ level, children, ...rest }) => {
```
altaudio, line 9: 👍 Nice component!

tailwind.config.js
```diff
@@ -12,9 +12,19 @@ module.exports = {
...
+    boxShadow: {
+      small: '0px 2px 4px rgba(0, 0, 0, 0.06)',
+      default: '0px 3px 6px rgba(0, 0, 0, 0.06)',
+      large: '0px 8px 16px rgba(0, 0, 0, 0.1)',
+      xlarge: '0px 15px 30px rgba(0, 0, 0, 0.2)',
+    },
```
altaudio, line 27: 🚩 Worth pinging Peaks on these, he updated the box shadows to work better on dark backgrounds (unless these are from GCDS)

lib/Notification/styles.scss
AmeeMorris: no, its a limitation :(

tailwind.config.js
AmeeMorris: These are from GCDS!

lib/Notification/bar/NotificationBarBase.js
AmeeMorris: I think more descriptive classes won't really hurt being in there

stories/components/Icons.js
```diff
@@ -81,6 +81,8 @@ const allIcons = additionalProps => (
     <Icon name="userEdit" {...additionalProps} />
     <Icon name="key" {...additionalProps} />
     <Icon name="warningTriangle" {...additionalProps} />
+    <Icon name="warningOctogon" {...additionalProps} />
```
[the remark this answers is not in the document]
AmeeMorris: Yeah I like this idea!

lib/Notification/styles.scss
AmeeMorris: We may not need the .notification__close selector anymore, but i'd still like to do this for all icons that may be used as children

PR #692, opened 2020-04-08
Feature/linkify field instructions

AmeeMorris:
### 💬 Description
As we do with our lovely comments, we've now added Linkify to our field instructions so any url that gets pasted gets converted to a lovely link. wow!


AmeeMorris:
[Add hyperlinks to guidelines and instructions.](https://trello.com/c/FWRVMTtW/153-add-hyperlinks-to-guidelines-and-instructions)

PR #695, opened 2020-04-09
Feature/wordpress modal story

petemcfarlane:
### 💬 Description
Built a non-trivial modal component in the gather-ui library just as a story, so I can see how it will be built in the app. 

If you think this should just belong in the app then that's fine, but there's a few minor commits that I made to gather-ui - like adding a new icon svg, and some minor css, that will still need to be merged in to gather-ui - feedback most welcome 😄 

Screenshot 2020-04-09 at 21 02 55

https://www.figma.com/file/pvcWU0mMV2wttjBNkedKDf/Wordpress-integration?node-id=2%3A534

### ✅ Checklist
- [ ] Tests written
- [x] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook


stories/webapp/accountdashboard/WordPressIntegrationStory.js
```diff
@@ -0,0 +1,105 @@
...
+});
+
+stories.add('WordPressIntegrationStory', () => {
+  const modalSubHeader = (
+    <Fragment>
+      <Icon name="integration" className="m-quarter mr-half" />
+      Wordpress v5.4
+      <span className="text-xs text-green-primary bg-green-90 px-quarter py-tenth inline-flex rounded-sm font-semi-bold mx-quarter">
+        CONNECTED
+      </span>
```
altaudio, line 20: 🚩 We should have a component called Pill that basically does this styling for you

stories/webapp/accountdashboard/WordPressIntegrationStory.js
```diff
@@ -0,0 +1,105 @@
...
+                className="w-1/4 font-semi-bold ml-half"
+              >
+                Admin URL
+              </label>
+              <div className="flex-grow">
+                <InputWithButton
+                  inputId="preview-admin-url"
+                  value="https://cms-wordpress.gatherpreview.com/acme/wp-admin"
+                  onClick={action('copy admin url')}
+                  // onClick={copyElementValueToClipboard}
```
altaudio, line 63: 🚩 Kill the commented out code

stories/webapp/accountdashboard/WordPressIntegrationStory.js
```diff
@@ -0,0 +1,105 @@
...
+  );
+
+  return (
+    <ModalTrigger>
+      <Modal.Container>
+        <Modal.Header text={modalSubHeader}>WordPress demo</Modal.Header>
+        <Modal.Body className="has-columns">
+          <div className="modal__column bg-neutral-98">
+            <div className="border border-solid border-neutral-90 rounded bg-white mb-half p-half flex items-center justify-between">
+              <label // eslint-disable-line jsx-a11y/label-has-associated-control
```
altaudio, line 34: 💚 How annoying you had to add this! I guess that's the problem with eslint thought, it looks at one file for issues, and doesn't know that `inputId` is being passed down to an `<input>` element further down the chain!

I wonder if changing the name of the `<InputWithButton />` prop from `inputId` => `id` would fix this? Up to you if you wan't to follow this line of enquiry, as it is out of scope of this PR, and would lead to follow up work in the App!

You can always disable a rule for the whole file with `/* eslint-disable jsx-a11y/label-has-associated-control */`, but again, optional as this currently works 👍

stories/webapp/accountdashboard/WordPressIntegrationStory.js
```diff
@@ -0,0 +1,105 @@
...
+
+const stories = storiesOf('Web app', module);
+
+const ModalTrigger = withModalTrigger({
+  children: <Icon name="integration" />,
+  types: ['icon-only']
+});
+
+stories.add('WordPressIntegrationStory', () => {
+  const modalSubHeader = (
```
altaudio, line 14: 💚 I would stick this in a separate component when you get in the App, to keep this file readable and the component nice and composed.

In here I think this works fine though 🙏

stories/webapp/accountdashboard/WordPressIntegrationStory.js
```diff
@@ -0,0 +1,105 @@
...
+        Expires on 30/04/2020
+      </span>
+    </Fragment>
+  );
+
+  return (
+    <ModalTrigger>
+      <Modal.Container>
+        <Modal.Header text={modalSubHeader}>WordPress demo</Modal.Header>
+        <Modal.Body className="has-columns">
```
altaudio, line 31: 💚 I think when this goes into the App, it would be nice to have the `<Modal.Body/>` and it's contents in a separate component. 

This file is a little long as is (but is ok as it's a story). Once you shift to the App and add state in etc. it's going to get lengthy.

tailwind.config.js
```diff
@@ -18,6 +19,7 @@ module.exports = {
     },
     borderRadius: {
       default: '6px',
+      sm: '4px',
```
altaudio, line 22: 🚩 You'll need to check with Peaks whether this can go in - we should be trying to keep aligned with the new GCDS library with the styles in here.

That said, I don't think you will need this style anymore once you use the Pill component (comment above)

styles/overrides/_modal.scss
```diff
@@ -32,8 +32,7 @@
   .close {
     position: absolute;
     right: $layout-spacing-base;
-    top: 50%;
-    transform: translateY(-50%);
+    top: $layout-spacing-base;
```
altaudio, line 35: 💚 It's worth checking that this doesn't bork any of the other current modal instances in the App

tailwind.config.js
kyleharper: @p34ks

stories/webapp/accountdashboard/WordPressIntegrationStory.js
```diff
@@ -0,0 +1,105 @@
...
+                  value="https://cms-wordpress.gatherpreview.com/acme"
+                  onClick={action('copy preview url')}
+                  // onClick={copyElementValueToClipboard}
+                  buttonText="Copy link"
+                  buttonTextAfterClick="Copied!"
+                />
+              </div>
+            </div>
+            <div className="border border-solid border-neutral-90 rounded bg-white mb-half p-half flex items-center justify-between">
+              <label // eslint-disable-line jsx-a11y/label-has-associated-control
```
kyleharper, line 52: 🔴 why do we need ` jsx-a11y/label-has-associated-control` here?

stories/webapp/accountdashboard/WordPressIntegrationStory.js
```diff
@@ -0,0 +1,105 @@
...
+            </div>
+            <div className="border border-solid border-neutral-90 rounded bg-white p-half flex items-center justify-between">
+              <label // eslint-disable-line jsx-a11y/label-has-associated-control
+                htmlFor="password"
+                className="w-1/4 font-semi-bold ml-half"
+              >
+                Password
+              </label>
+              <div className="flex-grow">
+                <InputWithButton
```
kyleharper, line 77: 🔴 If this is to be a password, then we should use a password type? Do we spread props over the input (🙏) and if so can we pass the password type in here?

stories/webapp/accountdashboard/WordPressIntegrationStory.js
```diff
@@ -0,0 +1,105 @@
...
+                  buttonText="Copy link"
+                  buttonTextAfterClick="Copied!"
+                />
+              </div>
+            </div>
+          </div>
+        </Modal.Body>
+        <Modal.Footer>
+          <div className="flex-row-reverse p-quarter">
+            <Button types={['link']} clickHandler={action('clickHandler')}>
```
kyleharper, line 91: 🔴 `Button` should use `onClick`. `clickHandler` is debt and will be phased out 👍

stories/webapp/accountdashboard/WordPressIntegrationStory.js
petemcfarlane: Yes - unfortunately we _do_ have a component called `Pill` and it does do styling, but I needed to add a bunch of styles to override the styling. There's also a class `.pill` in the app - which we use for the purple BETA labels. I didn't want to clash with that, but if you think it's worth it I'll split this into a new PR to update our gather-ui Pill component - offering size/colour variants.

stories/webapp/accountdashboard/WordPressIntegrationStory.js
petemcfarlane: because the corresponding id (that the label is for) is on an `InputWithButton`, eslint doesn't understand that the inputId prop is translated down the chain to an `id` on a regular `input`  - related to this comment: https://github.com/gathercontent/gather-ui/pull/695#discussion_r407377805

stories/webapp/accountdashboard/WordPressIntegrationStory.js
petemcfarlane: Yeah - we had this debate in 🐢. The point is it's not really a password - it's more of an access code, we're trying to make the site accessible, we want to make it easy to copy and paste/view this value, which is why we left it as visible text. The user is hopefully only going to copy the value and paste it in an email or chat message anyway

tailwind.config.js
petemcfarlane: @p34ks and 🐢 chatted about this in standup - we've decided to drop the pill from the modal altogether, just to make less work for now 😄

petemcfarlane:
Thanks for reviews @kyleharper @altaudio - After the valid feedback, I started to extract a pill, but found we already have a `.pillbox` and `Pill` component that I don't want to interfere with right now - we decided to just kill the new pill from the design altogether, to make less work for now!

I've also decided that, whilst it was useful to have the storybook playground for designing/building and testing in GatherUi, there's little point in it living here instead of the App, because I had to override several styles in the App anyway so they were both quite different. I've dropped a few commits in here since, so we're just adding the icon and moving the cross on the modal form. I've viewed all the modals on the storybook and they all seem fine 👍 If anyone has a chance to re-review that'd be great! 🙏

altaudio:
@petemcfarlane 
>  I've viewed all the modals on the storybook and they all seem fine 👍

It will be worth linking GUI up to the app locally and checking the modal instances in the App as well, as they may be affected.

petemcfarlane:
Thanks @altaudio I've tested as many modals in the App as I can find and they all still work and look 👍

PR #696, opened 2020-04-16
Update/Remove Fixed Elements Item Editor Changes

altaudio:
### 💬 Description
A few changes required for removing fixed elements in the item editor:
- Adds `fullHeight` prop to BlankSlate, so that it can expand to the remaining space in it's parent
- Added `tailwind.config.js` to be watched by nodemon in the `build:watch` command, so changes to the config will rebuild the local GUI
- Removed no longer used `h-fit-viewport-vertically` helper class
- Added triple spacing to tailwind config
- Added `!important` rule to tailwind. I had a class that was being overruled. Will this be an issue?
### 🔗 Links
### 📹 GIF (optional)
### 🚪 Start Points
### 👫 Related PRs (optional)
https://github.com/gathercontent/app/pull/2824

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook


tailwind.config.js
```diff
@@ -1,4 +1,5 @@
 module.exports = {
+  important: true,
```
DavidHollins6, line 2: 💚 What is overriding the tailwind styles? Is it worth the time/effort to refactor them out of app to avoid using the important?

altaudio:
[Make the editor fields a scroll container](https://trello.com/c/fhSHMe0s/208-make-the-editor-fields-a-scroll-container)

package.json
```diff
@@ -56,7 +56,7 @@
     "lint": "eslint -c .eslintrc.json 'lib/**/*.js' 'tests/**/*.js'",
     "lint:fix": "eslint -c .eslintrc.json --fix 'lib/**/*.js' 'tests/**/*.js' 'stories/webapp/**/*.js'",
     "lint:check": "eslint --print-config . | eslint-config-prettier-check",
-    "copy:styles:watch": "nodemon --exec \"yarn copy-styles\" --watch lib --watch styles -e scss",
+    "copy:styles:watch": "nodemon --exec \"yarn copy-styles\" --watch lib --watch styles -e scss --watch tailwind.config.js",
```
AmeeMorris, line 59: nice!

tailwind.config.js
```diff
@@ -15,6 +16,7 @@ module.exports = {
       threeQuarter: '15px',
       base: '20px',
       double: '40px',
+      triple: '60px',
```
AmeeMorris, line 19: I'm not sure if im into these names the more that we add to them 😅not something for you to change, but something we should chat about in a FE catchup or something!

tailwind.config.js
AmeeMorris: Agree. not massively against adding this in if its saving headaches, I was just trying to avoid for as long as possible.

I had it turned on originally when I was testing out tailwind and I didn't see any issues from it from a cursory look though!

altaudio: Yeah I will give this a crack 👍

tailwind.config.js
altaudio: Agree! I wonder if these are valid, or if we should go for the tailwind method of basing it on `rem`.

tailwind.config.js
altaudio: Refactored and removed!

tailwind.config.js
altaudio: rem spacing added!

PR #699, opened 2020-04-22
How would we structure conversations components from scratch in a perfect world

DavidHollins6:
### 💬 Description
Hi All!

**I recommend reading the notion document before looking at the code!**

Here is a proposal for the layout of conversation components. This PR is huge because it's splitting up Conversations into lots of smaller components, so barely any of this code is new, it's just in a different place.

The actual proposal can be found on notion, in the spikes for the General Comments release
https://www.notion.so/gathercontent/17-04-2020-How-would-we-implement-conversations-from-scratch-in-a-perfect-world-6245dab87f6242f9bc6221e760665259

In terms of code here, I haven't touched any of the old components, these are all new files and are separate to what's already there. I have made a new stories folder that is supposed to exactly match the stories that are already there, that is to make sure I can reference the old stories and ensure I haven't broken anything.

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook


DavidHollins6:
[How would we structure conversations components in a perfect world](https://trello.com/c/xh0GmGKz/233-how-would-we-structure-conversations-components-in-a-perfect-world)

lib/Comment/CommentBody.js
```diff
@@ -0,0 +1,70 @@
...
+  onEditCancel,
+  onEditSubmit
+}) {
+  return (
+    <Fragment>
+      <Person person={comment.author} className="conversation" />
+      <div className="conversation__comment-body">
+        {!isEditing && (
+          <p className="conversation__text">
+            <Linkify>{comment.body}</Linkify>
```
AmeeMorris, line 22: 💚 we have a thing to highlight any mentions in the current version of Conversation, are you thinking of moving that over to the app side now?

PR #700, opened 2020-04-22
Update/Adds ButtonPrimary to GUI from GCDS

altaudio:
### 💬 Description
Adds the first of the new buttons to gather-ui from GCDS - please welcome, `ButtonPrimary`! 🎉

This PR creates the base button component, that implements the common styles and logic that all the buttons use.

`ButtonPrimary` is then created by extending (not literally) `ButtonBase`.

Let me know what you think about the pattern! There is also a comment in there from me I would love to get your opinions on, as I am stumped!

Check it! 👉https://share.getcloudapp.com/xQuWgqgy

> ⚠️EDIT: @AmeeMorris made the cracking suggestion to split the `PrimaryButton` so we now also have `PrimaryButtonDanger`, rather than passing a `danger` prop to `PrimaryButton`. Much neater 👌 
### 🔗 Links
https://www.figma.com/file/66iSurBAx32DWrJymo9NKL/GCDS-%E2%80%93-Web-UI?node-id=77%3A0
### 📹 GIF (optional)
### 🚪 Start Points
### 👫 Related PRs (optional)

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook


tailwind.config.js
```diff
@@ -10,13 +10,19 @@ module.exports = {
...
     boxShadow: {
       small: '0px 2px 4px rgba(0, 0, 0, 0.06)',
       default: '0px 3px 6px rgba(0, 0, 0, 0.06)',
       large: '0px 8px 16px rgba(0, 0, 0, 0.1)',
       xlarge: '0px 15px 30px rgba(0, 0, 0, 0.2)',
+      blue: '0px 2px 2px rgba(0, 110, 255, 0.5), inset 0px -2px 0px rgba(0, 0, 0, 0.1)',
+      'blue-inset': 'inset 0px 2px 0px rgba(0, 0, 0, 0.1)',
+      red: '0px 2px 2px rgba(235, 85, 71, 0.5), inset 0px -2px 0px rgba(0, 0, 0, 0.1)',
+      'red-inset': 'inset 0px 2px 0px rgba(0, 0, 0, 0.2)',
+      none: 'none'
```
altaudio, line 25: 🤔So I am a bit stumped with this one!

If you check out the designs for the buttons, they have a colored box shadow and inset on hover and active states https://www.figma.com/file/66iSurBAx32DWrJymo9NKL/GCDS-%E2%80%93-Web-UI?node-id=3485%3A1532

These new values `red`, `blue`, `red-inset` and `blue-inset` represent these.

My problem is, they are very specific to the buttons, and feel a bit off to keep them in this file. Any thoughts on this?

tailwind.config.js
```diff
@@ -91,7 +97,10 @@ module.exports = {
       }
     },
   },
-  variants: {},
+  variants: {
+    backgroundColor: ['responsive', 'hover', 'focus', 'active'],
+    boxShadow: ['responsive', 'hover', 'focus', 'active']
+  },
```
altaudio, line 103: Needed to add these in to allow `hover`, `focus`, `active` states (they are turned off by default)

altaudio:
[Add Button Primary to GCDS](https://trello.com/c/0D1Bkxnd/242-add-button-primary-to-gcds)

lib/ButtonNew/ButtonBase.js
```diff
@@ -0,0 +1,66 @@
...
+
+  if (size === md) {
+    return 'text-base py-4 px-5 m-5';
+  }
+
+  if (size === sm) {
+    return 'text-base p-3 m-3';
+  }
+
+  return 'text-sm py-1 px-3 m-3';
```
AmeeMorris, line 19: 🚩How come this isnt with the rest of our classes? I'm assuming there's a default prop for size that we can use for the default

lib/ButtonNew/ButtonPrimary.js
```diff
@@ -0,0 +1,45 @@
...
+  children,
+  className,
+  disabled,
+  connectedRight,
+  danger,
+  ...rest
+}) {
+  const classes = cx('text-white ', className, {
+    'bg-blue-primary': !danger,
+    'bg-red-primary': danger,
```
AmeeMorris, line 16: 💚Would it be better to have a separate components for BluePrimary and RedPrimary?

lib/ButtonNew/ButtonPrimary.js
```diff
@@ -0,0 +1,45 @@
...
+    'bg-neutral-90 text-neutral-primary': disabled,
```
AmeeMorris, line 17: 💚You'll have to add some more tailwind config but you can use tailwind for disabled styles like so https://tailwindcss.com/docs/pseudo-class-variants/#disabled-v1-1-0 up to you wether you go for this since having them here is fine!

lib/ButtonNew/ButtonPrimary.js
```diff
@@ -0,0 +1,45 @@
...
+    'border-blue-40': connectedRight && !disabled,
+    'border-neutral-80': connectedRight && disabled,
+    'hover:bg-blue-60 hover:shadow-blue active:bg-blue-40 active:shadow-blue-inset':
```
AmeeMorris, line 20: 💚same comment as above, and if blue & red were separate then we'd lose a couple of these conditional styles 👀

lib/ButtonNew/common.js
```diff
@@ -0,0 +1,40 @@
...
+
+export const propTypes = {
+  ...basePropTypes,
+  danger: bool
+};
+
+export const defaultProps = {
+  ...baseDefaultProps,
+  danger: false
+};
```
AmeeMorris, line 40: I like this file! 🙌

tailwind.config.js
AmeeMorris: I think it makes sense to keep them in here since they are still shadows. It'd be nice if we could just control the shadow colour, but its not possible :(

lib/ButtonNew/ButtonPrimary.js
altaudio: I like this idea. `ButtonPrimary` and `ButtonPrimaryDanger` certainly make the components easier to read 👍

lib/ButtonNew/ButtonPrimary.js
altaudio: I'm going to leave this as is for now, as I like the control we get with keeping this in JS rather than CSS.

I also believe (but haven't validated) that if we used CSS pseudo selector `disabled`, we would have to add the styles at the `ButtonBase` level, as `ButtonPrimary` etc render this component, which in turn renders the button. If so, I think this could cause the components to get a bit messier.

PR #709, opened 2020-04-24
Refactor conversation styles

DavidHollins6:
### 💬 Description
I've done goofed and ended up making a monster PR. I can't see a nice way to salvage what's happened here, so I will be commenting a lot of it to explain what has happened. Just make sure you've got your tea, you could be in for a longun.

Here's the brief overview:

Conversations now _mostly_ uses tailwind for styles. A lot of it was simple uses of the tailwind classes, there were also some really questionable parts that it just didn't seem possible with tailwind, they are inside a new comments .scss file all grouped together nicely.

There are also a couple of component structure changes, these have been added to the notion document and have been highlighted so you can see what is new! I wish I had gone and done these to the original branch, but I wasn't thinking.

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook


lib/Comment/CommentBody.js
```diff
@@ -1,51 +1,67 @@
...
+          return (
+            <b key={subStr} title={matches[0].name}>
+              {subStr}
+            </b>
+          );
+        }
+      }
+      return subStr;
+    });
+  };
```
DavidHollins6, line 34: @AmeeMorris had mentioned there was no mentions highlighting. It's back! I had taken it out thinking we'd do it in app, but I think it makes sense here

lib/Comment/CommentMeta.js
```diff
@@ -1,41 +1,63 @@
-import React from 'react';
-import { func, string } from 'prop-types';
+import React, { Fragment } from 'react';
+import { bool, func, string } from 'prop-types';
 import Button from '../Button';
 
-export function CommentMeta({ createdAt, onEditClick, onRemoveClick }) {
+export function MetaDot() {
+  return <span className="text-xs text-neutral-primary mr-1 relative">•</span>;
+}
```
DavidHollins6, line 7: A tiny helper component for displaying that dot that appears in places on comments!

lib/Comment/CommentResolveButton.js
```diff
@@ -0,0 +1,56 @@
+import React from 'react';
+import { bool, func } from 'prop-types';
+import Icon from '../Icon';
+import { MetaDot } from './CommentMeta';
+import Button from '../Button';
+
+export function CommentResolveButton({
```
DavidHollins6, line 7: This is a new compound component, in the other PR it lived inside `Conversation.Header`, but now i've extracted it out so it's completely separate!

lib/Comment/CommentSubscribeToggle.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import { bool, func, string } from 'prop-types';
+import CheckToggle from '../CheckToggle';
+
+export function CommentSubscribeToggle({ onToggle, isChecked, id }) {
```
DavidHollins6, line 5: Same with the above Resolve component, just extracting it from the header

lib/Comment/EditCommentForm.js
```diff
@@ -0,0 +1,145 @@
+import React, { useState, useRef } from 'react';
+import { arrayOf, bool, func, shape, string } from 'prop-types';
+import { Mention, MentionsInput } from 'react-mentions';
+import Avatar from '../Avatar';
+import AvatarInformation from '../Avatar/AvatarInformation';
+import ShortcutTrigger from '../ShortcutTrigger';
+import Button from '../Button';
+import { MetaDot } from './CommentMeta';
+
+export function EditCommentForm({
```
DavidHollins6, line 10: Comments originally had a single component for creating a new comment and editing an existing one. I have no separated them out to different ones, so we can explicitly render the correct looking form.

lib/Comment/EditCommentForm.js
```diff
@@ -0,0 +1,145 @@
...
+    e.preventDefault();
+
+    onSubmit(value);
+  };
+
+  return (
+    <form onSubmit={handleSubmit}>
+      <div ref={inputRef}>
+        <MentionsInput
+          className="edit-comment w-full text-sm "
```
DavidHollins6, line 64: So `react-mentions`... It seems to take the first class that's passed into `MentionsInput`, and place a class with that name appended with `__input` on the `textarea` that it renders. So to style it I had to write some actual css to target that `textarea`

lib/Comment/NewCommentForm.js
```diff
@@ -0,0 +1,142 @@
+import React, { useState, useRef } from 'react';
+import { arrayOf, bool, func, shape, string } from 'prop-types';
+import { Mention, MentionsInput } from 'react-mentions';
+import Avatar from '../Avatar';
+import AvatarInformation from '../Avatar/AvatarInformation';
+import ShortcutTrigger from '../ShortcutTrigger';
+import Button from '../Button';
+
+export function NewCommentForm({
```
DavidHollins6, line 9: This is the other half of comment forms. This one is for a new comment!

lib/Comment/NewCommentForm.js
```diff
@@ -0,0 +1,142 @@
...
+
+  return (
+    <form onSubmit={handleSubmit}>
+      <Avatar
+        className="absolute mr-2"
+        url={author.avatar}
+        initials={author.initials}
+      />
+      <div ref={inputRef} className="pl-10 mb-2">
+        <MentionsInput
```
DavidHollins6, line 66: `react-mentions`...

lib/Comment/styles.scss
```diff
@@ -0,0 +1,56 @@
...
+    left: 40px;
+    top: 96px;
+    background: linear-gradient(
+      to bottom,
+      rgba(255, 255, 255, 0) 0%,
+      rgba(253, 253, 253, 0.5) 50%,
+      #fff 100%
+    );
+  }
+}
```
DavidHollins6, line 14: This is a pseudo element that blurs out the bottom of a comments text if it's too long.

Tailwind seems to get a bit weird when you want to position things absolutely, it doesn't have a lot of values for `top, bottom, left, right`. 

It also doesn't seem to have any functionality for `linear-gradient`!

lib/Comment/styles.scss
```diff
@@ -0,0 +1,56 @@
...
+
+.new-comment__input {
+  @apply border-0 p-2;
+}
+
+.edit-comment__input {
+  @apply border-0 pr-2 py-0 pl-0;
+}
```
DavidHollins6, line 37: `react-mentions`...

lib/Comment/styles.scss
```diff
@@ -0,0 +1,56 @@
...
+
+.conversation-comment__line {
+  &:before {
+    @apply absolute h-full border-l-2 border-solid border-neutral-90 border-r-0 border-b-0;
+    content: ' ';
+    top: 38px;
+    left: 22px;
+  }
+}
```
DavidHollins6, line 46: This is the line that joins up the avatars. Tailwind doesn't like pseudo elements + `top left bottom right` values, hence it's here.

lib/Comment/styles.scss
```diff
@@ -0,0 +1,56 @@
...
+
+.conversation-blur-content {
+  background: rgba(255, 255, 255, 0.75);
+
+  @supports (filter: blur()) {
+    transition: filter $animation-time-micro/2;
+    filter: blur(4px);
+    background: none;
+  }
+}
```
DavidHollins6, line 56: This is the blur for when you click `Delete` on a comment, I can't seem to find a solution for most of these styles in tailwind.

lib/Conversation/Comment.js
```diff
@@ -113,8 +113,7 @@ class Comment extends Component {
     const userCanEdit = createdBy === user.id && userCanComment;
     const className = cx('conversation__comment', {
       'is-visually-disabled': this.state.confirmRemoval,
-      'has-failed': this.props.hasFailed,
-      'has-author-type': author && author.type
+      'has-failed': this.props.hasFailed
```
DavidHollins6, line 116: I have removed the `author-type` class from comments, the author type is now in brackets next to the users name!

lib/Conversation/ConversationFooter.js
```diff
@@ -1,59 +1,14 @@
...
-        onCommentChange={onCommentChange}
-        focusOnMount={focusOnMount}
-        placeholder={placeholder}
-        onRowCountChange={onRowCountChange}
-        users={users}
-      />
-    </div>
-  );
+export function ConversationFooter({ children }) {
+  return <div className="p-2">{children}</div>;
```
DavidHollins6: So the footer used to house the new comment form, but now it just renders children. Which probably will still be the new comment form

lib/Conversation/ConversationHeader.js
```diff
@@ -1,84 +1,18 @@
...
-  id,
-  icon,
-  onSubscribeChange,
-  isSubscribed,
-  resolved,
-  handleUndoResolve,
-  userCanResolve,
-  resolveConversation
-}) {
+export function ConversationHeader({ children }) {
```
DavidHollins6: Same with the header, it now just renders children instead

lib/Icon/index.js
```diff
@@ -101,6 +101,7 @@ import warningTriangleSVG from '../../assets/icons/warning-triangle.svg';
 import warningOctogonSVG from '../../assets/icons/warning-octogon.svg';
 import infoSquareSVG from '../../assets/icons/info-square.svg';
 import integrationSVG from '../../assets/icons/integration.svg';
+import resolvedSVG from '../../assets/icons/resolve.svg';
```
DavidHollins6, line 104: Turns out the icon for resolved wasn't using our `Icon` component, so I put it in there!

lib/Person/index.js
```diff
@@ -12,13 +12,11 @@ const Person = ({ person, className }) => (
     />
     <div>
       <span className={`person__name ${className}__person-name`}>
-        {person.name}
+        {person.name}{' '}
+        {person.type && (
+          <span className="text-neutral-primary">({person.type})</span>
+        )}
```
DavidHollins6, line 18: We now render the author type inside brackets next to the name

tailwind.config.js
```diff
@@ -90,11 +90,16 @@ module.exports = {
         }
       }
     },
+    maxHeight: {
+      '32': '8rem'
+    }
+  },
```
DavidHollins6, line 96: turns out max-height doesn't have values you'd want, so I added my own!

tailwind.config.js
```diff
@@ -90,11 +90,16 @@ module.exports = {
...
+  variants: {
+    padding: ['last']
```
DavidHollins6, line 98: We need to specify which properties we can use `last:` on, so here I am saying we can now set it for `padding`.

Tailwind say they don't default everything to have it because it would increase the bundle size considerably.

tailwind.config.js
```diff
@@ -2,21 +2,21 @@ module.exports = {
   theme: {
```
DavidHollins6, line 2: Prettier did some comma dangle fixes it seems.

DavidHollins6:
[How would we style conversations in a perfect world.. with tailwind ofc](https://trello.com/c/KLsNiC8f/241-how-would-we-style-conversations-in-a-perfect-world-with-tailwind-ofc)

lib/Comment/EditCommentForm.js
altaudio: Seems like the only solution if we can't pass tailwind classes to the element - this is bad design on `react-mentions` side

altaudio: The only way I can imagine us not using CSS here is by using the `inputRef` prop that `react-mentions` accepts and passes to it's input element to set the classes with JS.

I don't think this is any cleaner or netter than the CSS solution.

lib/Comment/styles.scss
altaudio: Looking at the tailwind docs, they're suggesting that for absolute positioning, you stick an item to the top/left/bottom/right of whatever parent element you want, and then use padding or margins instead of `top/left/bottom/right` values.

> Use the .{top|right|bottom|left|inset}-0 utilities to anchor absolutely positioned elements against any of the edges of the nearest positioned parent.
>
>Combined with Tailwind's padding and margin utilities, you'll probably find that these are all you need to precisely control absolutely positioned elements.

I think we should follow this pattern - we are using tailwind and we should follow it's rules almost religiously, it's great to have those patterns decided for you.

As for the gradient, we can either leave the SCSS file, or go back to design and say that we can't do a linear gradient as it's a technical limitation.

Both I think are fine options, redesigning the comments at this point is definitely out of scope 🔭

lib/Comment/styles.scss
```diff
@@ -0,0 +1,56 @@
...
+      rgba(253, 253, 253, 0.5) 50%,
+      #fff 100%
+    );
+  }
+}
+
+.conversation__inactive {
+  &:hover {
+    .conversation__text__cutoff {
+      &:after {
```
altaudio, line 19: As for the `after` pseudo selector, in a perfect world we would factor this to not use `after`. To be fair, after is a bit of a smell. It's sort of encroaching on Reacts territory of displaying rendering things based on conditions.

Again, out of scope though

lib/Comment/styles.scss
altaudio: Are we getting rid of these anyway in the new designs?

lib/Comment/styles.scss
AmeeMorris: Depends how far you want to go, it looks like we can extend the values for `top, bottom, left, right` in the theme, maybe we could have them match our new rem spacing values.

Gradient wise it looks like theres a couple of plug ins that people have made to add this in: https://www.npmjs.com/package/tailwindcss-gradients https://github.com/lorisleiva/tailwindcss-plugins/tree/master/gradients 

Both of these are totally optional and only worth it if you can foresee them being useful for other things too.

lib/Comment/styles.scss
altaudio: Same comments as before, if tailwind is our new master, the designs should heed to its technical limitations.

tailwind.config.js
altaudio: 🚩 Worth checking out these docs, as this implementation overrides all the `max-height` values, rather than adding your new value to them. You can extend it though 👉 https://tailwindcss.com/docs/theme/#extending-the-default-theme

lib/Comment/styles.scss
AmeeMorris: https://www.npmjs.com/package/tailwindcss-pseudo-elements there's a plug in that will apparently give better support for pseudo elements or this is probably config we can add in if we think its worth it (same with top/left/bottom/right values)

lib/Comment/styles.scss
AmeeMorris: Again missing styles could probably be added. For something like this maybe doing something like this would be neat https://tailwindcss.com/docs/extracting-components/#extracting-css-components-with-apply to have a `.blur` class we can use :)

tailwind.config.js
AmeeMorris: It'd be nice just to have all our rem values available here too. Bit sad that tailwind doesn't do that for us!

lib/Comment/styles.scss
DavidHollins6: I agree with that, and I had done it in other places. I will probably do the same for these now I'm looking at it

lib/Comment/styles.scss
DavidHollins6: > Looking at the tailwind docs, they're suggesting that for absolute positioning, you stick an item to the top/left/bottom/right of whatever parent element you want, and then use padding or margins instead of `top/left/bottom/right` values.

I read that same paragraph and completely misunderstood what it was saying. I will do this!

lib/Comment/styles.scss
DavidHollins6: > Are we getting rid of these anyway in the new designs?

The lines still appear on collapsed conversations

tailwind.config.js
DavidHollins6: Yeah I think I'll add them all onto our theme

PR #710, opened 2020-04-28
Update/Option Menu

altaudio:
### 💬 Description
Adds the `OptionMenu` component.

- Updates `DropdownContent` with our new GCDS border, shadow and border radius
- Adds `OptionMenu` component

Check it 👉 https://share.getcloudapp.com/KouqljJ0
### 🔗 Links
- GCDS Designs: https://www.figma.com/file/66iSurBAx32DWrJymo9NKL/GCDS-%E2%80%93-Web-UI?node-id=321%3A668
- Formatting Toolbar Designs: figma.com/file/nYWpuGuUnmpPMm0E4x0cKY/Static-Toolbar?node-id=266%3A0
### 📹 GIF (optional)
### 🚪 Start Points
### 👫 Related PRs (optional)

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook


lib/OptionMenu/Item.js
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import { string, func, bool } from 'prop-types';
+import cx from 'classnames';
+import Icon from '../Icon';
+
+function Item({ text, onClick, iconName, selected }) {
```
altaudio, line 6: Not item!

lib/OptionMenu/OptionMenu.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import { node } from 'prop-types';
+import { Item } from './Item';
+
+function OptionMenu({ children }) {
+  return <div>{children}</div>;
```
altaudio, line 6: A bit off to just have a `div` here that doesn't do anything, but I think in the future we may want to style the container, and it means we can use the compound component pattern.

altaudio:
[Add Dropdown Option Menu to GUI from GCDS](https://trello.com/c/EyhjoCxi/262-add-dropdown-option-menu-to-gui-from-gcds)

lib/Dropdown/DropdownContent.js
```diff
@@ -113,16 +113,19 @@ class DropdownContent extends Component {
...
+      `dropdown__content border border-solid border-neutral-90 rounded shadow ${className}`,
+      {
+        'is-active': showContent,
+        'dropdown__content--collapse': collapse,
+        'dropdown__content--right': right,
+        'dropdown__content--centre': centre,
+        'dropdown__content--top': top,
+        'dropdown__content--fix-right': fixRight,
+        'dropdown__content--full': full,
+        'dropdown__content--noborder': noBorder
```
AmeeMorris, line 126: 💚 Just checking these options all still work and tailwind doesnt override them?

lib/OptionMenu/MenuItem.js
```diff
@@ -0,0 +1,42 @@
...
+  const classes = cx(
+    'my-1 rounded-small w-48 border-0 flex items-center py-1 px-2 hover:bg-neutral-95 focus:outline-none focus:shadow-blue-focus-sm',
+    {
+      'text-neutral-20': !selected,
+      'text-blue-primary': selected
+    }
+  );
+
+  return (
+    <button className={classes} type="button" onClick={onClick}>
```
AmeeMorris, line 16: 💚 i think we should chuck in a `...rest` here too. I think its always handy for things like buttons. Maybe disabled could come in handy too?

lib/OptionMenu/OptionMenu.js
AmeeMorris: Do we wanna have the option to give the div some classes?

AmeeMorris: This component brings up a weird, point actually. It doesnt really do anything, just renders its children. 

Would we be better off just making `OptionMenu/MenuItem` a type of button or `DropdownAction`?

altaudio: It's definitely a smell just to have an empty div for the component, but the reasoning was to keep the GUI elements matched to the GCDS designs 👇

https://www.figma.com/file/66iSurBAx32DWrJymo9NKL/GCDS-%E2%80%93-Web-UI?node-id=321%3A0

In the designs the overarching element is `OptionMenu` and the sub element is `MenuItem`. Obviously these are open to change if they don't suit in the current form, but we should have a chat with @p34ks  before committing on our side.

I like the current pattern, as the designs suggest that further down the line we are going to have something like:
```
<OptionMenu>
  <OptionMenu.Heading>First Option Set</OptionMenu.Heading>
  <OptionMenu.MenuItem name="option 1" />
  <OptionMenu.MenuItem name="option 2" />
  <OptionMenu.MenuItem name="option 3" />
  <OptionMenu.Divider />
  <OptionMenu.MenuItem name="option 4" />
  <OptionMenu.Heading>Second Option Set</OptionMenu.Heading>
  <OptionMenu.MenuItem name="option 5" />
</OptionMenu>
```

I'm not sure about having the `MenuItem` as a generic button, as it has specific functionality that ties it to being a menu item (eg. the `selected`) prop.

My take away from the chat yesterday was that we were moving away from the `Dropdown` component rendering specific UI such as `DropdownAction`, and moving towards it just providing functionality to wrap a trigger and render arbitrary content when it is clicked. In this case, the arbitrary content is the `OptionMenu` and it's children.

lib/Dropdown/DropdownContent.js
altaudio: Great catch! TY - checked these, and they're all good 👍

lib/OptionMenu/OptionMenu.js
DavidHollins6: I really like how we would use it and feel fine with the fact that `OptionMenu` sort of does nothing. I think adding a `...rest` prop would be all we need, so we could still pass it classes or a test id.

What would this look like with the trigger?

AmeeMorris: Sure, makes sense. Thanks for explaining!

💚 agree that it should be a `...rest` on the div though. I originally said just the className so my bad! 🙈

PR #716, opened 2020-04-30
Refactor ConversationContext

DavidHollins6:
### 💬 Description
ConversationContext is what's used in the comment history. So i've refactored them to use the new components!
### 🚪 Start Points
Top to bottom...

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook


lib/ConversationContext/styles.scss
```diff
@@ -17,7 +17,7 @@
       border-left: 2px solid $color-border-base;
       position: absolute;
       top: 100%;
-      height: $layout-spacing-base*2;
+      height: 17px;
```
DavidHollins6, line 20: This is because the new conversations are a bit further away, so I increased the size of the line that connects them. I didn't want to fall down a hole so I just changed the value, maybe it's something else we could refactor later!?

DavidHollins6:
[Integrate the new conversation components into comment history](https://trello.com/c/tynLsxUu/273-integrate-the-new-conversation-components-into-comment-history)

lib/ConversationContext/styles.scss
kyleharper: If possible we should fix this soon but I'm down for letting this pass as I assume you only just found this etc.

PR #719, opened 2020-05-06
Conversation refactor mob feedback

DavidHollins6:
### 💬 Description
Just some little fixes in here, They are as follows:

- Boundary click watcher boundaries were a bit dodgy. I have moved the margin up a level to exclude the boundary.

- `split of undefined` error. This was because we don't refresh the comments body when it's created. I had to do a bit of useEffect fanciness to fix this.

- CMD + Enter didn't submit the edit form. This was because focus was going a bit funky. I fixed how it sets whether it's being focussed in react-mentions.

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook

### DIFF
--- a/lib/Comment/EditCommentForm.js
+++ b/lib/Comment/EditCommentForm.js
@@ -67,8 +67,8 @@ export function EditCommentForm({
           className="edit-comment w-full text-sm "
           value={value}
           onChange={handleChange}
-          onFocus={() => setInputHasFocused(!inputHasFocused)}
-          onBlur={() => setInputHasFocused(!inputHasFocused)}
+          onFocus={() => setInputHasFocused(true)}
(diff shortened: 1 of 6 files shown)


lib/Comment/EditCommentForm.js
```diff
@@ -67,8 +67,8 @@ export function EditCommentForm({
           className="edit-comment w-full text-sm "
           value={value}
           onChange={handleChange}
-          onFocus={() => setInputHasFocused(!inputHasFocused)}
-          onBlur={() => setInputHasFocused(!inputHasFocused)}
+          onFocus={() => setInputHasFocused(true)}
```
altaudio, line 70: Like this explicitness! 👍

PR #721, opened 2020-05-07
Feature/toolbar tweaks

altaudio:
### 💬 Description
A few tweaks to we had to make while adding the GCDS components for the toolbar.

7 small changes - I have separated them out so following this description should make things clear!
#### Added caret SVG to `<Icon />`
- https://github.com/gathercontent/gather-ui/pull/721/files#diff-6d3fc66085c15c4c4eeb0fe542dddfb3
- https://github.com/gathercontent/gather-ui/pull/721/files#diff-b1b664b8ef093b0b15721ca8078625ea
#### Removed margins from buttons
We decided buttons shouldn't have margins by default, as they would invariably be wrong
- https://github.com/gathercontent/gather-ui/pull/721/files#diff-e51058fb24dd7c5fb4c1b3e9a86691f8
#### Adds active prop to `ButtonTertiary`
To force active styling
- https://github.com/gathercontent/gather-ui/pull/721/files#diff-89af281b98705ead2d6e7874c2dcd73b
- https://github.com/gathercontent/gather-ui/pull/721/files#diff-3f1907ce8fffe5ac2402ba784cab25af
#### Adds `renderChildrenOnly` prop to `DropdownTrigger`
Allows you to bypass rendering a button from `DropdownTrigger`. This moves more towards the `Dropdown` component being purely functional.
- https://github.com/gathercontent/gather-ui/pull/721/files#diff-86e36a5fa72a075c30c5084921d7f083
#### Adds `className` prop to OptionMenu, MenuItem and Toolbar
- https://github.com/gathercontent/gather-ui/pull/721/files#diff-866a907e47d582458244d02c5eb437ac
- https://github.com/gathercontent/gather-ui/pull/721/files#diff-03fb8e0b57620220485c896b712f404c
- https://github.com/gathercontent/gather-ui/pull/721/files#diff-3b6086a3fcb6507f5b08e0c6d91dce4c
#### Adds `shadow` prop to TopBar, so we can toggle it's shadow on and off
- https://github.com/gathercontent/gather-ui/pull/721/files#diff-a0d91f3bef3bb7238f9cc29e8db9cac3
- https://github.com/gathercontent/gather-ui/pull/721/files#diff-9dc0d6b40c65529e6f0fdbece3d319c8
#### Sets the important flag in tailwind.
We finally had to do it. This needs testing!
- https://github.com/gathercontent/gather-ui/pull/721/files#diff-8927db3d5d9a070423893c3ca4c68042
### 🔗 Links
https://www.figma.com/file/nYWpuGuUnmpPMm0E4x0cKY/Fixed-Toolbar?node-id=266%3A0
### 📹 GIF (optional)
### 🚪 Start Points
### 👫 Related PRs (optional)

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook


lib/Dropdown/DropdownTrigger.js
```diff
@@ -48,6 +49,10 @@ class DropdownTrigger extends Component {
 
     const buttonTypes = useSelect ? types.concat('outline') : types;
 
+    if (renderChildrenOnly) {
+      return children;
```
AmeeMorris, line 53: Nice! One q though do we need to wrap this in the div so we can use the ref? (im not sure what the ref is used for without digging into it! so might be completely redundant)

altaudio: Yes we do, good catch 👍

altaudio: @AmeeMorris @DavidHollins6 

This was a great catch as it turns out! My implementation was totally wrong.

Not only do we need to wrap the `children` in a `div` to attach the `ref`, I also need to pass down `this.handleClick` from this function to `children`. This function does all the auto positioning logic.

Previously I was just using `setShowContent` that is passed out of the overarching `Dropdown` component, which does not take into account any autopositioning.

I have updated this so the component now checks to see if its `children` is a function, and then passes the correct `this.handleClick` function to it. Plus the `div` wrapping with the `ref`.

Commit is here! https://github.com/gathercontent/gather-ui/pull/721/commits/05d21fc239d43a3f20e5b5bd76e05a81179f94b7

PR #723, opened 2020-05-07
Update/pill input fix

altaudio:
### 💬 Description
Carys reported an issue where hitting enter when you were in the PillInput in the bulk share form would submit the form, when it is only supposed to add a Pill.

This PR fixes that by using `event.preventDefault()` for the enter key handler.

As a bonus, I have added the functionality to PillInput to delete a Pill by pressing backspace when you are in the input.

I have moved the keydown handlers to a custom hook, `useKeyDownHandlers`.

I also attached the listeners to the `input` element rather than the `document` element.

Video demo 👉 https://share.getcloudapp.com/E0ubypWk
### 🔗 Links
### 🚪 Start Points
https://github.com/gathercontent/gather-ui/pull/723/files#diff-206d0e1f7234e30c362c2abfdab1d40dlisting multiple files._
### 👫 Related PRs (optional)

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook

### DIFF
--- a/lib/PillInput/PillInput/PillInput.js
+++ b/lib/PillInput/PillInput/PillInput.js
@@ -36,7 +36,7 @@ const PillInput = ({
   };
 
   const removeLastPill = () => {
-    const newPills = pills.length > 1 ? pills.slice(0, pills.length - 1) : [];
+    const newPills = pills.slice(0, pills.length - 1);
     setPills(newPills);
   };
 
--- a/lib/PillInput/PillInput/PillInput.js
+++ b/lib/PillInput/PillInput/PillInput.js
@@ -1,8 +1,9 @@
-import React, { useState, useEffect } from 'react';
+import React, { useState, useEffect, useRef } from 'react';
 import uuid from 'uuid/v1';
 import { string, arrayOf, bool, instanceOf, shape, func } from 'prop-types';
 import cx from 'classnames';
-import { DeleteablePill } from './DeleteablePill';
+import { DeleteablePill } from '../DeleteablePill';
+import { useKeyDownHandlers } from './useKeyDownHandlers';
 
 const PillInput = ({
   className,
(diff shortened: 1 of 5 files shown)


altaudio:
[[REGULAR] When bulk sending items, clicking return after entering an email address will automatically close the share modal and send the items selected](https://trello.com/c/ROMCtfmd/256-regular-when-bulk-sending-items-clicking-return-after-entering-an-email-address-will-automatically-close-the-share-modal-and-sen)

lib/PillInput/PillInput/PillInput.js
```diff
@@ -34,6 +35,11 @@ const PillInput = ({
     setPills(newPills);
   };
 
+  const removeLastPill = () => {
+    const newPills = pills.length > 1 ? pills.slice(0, pills.length - 1) : [];
```
DavidHollins6, line 39: I'm not too up on how slice works, but if `pills.length` = 1, wouldn't slice still return `[]`? Meaning this ternary might not be needed?

lib/PillInput/PillInput/useKeyDownHandlers.js
```diff
@@ -0,0 +1,49 @@
...
+  useEffect(() => {
+    if (!inputRef) {
+      return () => {};
+    }
+
+    const KEYDOWN_EVENT = 'keydown';
+    document.addEventListener(KEYDOWN_EVENT, handleKeyDown);
+    return () => document.removeEventListener(KEYDOWN_EVENT, handleKeyDown);
+  }, [inputValue, inputRef, pills]);
+}
```
DavidHollins6, line 47: Would we need to check `inputValue.current`? As that reflects what the ref currently is.

lib/PillInput/PillInput/PillInput.js
altaudio: 100% 🙏 

I added this when I had another bug thinking it was a fix - then forgot to remove it 👍

PR #727, opened 2020-05-08
Add RepeatableSettings component to field

petemcfarlane:
### 💬 Description
This creates a `RepeatableSettings` component, used by the field to manage, would you believe, repeatable field settings.

There's a Delete, Move Up / Down buttons, and a label which shows the numeric index.

I've had to make a couple of tweaks to our tailwind config worth mentioning, namely I've added an animation curve which matches our existing animation curve used elsewhere in the Field component (to animate the field actions coming in and out, otherwise it looks weird if different animation curves/durations are used!) I've also enabled the opacity group-hover utility, so we can get hover effects on parents passed down to children - super cool!

There's a bit more to add to get this fully matching the design (adding the "Add another"/"Limit reached button" but I'll save that for another PR.

### 🔗 Links
https://www.figma.com/file/HNH1EAEtLOVJRsydtdrxUq/Repeatable-Fields?node-id=14%3A1454

### 🚪 Start Points
Storybook, tweak those knobs!

### 👫 Related PRs (optional)

### ✅ Checklist
- [ ] Tests written
- [x] Browser tested
- [ ] Added to documentation
- [x] Added to storybook

### DIFF
--- a/lib/Field/FieldInStructureEditModeProvider.js
+++ b/lib/Field/FieldInStructureEditModeProvider.js
@@ -1,0 +1,28 @@
+import React from 'react';
+import { node, bool } from 'prop-types';
+
+export const FieldInStructureEditModeContext = React.createContext({});
+
+export const FieldInStructureEditModeProvider = ({
+  children,
+  inStructureEditMode
+}) => {
+  const sharedState = {
+    inStructureEditMode
+  };
+
+  return (
+    <FieldInStructureEditModeContext.Provider value={sharedState}>
+      {children}
+    </FieldInStructureEditModeContext.Provider>
+  );
+};
+
+FieldInStructureEditModeProvider.propTypes = {
+  children: node.isRequired,
+  inStructureEditMode: bool
+};
+
+FieldInStructureEditModeProvider.defaultProps = {
+  inStructureEditMode: false
+};
--- a/lib/Field/FieldLabel.js
+++ b/lib/Field/FieldLabel.js
@@ -1,0 +1,36 @@
+import React, { useContext } from 'react';
+import { func, string } from 'prop-types';
+import EditableTextWrapper from '../EditableTextWrapper';
+import { FieldInStructureEditModeContext } from './FieldInStructureEditModeProvider';
+
+export const FieldLabel = ({ label, onChange, className }) => {
+  const { inStructureEditMode } = useContext(FieldInStructureEditModeContext);
+
+  if (inStructureEditMode) {
+    return (
+      <EditableTextWrapper
+        value={label}
+        className={`field__label ${className}`}
+        onChange={onChange}
+        multiline
+        inputLabel={`edit field label: ${label}`}
+      >
+        {label}
+      </EditableTextWrapper>
+    );
+  }
+
+  return <div className={`field__label ${className}`}>{label}</div>;
+};
+
+FieldLabel.propTypes = {
+  label: string,
+  onChange: func,
+  className: string
+};
+
+FieldLabel.defaultProps = {
+  label: '',
+  onChange: () => {},
+  className: ''
+};
--- a/lib/Field/FieldLeft.js
+++ b/lib/Field/FieldLeft.js
@@ -1,0 +1,27 @@
+import React, { useContext } from 'react';
+import Col from 'react-bootstrap/lib/Col';
+import { node, string } from 'prop-types';
+import { FieldInStructureEditModeContext } from './FieldInStructureEditModeProvider';
+
+export const FieldLeft = ({ children, className }) => {
+  const { inStructureEditMode } = useContext(FieldInStructureEditModeContext);
+
+  return (
+    <Col
+      xs={12}
+      sm={inStructureEditMode ? 3 : 4}
+      className={`field__data ${className}`}
+    >
+      {children}
+    </Col>
+  );
+};
+
+FieldLeft.propTypes = {
+  children: node.isRequired,
+  className: string
+};
+
+FieldLeft.defaultProps = {
+  className: ''
+};
--- a/lib/Field/FieldMiddle.js
+++ b/lib/Field/FieldMiddle.js
@@ -1,0 +1,20 @@
+import React from 'react';
+import { node, string } from 'prop-types';
+import Col from 'react-bootstrap/lib/Col';
+
+export const FieldMiddle = ({ children, className }) => (
+  <Col xs={12} sm={8} className={`field__content--wrapper ${className}`}>
+    <div className="field__content border border-solid border-neutral-90">
+      <div className="field__child">{children}</div>
+    </div>
+  </Col>
+);
+
+FieldMiddle.propTypes = {
+  children: node.isRequired,
+  className: string
+};
+
+FieldMiddle.defaultProps = {
+  className: ''
+};
--- a/lib/Field/RepeatableSettings.js
+++ b/lib/Field/RepeatableSettings.js
@@ -3,35 +3,20 @@ import { bool, number } from 'prop-types';
 import { ButtonIcon } from '../ButtonNew/ButtonIcon/ButtonIcon';
 import { ButtonIconDanger } from '../ButtonNew/ButtonIcon/ButtonIconDanger';
 
-const RepeatableSettings = ({
-  isTextDirLTR,
-  fieldIsActive,
-  repeatPosition,
-  isLastRepeat
-}) => {
+export const RepeatableSettings = ({ repeatPosition, isLastRepeat }) => {
   const label = `${repeatPosition < 9 ? '0' : ''}${repeatPosition + 1}`;
 
   return (
     <div
-      className={`
+      className="field__repeatable-settings
       text-neutral-primary
       text-lg
       leading-snug
       flex
       items-center
-      -my-3
-      ${isTextDirLTR ? 'ml-auto' : 'mr-auto'}
-    `}
+      -my-3"
     >
-      <div
-        className={`
-        ${fieldIsActive ? '' : 'opacity-0'}
-        group-hover:opacity-100
-        transition-opacity
-        duration-200
-        ease-animation-curve
-        flex`}
-      >
+      <div className="flex field__repeatable-settings-controls">
         {repeatPosition > 0 && (
           <ButtonIconDanger name="trash" className="repeat-icon m-0" />
         )}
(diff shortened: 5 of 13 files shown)


lib/Field/RepeatableSettings.js
```diff
@@ -0,0 +1,63 @@
...
+import { ButtonIcon } from '../ButtonNew/ButtonIcon/ButtonIcon';
+import { ButtonIconDanger } from '../ButtonNew/ButtonIcon/ButtonIconDanger';
+
+const RepeatableSettings = ({
+  isTextDirLTR,
+  fieldIsActive,
+  repeatPosition,
+  isLastRepeat
+}) => {
+  const label = `${repeatPosition < 9 ? '0' : ''}${repeatPosition + 1}`;
```
petemcfarlane, line 12: Is there a better way to do left pad in JS? Or is this fine 😄

lib/Field/RepeatableSettings.js
```diff
@@ -0,0 +1,63 @@
...
+  return (
+    <div
+      className={`
+      text-neutral-primary
+      text-lg
+      leading-snug
+      flex
+      items-center
+      -my-3
+      ${isTextDirLTR ? 'ml-auto' : 'mr-auto'}
```
petemcfarlane, line 23: I don't like how little _options_ like text direction trickle down to affect child components, but I don't know what else to do 🤷

lib/Field/RepeatableSettings.js
```diff
@@ -0,0 +1,63 @@
...
+    `}
+    >
+      <div
+        className={`
+        ${fieldIsActive ? '' : 'opacity-0'}
+        group-hover:opacity-100
```
petemcfarlane, line 29: This pattern means that if the prop `fieldIsActive` is true (e.g. when editing content in the field), it will just have `opacity: inherit` (i.e. fully visible).
The group-hover trick means that anytime someone hovers over the parent Field (with class `.group`) that these styles will be applied, so even if a field isn't active, if you hover anywhere over the field the opacity will animate to 100%. Pretty neat!

lib/Field/RepeatableSettings.js
```diff
@@ -0,0 +1,63 @@
...
+        transition-opacity
+        duration-200
+        ease-animation-curve
```
petemcfarlane, line 32: Added our own custom animation curve to match what we were already using for other parts, including the `FieldActions`

lib/Field/RepeatableSettings.js
```diff
@@ -0,0 +1,63 @@
...
+        flex`}
+      >
+        {repeatPosition > 0 && (
+          <ButtonIconDanger name="trash" className="repeat-icon m-0" />
```
petemcfarlane, line 36: Somewhat annoyingly, I can't override the margin on these components because of the way the ButtonIcon is written... adding `m-0` doesn't have any effect. I wonder if we could rearrange the way these classNames are merged in the ButtonIcon, so we can override our own styles - any ideas @altaudio ? - context is I wanted to reduce the horizontal margin between the buttons.

lib/Field/index.js
```diff
@@ -8,12 +8,13 @@ import FieldValidations from './FieldValidations';
 import FieldActions from './FieldActions';
 import ExpandingTextArea from '../ExpandingTextArea';
 import EditableTextWrapper from '../EditableTextWrapper';
+import { RepeatableSettings } from './RepeatableSettings';
 
 const Field = props => {
-  const classNames = cx(`field field--${props.dir}`, {
+  const classNames = cx(`group field field--${props.dir}`, {
```
petemcfarlane, line 14: This `group` class is used by tailwind for the group-hover, down inside `RepeatableSettings`. IMO it's a shame that this is not nearer, yet hidden in another file...

lib/Field/index.js
```diff
@@ -8,12 +8,13 @@ import FieldValidations from './FieldValidations';
...
     field__editor: props.canEdit,
     'has-formatting': props.hasFormatting,
-    'is-active': props.isActive,
+    'field-is-active': props.isActive,
```
petemcfarlane, line 17: Had to rename this, as we use `.is-active` class all over the place for passing fills down on SVGs

lib/Field/styles.scss
```diff
@@ -314,7 +313,7 @@ $field-data-p-line-height: 1.688rem !default;
   }
 }
 
-.field.has-formatting {
+.has-formatting .field__content {
```
petemcfarlane, line 316: This still looks right to me but I'd like a second opinion here, as this could be a potentially big change if I got it wrong!

lib/Field/RepeatableSettings.js
DavidHollins6: Even if there was I wouldn't have an issue with this, it's readable!

lib/Field/RepeatableSettings.js
DavidHollins6: I've also had an issue with classes not overriding others. I found. that re-ordering the classes did little too. I'd be interested in hearing a solution

petemcfarlane:
Big thanks to @AmeeMorris who's helped with this PR feedback/shuffle. This is now a breaking change - we have split the `Field` component into a bunch of small `Field*` components, that are (arguably) easier to compose together. The last commit is a bit of a big PR, but take a look at the FieldStory for an example of how this can be used. 
Ended up refactoring some of the tailwind stuff out to just use the existing BEM styling we had - made things easier to reason about for now.
Feedback much appreciated 🙏

petemcfarlane:
Screenshot 2020-05-11 at 16 25 50

kyleharper:
Thanks for the refactoring @petemcfarlane & @AmeeMorris I'll review it today and get some comments added 👍

lib/Field/FieldInStructureEditModeProvider.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import { node, bool } from 'prop-types';
+
+export const FieldInStructureEditModeContext = React.createContext({});
+
+export const FieldInStructureEditModeProvider = ({
```
kyleharper, line 6: 🔴 After re-reading the PR a few times I'm still a little confused by the need for this context.

We've not got the ability to pass in props directly to the child components to say if they are `editable` or not so do we need this anymore?

---

The only argument I can see is the need to set this once and only once. So let's dig a little deeper;

There is currently three occurrences of this context. Twice for showing form elements and once for deciding what layout class we render.

For the layout class, I have added some separate feedback (basically we don't need to do this). 

(IMO) The other 2 instances don't support the need to have context. It's actually quite nice to pass the editable prop directly to the child for readability.

```
// This label is editable cause I can see the prop - no need for a new abstraction
<FieldLabel editable />
```

It's also fixed, which is quite strange. If it was sharing state that change based on user interaction then that would make more sense because we're abstracting logic but here we're just sharing a prop...with composable components.

-----

Let me know if you need to pair on this for clarification.

lib/Field/FieldLabel.js
```diff
@@ -0,0 +1,36 @@
...
+        {label}
+      </EditableTextWrapper>
+    );
+  }
+
+  return <div className={`field__label ${className}`}>{label}</div>;
+};
+
+FieldLabel.propTypes = {
+  label: string,
```
kyleharper, line 27: Should a label be required?

lib/Field/FieldLeft.js
```diff
@@ -0,0 +1,27 @@
+import React, { useContext } from 'react';
+import Col from 'react-bootstrap/lib/Col';
+import { node, string } from 'prop-types';
+import { FieldInStructureEditModeContext } from './FieldInStructureEditModeProvider';
+
+export const FieldLeft = ({ children, className }) => {
```
kyleharper, line 6: 🔴 What was the thinking around `FieldLeft`? Naming things is hard but left doesn't tell us anything about what it should house. Here are some ideas;

- `FieldData` (funnily enough this has a class name for this)
- `FieldInformation`
- `FieldAside`

Another point is we have right to left settings in some usages. Does `FieldLeft` make sense in those cases?

lib/Field/FieldLeft.js
```diff
@@ -0,0 +1,27 @@
...
+  const { inStructureEditMode } = useContext(FieldInStructureEditModeContext);
+
+  return (
+    <Col
+      xs={12}
+      sm={inStructureEditMode ? 3 : 4}
```
kyleharper, line 12: Why do we need to change this prop for the editable state in `FieldLeft` and not have to change anything for what is currently called `FieldMiddle`?

`sm` in middle is currently fixed at `8` so this should always be `4` to make up a `12` grid.

lib/Field/FieldMiddle.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import { node, string } from 'prop-types';
+import Col from 'react-bootstrap/lib/Col';
+
+export const FieldMiddle = ({ children, className }) => (
```
kyleharper, line 5: Similar feedback to `FieldLeft` there are possible names changes here;

- `FieldBody`
- `FieldContent`
- `FieldEditor`

What do you think?

lib/Field/RepeatableSettings.js
kyleharper: I and @DavidHollins6 found out why today. The class order in the CSS file is the reason. `m-0` is before the other classes so it will always lose.

lib/Field/stories/FieldStory.js
```diff
@@ -36,64 +43,84 @@ storiesOf('Components', module).add('Field', () => {
...
+        <FieldLeft
+          className={isRepeatable && !inStructureEditMode ? '' : 'pt-6'}
+        >
+          {isRepeatable && !inStructureEditMode && (
+            <RepeatableSettings
+              isLastRepeat={isLastRepeat}
+              repeatPosition={repeatPosition}
+            />
+          )}
+          <FieldLabel label={label} />
```
kyleharper, line 81: Should this be implemented as `Field.Label` as well as all the other components?

It doesn't make a massive difference other than clarity of structure.

lib/Field/FieldInStructureEditModeProvider.js
kyleharper: After chatting I can see the benefits of this!

If possible let's utilise the base `Field` component and share some state (using context) in there rather than having this abstraction.

lib/Field/FieldLeft.js
kyleharper: Dismiss this feedback. Too much effort to name things 😂

lib/Field/FieldMiddle.js
kyleharper: Dismiss this feedback. Too much effort to name things 😂

petemcfarlane:
Thanks @DavidHollins6 @AmeeMorris and @kyleharper had a chat about it and will be making some changes:
- Leave the existing field alone, create a `FieldNew` directory, which can start as a fresh start for field, that way we can continue making releases to gather-ui without blocking the app, and gradually strangle out the old `Field` from the app (structure editor, content editor and revisions view).
- Will pair with someone when we try to rebuild the field in the Content editor, something to do with memoization
- Will rename `FieldLeft` to `FieldMeta`, and `FieldMiddle` to `FieldContent`
- Will stick with the bootstrap `<Row>` and `<Col>` implementations for now - might be too much to refactor them as part of this work
- Will keep the `FieldInStructureEditModeContext` but will just move it to the parent `Field` component, rather than it's own provider file.

Did I miss anything major?

PR #728, opened 2020-05-11
Feature/icons field tweaks

AmeeMorris:
### 💬 Description
Some new icons and some styling tweaks for fields.

New icons are self explainatory.

Styling tweaks are all around having icons in fields, i've changed the 'is-active' class that we use everywhere to a 'field-active', that way my icons aren't needlessly blue just because i've clicked into the field. I've also changed some really broad selectors we have to ignore buttons, that way my buttons wont have their padding/margin removed for no reason.

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook


lib/Field/styles.scss
```diff
@@ -412,7 +412,7 @@ $field-data-p-line-height: 1.688rem !default;
     }
   }
 
-  *:last-child {
+  *:last-child:not(button) {
```
petemcfarlane, line 415: I also had a problem with the `.field.has-formatting` in that I _think_ it's only supposed to apply formatting to the content editor field bit, yet it is applying styling to the elements on the left (settings/validation/label). But maybe that's a problem outside your PR? 
See https://github.com/gathercontent/gather-ui/pull/727/files#diff-5871cfcb3e11ef3631cff523e1bce82cR316

AmeeMorris: sadly my stuff is inside the field so will be effected either way!

lib/Field/index.js
```diff
@@ -13,7 +13,7 @@ const Field = props => {
   const classNames = cx(`field field--${props.dir}`, {
     field__editor: props.canEdit,
     'has-formatting': props.hasFormatting,
-    'is-active': props.isActive,
+    'field-active': props.isActive,
```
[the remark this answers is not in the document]
AmeeMorris: I'll go with this one just for lazynesses sake, but I dont have any strong feelings about it either!

PR #730, opened 2020-05-12
Add secondary button component

petemcfarlane:
### 💬 Description
Adds the secondary button from the GC Design system. I've not added the danger variation yet, as it's not currently needed and I'm not feeling that generous 😄 

### 🔗 Links
https://www.figma.com/file/66iSurBAx32DWrJymo9NKL/GCDS-%E2%80%93-Web-UI?node-id=448%3A544

### 📹 GIF (optional)
Screen Recording 2020-05-12 at 11 38 am

### 🚪 Start Points
It's pretty similar to Button Primary. The ButtonBase I noticed set the border-0 property, which is awkward to override, therefore I implemented the blue/grey outlines with a new named box shadow in the tailwind.config.
You might argue a `border` might be better here, but to do that I'd have to refactor the `border-0` out of the ButtonBase and repeat it in all the other child components, ButtonPrimary, ButtonPrimaryDanger etc... At least this is consistent with the other button patterns, and it seems to work well too with collections of buttons.

### ✅ Checklist
- [ ] Tests written
- [x] Browser tested
- [ ] Added to documentation
- [x] Added to storybook


petemcfarlane:

petemcfarlane:
I refactored the new buttons to use a proper border instead of a box-shadow, the one caveat is that there's always a border (top, right, bottom and left) on the button groups - this is because we can't easily set a border colour to a specific side. But I don't think it looks that bad!

Screenshot 2020-05-15 at 21 43 09

I've also updated some of the shadows to better match the Figma designs, consolidate some unused tailwind shadows, and tried to split the code so it's somewhat easier to read. Let us know if you think it's better or worse!

lib/ButtonNew/ButtonIcon/ButtonIconDanger.js
```diff
@@ -17,17 +17,22 @@ function ButtonIconDanger({
...
-      'opacity-50': disabled,
+      'hover:bg-red-95 active:bg-red-90': !active,
       'bg-red-90': active
+    }
+  );
+
+  const classes = cx(
+    className,
+    {
+      [nonDisabledClasses]: !disabled,
```
petemcfarlane, line 34: oOooh, is that because `nonDisabledClasses` has already evaluated as a string by now?

altaudio:
@AmeeMorris Pete and I found that it was a nightmare to try and make all the hover, active and focus styles actual borders rather than box shadows, so we have updated this PR to have some as borders and some as shadows.

The rule that we followed is:

- If a button has a border when it is not in a `active/hover/focus` state, then we modify that border style when those states change
- If a button does not start with a border, we use a box shadow

Without this method we would have struggled to do the ButtonPrimary group.

PR #738, opened 2020-05-14
New Fields component (FieldNew)

petemcfarlane:
### 💬 Description
This is the ground work for adding the new Fields

I found it quite straightfowrad to remove <Col> and <Row> and replace with tailwind grid css, see what you think?

### 🚪 Start Points
### 👫 Related PRs (optional)

### ✅ Checklist
- [ ] Tests written
- [x] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook

### DIFF
--- a/lib/Field/FieldInstructions.js
+++ b/lib/Field/FieldInstructions.js
@@ -1,0 +1,59 @@
+import React, { useContext } from 'react';
+import { func, number, string } from 'prop-types';
+import Linkify from 'linkifyjs/react';
+import { FieldNew } from 'lib';
+import ExpandingTextArea from '../ExpandingTextArea';
+
+export const FieldInstructions = ({
+  children,
+  onChange,
+  minRows,
+  placeholder,
+  className
+}) => {
+  const { inStructureEditMode } = useContext(
+    FieldNew.InStructureEditModeContext
+  );
+
+  if (inStructureEditMode) {
+    return (
+      <ExpandingTextArea
+        type="text"
+        className={`field__instructions
+          text-sm
+          border-0 border-t rounded-none
+          ${className}`}
+        handleOnChange={onChange}
+        value={children}
+        placeholder={placeholder}
+        minRows={minRows}
+      />
+    );
+  }
+
+  if (children) {
+    return (
+      <div className={`field__instructions text-sm ${className}`}>
+        <Linkify>{children}</Linkify>
+      </div>
+    );
+  }
+
+  return null;
+};
+
+FieldInstructions.propTypes = {
+  children: string,
+  onChange: func,
+  minRows: number,
+  placeholder: string,
+  className: string
+};
+
+FieldInstructions.defaultProps = {
+  minRows: 1,
+  onChange: () => {},
+  children: '',
+  placeholder: 'Add field instructions here...',
+  className: ''
+};
--- a/lib/Field/FieldLabel.js
+++ b/lib/Field/FieldLabel.js
@@ -1,0 +1,45 @@
+import React, { useContext } from 'react';
+import { func, string } from 'prop-types';
+import { FieldNew } from 'lib';
+import cx from 'classnames';
+import EditableTextWrapper from '../EditableTextWrapper';
+
+export const FieldLabel = ({ children, onChange, className }) => {
+  const { inStructureEditMode } = useContext(
+    FieldNew.InStructureEditModeContext
+  );
+
+  if (inStructureEditMode) {
+    return (
+      <EditableTextWrapper
+        value={children}
+        className={cx(
+          className,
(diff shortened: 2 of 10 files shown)


lib/Field/FieldNew.js
```diff
@@ -0,0 +1,47 @@
+import React, { createContext } from 'react';
+import { bool, node, oneOf, string } from 'prop-types';
+import cx from 'classnames';
+import { FieldContent } from './FieldContent';
+import { FieldMeta } from './FieldMeta';
+import { FieldLabel } from './FieldLabel';
+import { FieldInstructions } from './FieldInstructions';
+
+const InStructureEditModeContext = createContext({
```
altaudio, line 9: Nice use of context here 👍

lib/Field/styles.scss
```diff
@@ -449,3 +450,28 @@ $field-data-p-line-height: 1.688rem !default;
...
+    resize: inherit;
+    overflow: inherit;
+  }
+  .field__instructions {
+    resize: inherit;
+    overflow: inherit;
+    height: inherit;
+  }
+}
+
```
altaudio, line 477: 🚩Can we do this with tailwind?

lib/Field/stories/FieldStory.js
```diff
@@ -36,65 +41,143 @@ storiesOf('Components', module).add('Field', () => {
...
+        title="FieldNew"
+        description="A newer, Field component, can contain Content and Meta"
+      >
+        <FieldNew
+          className={cx({ 'field-active': isActive })}
+          dir={dir}
+          inStructureEditMode={inStructureEditMode}
+        >
+          <FieldNew.Meta className="">
+            <FieldLabel
```
altaudio, line 64: 🚩 What do you think to doing `<FieldNew.Label />` here? As storybook is the only source of documentation, I think it would be great to push people towards using this method!

lib/Field/FieldNew.js
```diff
@@ -0,0 +1,47 @@
...
+  inStructureEditMode: bool
+};
+
+FieldNew.defaultProps = {
+  className: '',
+  dir: 'ltr',
+  inStructureEditMode: false
+};
+
+FieldNew.Content = FieldContent;
```
altaudio, line 43: 💚What do you think to also adding `FieldActions`, `FieldValidations` here, as they are both also used to compose a field?

lib/Field/FieldNew.js
altaudio: This raises another question, should the editing and viewing version of a field be the same component, of should we have `Field` and `EditableField` components with a shared base.

If we were going from the ground up, I imagine we would go with the latter. We like the idea of components having single responsibilities, and when we start to see `if` statements switching a component between modes, that is usually a sign that it may need splitting.

That however is a job for another day, as the current version is already a vast improvement 💯

lib/Field/FieldInstructions.js
```diff
@@ -0,0 +1,59 @@
...
+
+  if (children) {
+    return (
+      <div className={`field__instructions text-sm ${className}`}>
+        <Linkify>{children}</Linkify>
+      </div>
+    );
+  }
+
+  return null;
```
altaudio, line 42: 💚 Can we get rid of the `if`, and just return at 35. I feel like this would be ok, as we have defaulted children to an empty string.

lib/Field/FieldLabel.js
```diff
@@ -0,0 +1,45 @@
...
+    FieldNew.InStructureEditModeContext
+  );
+
+  if (inStructureEditMode) {
+    return (
+      <EditableTextWrapper
+        value={children}
+        className={cx(
+          `field__label block font-semi-bold leading-normal ${className}`,
+          { 'pr-5': false }
```
altaudio, line 18: 🚩 Does this need removing, as it will never be applied?

lib/Field/FieldInstructions.js
petemcfarlane: Hmm I don't want to render the <div> (which has a border top, padding etc) if we don't have any instructions. I could change it so it's up to the client to not render a FieldInstructions, but I think that might look a bit messy in the client code, something like
```jsx
fields.map(field => {
  return <FieldNew>
    <FieldContent
      instructions={field.instructions
        ? (<FieldNew.Instructions>{field.instructions}</FieldNew.Instructions>)
        : null
      }
    />
  </FieldNew>
}
```

whereas currently we can just do 
```jsx
fields.map(field => {
  return <FieldNew>
    <FieldContent
      instructions={<FieldNew.Instructions>{field.instructions}</FieldNew.Instructions>}
    />
  </FieldNew>
}
```
even when there are no instructions (just "")

PR #739, opened 2020-05-15
Add states to new fields

petemcfarlane:
### 💬 Description
Where previously we just had an `isActive` toggle (which greyed out the background when false) we will now show when fields have been added, deleted or moved.

The FieldNew takes a `state` prop, which is one of
`ACTIVE`
`ADDED`
`DELETED`
`MOVED_UP`
`MOVED_DOWN`
`UNCHANGED`

You can toggle between these in the storybook.
 
### 🔗 Links
https://www.figma.com/file/Q3f2o4VMBAb0RAbESE1Fat/Revisions?node-id=149%3A1850

### 📹 GIF (optional)
Screen Recording 2020-05-15 at 09 29 am

### 🚪 Start Points
### 👫 Related PRs (optional)

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [x] Added to storybook

### DIFF
--- a/lib/Field/FieldContent.js
+++ b/lib/Field/FieldContent.js
@@ -1,9 +1,15 @@
 import React, { useContext } from 'react';
-import { element, node, string } from 'prop-types';
+import { element, node, oneOf, string } from 'prop-types';
 import cx from 'classnames';
 import { FieldNew } from 'lib';
 
-export function FieldContent({ children, className, instructions }) {
+export const ADDED = 'added';
+export const DELETED = 'deleted';
+export const MOVED_UP = 'moved_up';
+export const MOVED_DOWN = 'moved_down';
+export const UNCHANGED = 'unchanged';
+
+export function FieldContent({ children, className, instructions, state }) {
   const { inStructureEditMode } = useContext(
     FieldNew.InStructureEditModeContext
   );
@@ -14,13 +20,18 @@ export function FieldContent({ children, className, instructions }) {
         className,
         'col-span-12 md:col-span-8',
         'col-start-1',
-        'border border-solid border-neutral-90',
+        'border-solid',
         'rounded',
         'shadow',
         {
           'p-6': !inStructureEditMode,
           'md:col-start-5': !inStructureEditMode,
-          'md:col-start-4': inStructureEditMode
+          'md:col-start-4': inStructureEditMode,
+          'border border-neutral-90': state === UNCHANGED,
+          'border-3 border-green-primary bg-green-98': state === ADDED,
+          'border-3 border-red-primary bg-red-98': state === DELETED,
+          'border-3 border-purple-primary bg-purple-98':
+            state === MOVED_DOWN || state === MOVED_UP
         }
       )}
     >
@@ -33,10 +44,12 @@ export function FieldContent({ children, className, instructions }) {
 FieldContent.propTypes = {
   children: node.isRequired,
   className: string,
-  instructions: element
+  instructions: element,
+  state: oneOf([ADDED, DELETED, MOVED_UP, MOVED_DOWN, UNCHANGED])
 };
 
 FieldContent.defaultProps = {
   className: '',
-  instructions: null
+  instructions: null,
+  state: UNCHANGED
 };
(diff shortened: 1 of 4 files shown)


petemcfarlane:

lib/Field/FieldContent.js
```diff
@@ -33,10 +48,12 @@ export function FieldContent({ children, className, instructions }) {
 FieldContent.propTypes = {
   children: node.isRequired,
   className: string,
-  instructions: element
+  instructions: element,
+  state: oneOf([ACTIVE, ADDED, DELETED, MOVED_UP, MOVED_DOWN, UNCHANGED])
```
AmeeMorris, line 52: 🚩Here's another annoying naming nag. I think state isnt a great name just because state can mean something else thats pretty core to react.

Status instead?

lib/Field/FieldLabel.js
```diff
@@ -14,7 +14,8 @@ export const FieldLabel = ({ children, onChange, className }) => {
       <EditableTextWrapper
         value={children}
         className={cx(
-          `field__label block font-semi-bold leading-normal ${className}`,
+          className,
+          'field__label block font-semi-bold leading-normal -mr-2',
```
AmeeMorris, line 18: 💚Is there a way we dont have to have this minus margin 👀 ?

petemcfarlane: No 😄 
Basically, the EditableTextWrapper has that floating pencil, which we were moving with a bit of negative margin, so the label lines up with the actions/validations around it. If you have a suggestion I'm all ears? That EditableTextWrapper was a bit of a pain in general, trying to control the size of the text area, suspicious scroll bars, and Right To Left text support... grrr.

lib/Field/FieldContent.js
petemcfarlane: Thanks, this is good feedback, `state` has now been renamed to `status` 👍

lib/Field/FieldContent.js
```diff
@@ -14,13 +21,21 @@ export function FieldContent({ children, className, instructions }) {
...
         'rounded',
         'shadow',
+        'transition-all, transition-ease duration-200',
         {
           'p-6': !inStructureEditMode,
           'md:col-start-5': !inStructureEditMode,
-          'md:col-start-4': inStructureEditMode
+          'md:col-start-4': inStructureEditMode,
+          'border border-neutral-90': status === ACTIVE,
+          'border border-neutral-90 bg-neutral-98 group-hover:bg-white':
```
altaudio, line 33: Nice use of group-hover 👍 I guess the group class is already set on a parent somewhere?

lib/Field/FieldContent.js
```diff
@@ -14,13 +21,21 @@ export function FieldContent({ children, className, instructions }) {
...
+        'border-solid',
         'rounded',
         'shadow',
+        'transition-all, transition-ease duration-200',
         {
           'p-6': !inStructureEditMode,
           'md:col-start-5': !inStructureEditMode,
-          'md:col-start-4': inStructureEditMode
+          'md:col-start-4': inStructureEditMode,
+          'border border-neutral-90': status === ACTIVE,
```
altaudio, line 32: 💚 It might be quite clean to move the `status === ` styles to their own object.

Super 💚 comment, we could do this another time.

lib/Field/FieldNew.js
```diff
@@ -17,7 +17,7 @@ export function FieldNew({ children, className, dir, inStructureEditMode }) {
     <InStructureEditModeContext.Provider value={{ inStructureEditMode }}>
       <div
         className={cx(
-          'grid grid-cols-12 gap-4 field field-new has-formatting',
+          'group grid grid-cols-12 col-gap-5 field field-new has-formatting',
```
petemcfarlane, line 20: @altaudio this is where the `group` parent class is added

lib/Field/FieldContent.js
petemcfarlane: Not sure I fully understand, do you mean we'd have some consts like
```
const isActive = status === ACTIVE;
const isAdded = status === ADDED;
const isDeleted = status === DELETED;
// and so on...
```

PR #740, opened 2020-05-15
Add the Repeable field controls component

petemcfarlane:
### 💬 Description
This will add a row above the field label, which contains the buttons to delete and reorder repeatable fields. It should only show when hovering over the field, or when the field is in an active state.
The repeat position determines which button should show, as well as calculating the repeat position label text, e.g. "01", "02", "03"...

### 🔗 Links
https://www.figma.com/file/HNH1EAEtLOVJRsydtdrxUq/Repeatable-Fields

### 📹 GIF (optional)
Screen Recording 2020-05-15 at 09 34 am

### 🚪 Start Points
### 👫 Related PRs (optional)

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook

### DIFF
--- a/lib/Field/FieldRepeatableControls.js
+++ b/lib/Field/FieldRepeatableControls.js
@@ -7,20 +7,18 @@ export function FieldRepeatableControls({ repeatPosition, isLastRepeat }) {
   const label = `${repeatPosition < 9 ? '0' : ''}${repeatPosition + 1}`;
 
   return (
-    <div className="flex flex-row-reverse">
-      <div className="flex">
-        {repeatPosition > 0 && (
-          <ButtonIconDanger name="trash" className="mx-1 repeat-icon" />
-        )}
-        {repeatPosition > 0 && (
-          <ButtonIcon name="arrowUp" className="mx-1 repeat-icon" />
-        )}
-        {!isLastRepeat && (
-          <ButtonIcon name="arrowDown" className="mx-1 repeat-icon" />
-        )}
-        <span className="w-1" />
-        <span className="leading-8 text-neutral-primary text-lg">{label}</span>
-      </div>
+    <div className="flex justify-end">
+      {repeatPosition > 0 && (
+        <ButtonIconDanger name="trash" className="mx-1 repeat-icon" />
+      )}
+      {repeatPosition > 0 && (
+        <ButtonIcon name="arrowUp" className="mx-1 repeat-icon" />
+      )}
+      {!isLastRepeat && (
+        <ButtonIcon name="arrowDown" className="mx-1 repeat-icon" />
+      )}
+      <span className="w-1" />
+      <span className="leading-8 text-neutral-primary text-lg">{label}</span>
     </div>
   );
 }
--- a/lib/Field/FieldRepeatableControls.js
+++ b/lib/Field/FieldRepeatableControls.js
@@ -1,22 +1,31 @@
 import React from 'react';
-import { bool, number } from 'prop-types';
+import { bool, number, string } from 'prop-types';
+import cx from 'classnames';
 import { ButtonIcon } from '../ButtonNew/ButtonIcon/ButtonIcon';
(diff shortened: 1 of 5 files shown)


petemcfarlane:

lib/Field/FieldRepeatableControls.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import { bool, number } from 'prop-types';
+import { ButtonIcon } from '../ButtonNew/ButtonIcon/ButtonIcon';
+import { ButtonIconDanger } from '../ButtonNew/ButtonIcon/ButtonIconDanger';
+
+export function FieldRepeatableControls({ repeatPosition, isLastRepeat }) {
+  const label = `${repeatPosition < 9 ? '0' : ''}${repeatPosition + 1}`;
+
+  return (
+    <div className="flex flex-row-reverse">
```
DavidHollins6, line 10: Any particular reason for the `row-reverse`?

petemcfarlane: Yes - this was to get the row to "float" to the right - is there a better way to do this?? I didn't want do use margin-left: auto, because then I'd need to swap it when the text dir is RTL - this seems to work magically with RTL.

petemcfarlane: ah we can use `flex-end` to justify to the right instead 😅 knew there was a better way...

lib/Field/styles.scss
```diff
@@ -464,4 +464,19 @@ $field-data-p-line-height: 1.688rem !default;
...
+  &:hover {
+    .repeat-icon {
+      opacity: 1;
+    }
+  }
+  &.field-active {
+    .repeat-icon {
+      opacity: 1;
+    }
+  }
```
altaudio, line 481: 💚 What forced your hand into using SCSS over tailwind at this point? Was it the animations?

If so, it may be worth dropping them to keep everything in tailwind 🤔

lib/Field/FieldRepeatableControls.js
```diff
@@ -0,0 +1,29 @@
...
+      {repeatPosition > 0 && (
+        <ButtonIconDanger name="trash" className="mx-1 repeat-icon" />
+      )}
+      {repeatPosition > 0 && (
+        <ButtonIcon name="arrowUp" className="mx-1 repeat-icon" />
+      )}
+      {!isLastRepeat && (
+        <ButtonIcon name="arrowDown" className="mx-1 repeat-icon" />
+      )}
+      <span className="w-1" />
```
altaudio, line 20: 💚 Is this possible with padding rather than adding a new HTML element - empty tags are potentially confusing for screen reader users (conjecture here, David will know better!)

lib/Field/styles.scss
petemcfarlane: Fair enough, I've re-written this piece to use tailwind. My reason was because the Validations and Actions (these are NOT new components, they are used by both Field and FieldNew) are written with this pattern in mind, and for the sake of not rewriting them, I thought it would be consistent to keep it in SCSS, but I do prefer to have it in tailwind. I think we'll just treat Validations and Actions SCSS as tech debt, for another PR 👍

lib/Field/FieldRepeatableControls.js
petemcfarlane: @DavidHollins6 any input here?

The reason I went with a span is because I'm using flex and justify end, to align all the buttons and labels to the right. When I change text-dir to RTL the span is in the right place already. If I was to use a padding on the label, I'd have to swap from `pr-1` to `pl-1` depending on the text direction, and so I'd have to pass the `dir` down as a prop to the controls too. See what I mean here:

Screenshot 2020-05-19 at 10 22 56

But I'm happy to remove the span and go with padding if that would be better?

PR #747, opened 2020-05-19
Feature/more repeatable field controls

petemcfarlane:
### 💬 Description
I _think_ this may be the last of the repeatable field PRs, thanks for your help and patience!

This adds the vertical line between repeatable fields, and the "Add another" or "Limit reached" buttons at the bottom of the last repeated field.

I've also added 2 more fields to the story, for a bit more context. You can see the vertical spacer between all fields, and notice that we don't get the "trash" or "move up" button on the first field, and we don't get the "move down" button on the last repeated field.

I've had to override the bottom margin on the <FieldNew> to get the next field to join to the previous field (when repeated). Do you think this is fine? Another option I though was to not set an explicit margin on the <FieldNew> but set a negative margin (the size of the vertical bar) so the next elements would rise up to meet the line - but that also seems a bit hacky? 

### 🔗 Links
https://www.figma.com/file/HNH1EAEtLOVJRsydtdrxUq/Repeatable-Fields?node-id=0%3A1

### 📹 GIF (optional)

Screen Recording 2020-05-19 at 11 32 am

### ✅ Checklist
- [ ] Tests written
- [x] Browser tested
- [ ] Added to documentation
- [x] Added to storybook

### DIFF
--- a/lib/Field/FieldRepeatButton.js
+++ b/lib/Field/FieldRepeatButton.js
@@ -1,60 +1,25 @@
-import React, { useContext } from 'react';
-import { bool, string } from 'prop-types';
-import cx from 'classnames';
-import { FieldNew } from 'lib';
+import React from 'react';
+import { bool } from 'prop-types';
 import { ButtonSecondary } from '../ButtonNew/ButtonSecondary/ButtonSecondary';
 import Icon from '../Icon';
 
-function ConnectorLine() {
-  return (
-    <span className="h-10 border-l border-0 border-solid border-neutral-90 " />
-  );
-}
-
-export function FieldRepeatButton({
-  isLastRepeat,
-  repeatLimitReached,
-  className
-}) {
-  const { inStructureEditMode } = useContext(
-    FieldNew.InStructureEditModeContext
-  );
+export function FieldRepeatButton({ repeatLimitReached }) {
+  if (repeatLimitReached) {
+    return <ButtonSecondary disabled>Limit reached</ButtonSecondary>;
+  }
 
   return (
-    <div
-      className={cx(
-        className,
-        'col-span-12 md:col-span-8',
-        'col-start-1',
-        'flex flex-col items-center',
-        {
-          'md:col-start-5': !inStructureEditMode,
-          'md:col-start-4': inStructureEditMode
-        }
-      )}
-    >
-      {<ConnectorLine />}
-      {isLastRepeat &&
-        (repeatLimitReached ? (
-          <ButtonSecondary disabled>Limit reached</ButtonSecondary>
-        ) : (
-          <ButtonSecondary>
-            <Icon name="plus" className="mr-2" />
-            Add another
-          </ButtonSecondary>
-        ))}
-    </div>
+    <ButtonSecondary>
+      <Icon name="plus" className="mr-2" />
+      Add another
+    </ButtonSecondary>
   );
 }
 
 FieldRepeatButton.propTypes = {
-  isLastRepeat: bool,
-  repeatLimitReached: bool,
-  className: string
+  repeatLimitReached: bool
 };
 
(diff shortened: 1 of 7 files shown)


petemcfarlane:

lib/Field/FieldRepeatableControls.js
```diff
@@ -1,45 +1,47 @@
...
   );
 
   return (
-    <div className={classes}>
-      {repeatPosition > 0 && <ButtonIconDanger name="trash" className="mx-1" />}
-      {repeatPosition > 0 && <ButtonIcon name="arrowUp" className="mx-1" />}
-      {!isLastRepeat && <ButtonIcon name="arrowDown" className="mx-1" />}
+    <div className="flex justify-end">
+      {repeatPosition > 0 && (
+        <ButtonIconDanger name="trash" className={buttonClasses} size="sm" />
```
altaudio, line 25: 💚 We've attached the size information to the new buttons, so in this case you can do 
```suggestion
        <ButtonIconDanger name="trash" className={buttonClasses} size={ButtonIconDanger.sizes.sm} />
```

Not a blocking change, just a knowledge share! We thought it was a nice way of doing it, as most IDE's will automatically show you what sizes you have available through autocomplete, so you don't have to dip into GUI!

lib/Field/FieldRepeatButton.js
```diff
@@ -0,0 +1,60 @@
...
+import { ButtonSecondary } from '../ButtonNew/ButtonSecondary/ButtonSecondary';
+import Icon from '../Icon';
+
+function ConnectorLine() {
+  return (
+    <span className="h-10 border-l border-0 border-solid border-neutral-90 " />
+  );
+}
+
+export function FieldRepeatButton({
```
altaudio, line 14: 🚩I think this should really be two separate components, `FieldNew.FieldRepeatButton` and `FieldNew.FieldConnectorLine`.

Whenever we have a component that has an `if` statement in switching it between two different uses like this, it's usually a sign we would be better to factor the two use cases out to separate components.

This follows our single responsibility pattern, and we have found that although it is slightly more code, it is far easier to read than parsing a component with `if` statements switching things over.

lib/Field/stories/FieldStory.js
```diff
@@ -76,6 +76,176 @@ storiesOf('Components', module).add('Field', () => {
...
+                    </tr>
+                  </tbody>
+                </table>
+              </div>
+            ) : (
+              <div className="p-3" />
+            )}
+          </FieldNew.Content>
+          {isRepeatable && (
+            <FieldNew.RepeatButton
```
altaudio, line 158: With reference to my comment on splitting this component, this would make this more readable.

Currently I'm reading it, thinking "Ok we have a button here" - when in reality we are actually only rendering the divider. This is a mental model we don't have to investigate if instead we have here `FieldNew.ConnectorLine`

lib/Field/FieldRepeatableControls.js
petemcfarlane: Nice, thanks!
what do you think about limiting it in the button prop types too, so

```js
// ButtonNew/common.js
export const propTypes = {
  // ...
  // size: string, becomes
  size: oneOf(Object.keys(sizes)),
  // ...
}
```

lib/Field/stories/FieldStory.js
```diff
@@ -76,6 +76,180 @@ storiesOf('Components', module).add('Field', () => {
...
+                  </tbody>
+                </table>
+              </div>
+            ) : (
+              <div className="p-3" />
+            )}
+          </FieldNew.Content>
+          {isRepeatable && (
+            <FieldNew.MainColumn>
+              <FieldNew.ConnectorLine />
```
altaudio, line 162: 💯

PR #748, opened 2020-05-19
Sliding sidebar

DavidHollins6:
### 💬 Description
This is a sliding sidebar that will be used for the new commenting sidebar.

I ended up going with `react-spring` for animations, according to their github they're recommended by some hefty fellas. And it's also being used in some big projects. It had a really easy to use hook too, meaning I had the animation in minutes.

There is also an accompanying story that just has some conversations in to make it look nicer!

### 🔗 Links
### 📹 GIF (optional)
https://share.getcloudapp.com/X6uOLA15
### 🚪 Start Points
`lib/Sidebar/Sidebar.js`
### 👫 Related PRs (optional)

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook


lib/Sidebar/Sidebar.js
```diff
@@ -0,0 +1,52 @@
...
+      width,
+      [direction]: `-${width}`
+    },
+    enter: {
+      [direction]: '0'
+    },
+    leave: {
+      [direction]: `-${width}`
+    }
+  });
```
DavidHollins6, line 28: I had the biggest issue with `width`. The width of the sidebar is required because it then knows how far to animate to get the whole sidebar showing. I had thought about not passing in the width, and instead using ref and getting it's width. But that felt like a lot effort.

Instead we can just pass in a width, which can be `px` or `%`. It can't be `rem` or `em` because the `enter` value needs to be the same type! I think this approach is okay because currently we only intend to pass in a `%`, just enough `%` to cover the existing inline comments.

PR #753, opened 2020-05-21
FieldNew repeatable updates

petemcfarlane:
Export the field status consts from FieldNew so these can be accessed from the app:
ACTIVE, ADDED, DELETED, MOVED_UP, MOVED_DOWN, UNCHANGED

And add the button handlers to repeatable field controls/repeat button - didn't actually need these until we tried to port to the app 😄 

### ✅ Checklist
- [ ] Tests written
- [x] Browser tested
- [ ] Added to documentation
- [x] Added to storybook

### DIFF
--- a/lib/Field/FieldContent.js
+++ b/lib/Field/FieldContent.js
@@ -8,7 +8,7 @@ export const ADDED = 'ADDED';
 export const DELETED = 'DELETED';
 export const MOVED_UP = 'MOVED_UP';
 export const MOVED_DOWN = 'MOVED_DOWN';
-export const UNCHANGED = 'UNCHAGED';
+export const UNCHANGED = 'UNCHANGED';
 
 export function FieldContent({ children, className, instructions, status }) {
   const { inStructureEditMode } = useContext(
--- a/lib/Field/FieldContent.js
+++ b/lib/Field/FieldContent.js
@@ -3,12 +3,12 @@ import { element, node, oneOf, string } from 'prop-types';
 import cx from 'classnames';
 import { FieldNew } from 'lib';
 
-export const ACTIVE = 'active';
-export const ADDED = 'added';
-export const DELETED = 'deleted';
-export const MOVED_UP = 'moved_up';
-export const MOVED_DOWN = 'moved_down';
-export const UNCHANGED = 'unchanged';
+export const ACTIVE = 'ACTIVE';
+export const ADDED = 'ADDED';
+export const DELETED = 'DELETED';
+export const MOVED_UP = 'MOVED_UP';
+export const MOVED_DOWN = 'MOVED_DOWN';
+export const UNCHANGED = 'UNCHAGED';
 
 export function FieldContent({ children, className, instructions, status }) {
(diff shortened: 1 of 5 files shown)


lib/Field/FieldContent.js
```diff
@@ -3,12 +3,12 @@ import { element, node, oneOf, string } from 'prop-types';
...
-export const DELETED = 'deleted';
-export const MOVED_UP = 'moved_up';
-export const MOVED_DOWN = 'moved_down';
-export const UNCHANGED = 'unchanged';
+export const ACTIVE = 'ACTIVE';
+export const ADDED = 'ADDED';
+export const DELETED = 'DELETED';
+export const MOVED_UP = 'MOVED_UP';
+export const MOVED_DOWN = 'MOVED_DOWN';
+export const UNCHANGED = 'UNCHAGED';
```
AmeeMorris, line 11: Typo! 'UNCHAGED'

lib/Field/FieldRepeatableControls.js
```diff
@@ -1,12 +1,15 @@
...
 import cx from 'classnames';
 import { ButtonIcon } from '../ButtonNew/ButtonIcon/ButtonIcon';
 import { ButtonIconDanger } from '../ButtonNew/ButtonIcon/ButtonIconDanger';
 
 export function FieldRepeatableControls({
   repeatPosition,
   isLastRepeat,
+  deleteRepeatedField,
+  moveRepeatedFieldUp,
+  moveRepeatedFieldDown,
```
altaudio, line 12: Not a request for change here, as I think that this is a very opinionated comment and not anywhere near worth changing, but wanted to share for the heck of it anyway!

I like the convention of prefixing handlers with `on`, eg. `onDeleteRepeatedField`.

If we call it `deleteRepeatedField`, in my mental model we are suggesting that what is passed in must be the action to delete a field.

However, `onDeleteRepeatedField` to me is saying, this callback will be called when the UI button is pressed to delete a field. It doesn't matter what you pass in. I like this, as it suggests that this component is more flexible.

lib/Field/FieldRepeatButton.js
```diff
@@ -1,14 +1,14 @@
 import React from 'react';
-import { bool } from 'prop-types';
+import { bool, func } from 'prop-types';
 import { ButtonSecondary } from '../ButtonNew/ButtonSecondary/ButtonSecondary';
 import Icon from '../Icon';
 import { FieldMainColumn } from './FieldMainColumn';
 
-export function FieldRepeatButton({ repeatLimitReached }) {
+export function FieldRepeatButton({ repeatLimitReached, repeatField }) {
```
kyleharper, line 7: 🔴 is there a reason why we didn't use onClick instead of repeatField? I would prefer onClick here because it's a common pattern/name & it's easier to understand.

petemcfarlane: Ah ok, I think it was because I'd added the other handlers (trash, move up, move down) explicitly on the RepeatControls and I just added this at the same time, but i like onClick here also.

PR #754, opened 2020-05-22
Layout Components!

DavidHollins6:
### 💬 Description
A lot of discussions have been happening around how the sidebar should work, and the running theme throughout these discussions is that we need some sort of layout that we can adhere to. The main reasons were that we can apply a good naming format to each section of our layout, and we can also define what should happen when a sidebar appears!

The outcome is as follows:

```javascript
<Layout>
    <Layout.Header />
    <Layout.Main>
        <Layout.SubHeader />
        <Layout.Body />
        <Layout.Footer />
        <Layout.Sidebar />
    </Layout.Main>
</Layout>
```

A `Main` wraps around these components because they are what the sidebar can affect. For example the sidebar will shrink the `SubHeader` and the `Footer`, so they need to be grouped accordingly to avoid any nasty css shenanigans.

Here is what that looks like:

Screenshot 2020-05-22 at 11 13 20

Here is a gif of it moving: https://share.getcloudapp.com/OAubXN7j (I can't figure out how to embed gifs)

### 🚪 Start Points
The `Layout` components are all compound, so I recommend starting from the top: `lib/Layout/index.js`

And the story is very useful:  `stories/components/Layout.js`

### 👫 Related PRs (optional)

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook


DavidHollins6:
> One thing from the GIF, when you open the sidebar the border on the right hand side of the body disappears, is this expected?

Yep, the sidebar will shrink the `SubHeader` + `Footer`, but overlay the `Body`

PR #755, opened 2020-05-22
Removes specific colors from ButtonLink

altaudio:
### 💬 Description
Removed specific colours from button link.

Button link doesn't need to care about colours, these can bet set on the app side the same as if it was an anchor tag.

Also makes the background transparent.
### 🔗 Links
### 📹 GIF (optional)
### 🚪 Start Points
### 👫 Related PRs (optional)

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook


altaudio:
[Switch back to new toolbar once reverted](https://trello.com/c/sr1wYG5q/357-switch-back-to-new-toolbar-once-reverted)

PR #758, opened 2020-05-28
Tabs UI Refactor

DavidHollins6:
### 💬 Description
Hello!

This PR contains the _layout_ for the new tabs. These components provide only the behaviour of the new version of the tabs. This is intentional, I wanted tabs to be as reusable as possible, so tried to take the implementation as far away as possible from the gather-ui components. Meaning that when the tabs in the editors get implemented, it will just involve styling what a tab should _look_ like, not how it _acts_

To help with explaining the above paragraph, here is the new tabs components without any additional styling:

Screenshot 2020-05-28 at 14 41 34

And here it is with additional styling added on top of the layout: 

Screenshot 2020-05-28 at 14 54 36

### 🚪 Start Points
`lib/Tabs/TabsBase.js` - Tab
`lib/Tabs/index.js` - Tabs Container

I really recommend checking the story.

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook


lib/TabsNew/TabsBase.js
```diff
@@ -0,0 +1,19 @@
+import { func } from 'prop-types';
+
+export function TabsBase({ tabsLength, children }) {
+  const style = {
+    'flex-basis': 0,
+    'min-width': tabsLength >= 8 ? '12.5%' : 0,
+    'max-width': tabsLength >= 8 ? '12.5%' : 'initial'
+  };
+
+  return children('flex group flex-grow h-10', style);
```
AmeeMorris, line 10: 💚can we reasonably assume that a tab will always render a button and just plop a button here instead?

lib/TabsNew/index.js
```diff
@@ -0,0 +1,36 @@
...
+    'overflow-y-hidden': tabs.length <= 24,
+    'overflow-y-scroll': tabs.length > 24
+  });
+
+  return (
+    <div className="relative">
+      <nav className={`flex max-h-30 ${tabsClassName} ${className}`}>
+        {children}
+      </nav>
+      {actions && <div className="absolute right-0 mr-4">{actions}</div>}
```
AmeeMorris, line 17: 💚big opinion here, I'm not a huge fan of passing elements/components as props. Instead you could have...

```
<Tabs>
    <Tabs.Nav>
        some tabs as children
    </Tabs/Nav>
    <Tabs.Actions>
        a button or something
    </Tabs.Actions>
</Tabs>

File: lib/index.js [hunk 3]
```diff
@@ -70,7 +70,8 @@ export Shortcut from './Shortcut';
 export ShortcutIcon from './Shortcut/ShortcutIcon';
 export ShortcutTrigger from './ShortcutTrigger';
 export SectionHeader from './SectionHeader';
-export Tabs from './Tabs';
+export TabsLegacy from './Tabs';
```
altaudio, line 73: 💚 Can we change this filename to be `./TabsLegacy`, that way the new tab can just be called `./Tabs`?

lib/TabsNew/index.js
```diff
@@ -0,0 +1,36 @@
...
+
+function Tabs({ children, className, actions, tabs }) {
+  const tabsClassName = cx('flex flex-wrap w-full relative', {
+    'overflow-y-hidden': tabs.length <= 24,
+    'overflow-y-scroll': tabs.length > 24
+  });
+
+  return (
+    <div className="relative">
+      <nav className={`flex max-h-30 ${tabsClassName} ${className}`}>
```
altaudio, line 14: Nice semantic use 👍

lib/TabsNew/index.js
altaudio: 💚 Same, but echoing what Amee said, this is totally opinion on my part - unless we sit down and decide as a team we should go one way and not the other, this is as valid as Amee's suggestion.

lib/TabsNew/TabsBase.js
DavidHollins6: Unfortunately we can't assume that. In our case they are going to be `react-router` `Link`s. This component isn't in gather-ui either! So I felt it made most sense to make it a render props component

PR #762, opened 2020-06-02
Creating InlineSidebar and Section in Layout

DavidHollins6:
### 💬 Description
There are two biggie changes in this PR:

There is now a `InlineSidebar`, which is a sidebar that is not positioned absolute on top of the page, instead it is a part of the flow!

To help with this, there's another addition `Section`. This is just a layout component to group other layout components together. In this case, we have grouped the `SubHeader`, `Body` and `Footer` together in a `Section`, so the sidebar can be next to them!

### 🚪 Start Points
`stories/components/Layout.js` - You'll find examples of these new components in the story!

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook

### DIFF
--- a/lib/Layout/Body.js
+++ b/lib/Layout/Body.js
@@ -4,7 +4,7 @@ import { node, string } from 'prop-types';
 export function Body({ children, className }) {
   return (
     <div
-      className={`ql-scroll-container relative py-10 w-full flex flex-col flex-1 min-h-0 overflow-y-auto container ${className}`}
+      className={`ql-scroll-container relative w-full flex flex-col flex-1 min-h-0 overflow-y-auto container ${className}`}
     >
       {children}
     </div>
--- a/lib/Layout/InlineSidebar.js
+++ b/lib/Layout/InlineSidebar.js
@@ -1,0 +1,15 @@
+import React from 'react';
+import { node, string } from 'prop-types';
+
+export function InlineSidebar({ children, className }) {
+  return <div className={`${className}`}>{children}</div>;
+}
+
+InlineSidebar.propTypes = {
+  children: node.isRequired,
+  className: string
+};
+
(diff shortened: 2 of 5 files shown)


lib/Layout/Body.js
```diff
@@ -4,7 +4,7 @@ import { node, string } from 'prop-types';
 export function Body({ children, className }) {
   return (
     <div
-      className={`ql-scroll-container relative py-10 w-full flex flex-col flex-1 min-h-0 overflow-y-auto container ${className}`}
+      className={`ql-scroll-container relative w-full flex flex-col flex-1 min-h-0 overflow-y-auto container ${className}`}
```
DavidHollins6, line 7: I removed this as a small change in this PR because the body should not choose padding!

DavidHollins6:
[Refactor revision history to use new Layout](https://trello.com/c/MsnDBDsf/395-refactor-revision-history-to-use-new-layout)

lib/Layout/InlineSidebar.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import { node, string } from 'prop-types';
+
+export function InlineSidebar({ children, className }) {
+  return <div className={`${className}`}>{children}</div>;
```
altaudio, line 5: 🚩 Can we make this an `aside` tag to make it more semantic?

lib/Layout/Section.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import { node, string } from 'prop-types';
+
+export function Section({ children, className }) {
+  return <section className={`${className}`}>{children}</section>;
```
altaudio, line 5: 💚 Could also be:
```suggestion
  return <section className={className}>{children}</section>;
```

PR #763, opened 2020-06-02
Alter small ButtonIcon border radius

AmeeMorris:
### 💬 Description
Match GCDS designs for the border radius on small ButtonIcons

The order in the config controls what overrides what, which is why i've rejigged!


AmeeMorris:
[24px button icons should have 4px rounded corners](https://trello.com/c/7Vw38o9F/412-24px-button-icons-should-have-4px-rounded-corners)

PR #764, opened 2020-06-02
reduce toolbar height

AmeeMorris:
### 💬 Description
By removing the padding!

### DIFF
--- a/lib/Toolbar/Toolbar.js
+++ b/lib/Toolbar/Toolbar.js
@@ -5,7 +5,7 @@ import { Divider } from './Divider';
 function Toolbar({ children, className, ...rest }) {
   return (
     <div
-      className={`flex items-center border-0 border-b border-solid border-neutral-90 px-4 py-1 bg-white ${className}`}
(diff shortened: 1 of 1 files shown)


AmeeMorris:
[Toolbar shouldn't exceed 40px height](https://trello.com/c/dVo2qlV4/415-toolbar-shouldnt-exceed-40px-height)

lib/Toolbar/Toolbar.js
```diff
@@ -5,7 +5,7 @@ import { Divider } from './Divider';
 function Toolbar({ children, className, ...rest }) {
   return (
     <div
-      className={`flex items-center border-0 border-b border-solid border-neutral-90 px-4 py-1 bg-white ${className}`}
+      className={`flex items-center border-0 border-b border-solid border-neutral-90 px-4 bg-white ${className}`}
```
altaudio, line 8: 🚩 I think I originally added this padding as it was required when the text type dropdown button was on its focus state, worth checking whether this still looks ok 👍

AmeeMorris: Screenshot 2020-06-03 at 11 59 35
he's good!

PR #765, opened 2020-06-02
update icon default colour and ButtonIcon disabled opacity

AmeeMorris:
### 💬 Description
Icon default colours now use the new neutral primary colour, unfortunately I've had to do this the old fashioned way as a sass variable. Once everything is tailwindy we can make this nicer.

Also a new opacity setting and applying that to ButtonIcons

### DIFF
--- a/tailwind.config.js
+++ b/tailwind.config.js
@@ -131,6 +131,9 @@ module.exports = {
       },
       transitionTimingFunction: {
         'animation-curve': 'cubic-bezier(0.68, -0.55, 0.265, 1.55)'
+      },
+      opacity: {
+        '30': 0.3
       }
     },
     maxHeight: {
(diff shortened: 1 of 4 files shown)


AmeeMorris:
[Update our base neutral colour to match GCDS](https://trello.com/c/hvnqs6B7/416-update-our-base-neutral-colour-to-match-gcds)

AmeeMorris:
[Make disabled icon buttons 30% opacity](https://trello.com/c/q8nLtJaV/417-make-disabled-icon-buttons-30-opacity)

tailwind.config.js
```diff
@@ -131,6 +131,9 @@ module.exports = {
       },
       transitionTimingFunction: {
         'animation-curve': 'cubic-bezier(0.68, -0.55, 0.265, 1.55)'
+      },
+      opacity: {
+        '30': 0.3
```
altaudio, line 136: 💚 I think this overrides all the preset opacity styles, worth checking we aren't using them anywhere currently, as this might break them.

AmeeMorris: Double checked and its all good! I think it's fine because its in the extend object, if its just in theme then it overrides

PR #768, opened 2020-06-03
Add '30' width config

AmeeMorris:
### 💬 Description
We had this option for maxWidth, but not for min or regular width (which i needed!)

### DIFF
--- a/tailwind.config.js
+++ b/tailwind.config.js
@@ -134,7 +134,10 @@ module.exports = {
       },
       opacity: {
         '30': 0.3
-      }
+      },
+      width: {
+        '30': '7.5rem'
+      },
     },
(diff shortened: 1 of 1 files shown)


tailwind.config.js
```diff
@@ -134,7 +134,10 @@ module.exports = {
       },
       opacity: {
         '30': 0.3
-      }
+      },
+      width: {
```
altaudio, line 138: 💚 Does this override all other width styles? I know I left this comment before, so apologies if the answer is no and I didn't see it!

AmeeMorris: Nah its fine, I think its because it's set up in `extend` rather than just in the `theme`, so it adds rather than replaces

PR #771, opened 2020-06-08
Update/Add Select Component

altaudio:
### 💬 Description
Adds a `<Select />` component to GUI, built on top of `react-select`. Required for the following designs: https://www.figma.com/file/nYWpuGuUnmpPMm0E4x0cKY/Fixed-Toolbar?node-id=432%3A32

We don't have a `<Select />` definition in GCDS yet, but this implementation has been left open to change when that time comes.

- Creates `<Select />` component. API is exactly the same as [`react-select`](https://react-select.com/home)

- The chunky function called `mergeDefaultStyles` merges two sets of `react-select` style objects, so we can pass the `styles` prop to our new component, as well as having default styles.

- Adds with default styles, using [`react-select`'s `styles` object](https://react-select.com/styles#style-object)
  - As our theme is now in JS, we can import it and use whatever bits of it we need.
  - In order for this to work when GUI is transpiled, we had to add `tailwind.config.js` to the babel build (check the `package.json`)
### 🔗 Links
### 📹 GIF (optional)
https://share.getcloudapp.com/JruLNXJ9
### 🚪 Start Points
### 👫 Related PRs (optional)

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook

### DIFF
--- a/lib/Select/styles/defaultStyles.js
+++ b/lib/Select/styles/defaultStyles.js
@@ -4,10 +4,14 @@ const { theme } = tailwindConfig;
 
 const {
   extend: {
-    colors,
     colors: {
-      neutral: { primary: neutralPrimary },
-      blue: { primary: primaryBlue }
+      neutral: {
+        primary: neutralPrimary,
+        80: neutral80,
+        20: neutral20,
+        90: neutral90
+      },
+      blue: { primary: primaryBlue, 90: blue90, 95: blue95 }
     }
   },
   boxShadow: { default: defaultBoxShadow }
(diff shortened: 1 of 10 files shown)


altaudio:
[Add dropdown for switching reasons](https://trello.com/c/IC5iksLs/423-add-dropdown-for-switching-reasons)

lib/Select/styles/defaultStyles.js
```diff
@@ -0,0 +1,74 @@
...
+    color: neutralPrimary
+  }),
+
+  option: (provided, { isSelected, isFocused }) => {
+    const getBackgroundColor = () => {
+      if (isSelected) {
+        return colors.blue['90'];
+      }
+      if (isFocused) {
+        return colors.blue['95'];
```
altaudio, line 54: Deconstruct

PR #772, opened 2020-06-08
Creating a blue line for dragging tabs

DavidHollins6:
### 💬 Description
A utility span component to render a blue line for dragging and hovering over a tab. It supports a left line, right line or a whole border.

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook

### DIFF
--- a/lib/TabsNew/TabsDragLine.js
+++ b/lib/TabsNew/TabsDragLine.js
@@ -1,0 +1,29 @@
+import React from 'react';
+import { string, bool } from 'prop-types';
+
+export function TabsDragLine({ side, firstTab }) {
+  const leftPosition = firstTab ? '0' : '-2px';
+
+  return (
+    <span
+      style={{
+        left: side === 'left' ? leftPosition : 'unset',
+        right: side === 'right' ? '0' : 'unset'
+      }}
+      className={`${
+        side === 'whole'
+          ? 'border-2 w-full top-0 left-0'
+          : 'border-t-0 border-b-0 border-l border-r'
+      } absolute border-blue-primary border-solid top-0 h-full`}
+    />
+  );
+}
+
+TabsDragLine.propTypes = {
+  side: string.isRequired,
+  firstTab: bool
+};
+
+TabsDragLine.defaultProps = {
+  firstTab: false
+};
--- a/lib/TabsNew/index.js
+++ b/lib/TabsNew/index.js
@@ -3,6 +3,7 @@ import { node, string } from 'prop-types';
 import { TabsBase } from './TabsBase';
 import { TabsActionGroup } from './TabsActionGroup';
 import { TabsGroup } from './TabsGroup';
+import { TabsDragLine } from './TabsDragLine';
 
 function Tabs({ children, className }) {
   return <div className={`relative ${className}`}>{children}</div>;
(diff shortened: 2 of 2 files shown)


DavidHollins6:
[Implement new tabs inside Structure editor](https://trello.com/c/IpvgkObg/419-implement-new-tabs-inside-structure-editor)

lib/TabsNew/styles.scss
```diff
@@ -0,0 +1,17 @@
+.tab-drag-line {
```
altaudio, line 1: 🤔 It is becoming more and more apparent that `tailwind` is good for the generic stuff, but trying to do the nitty gritty stuff (like this) is a pain!

lib/TabsNew/TabsDragLine.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import { string } from 'prop-types';
+import cx from 'classnames';
+
+export function TabsDragLine({ side }) {
+  const className = cx('tab-drag-line absolute top-0 h-full z-10', {
```
altaudio, line 6: 💚 Can we just add side in like this?
```suggestion
  const className = cx(`tab-drag-line absolute top-0 h-full z-10 ${side}`, {
```

lib/TabsNew/TabsDragLine.js
```diff
@@ -0,0 +1,15 @@
...
+    [side]: true
+  });
+
+  return <span className={className} />;
+}
+
+TabsDragLine.propTypes = {
```
altaudio, line 13: 💚 You could also attach the side options to the component, like
```javascript
TabsDragLine.sides = {
  left: 'left',
  right: 'right',
  whole: 'whole',
}
```
either that, or use `oneOf(['left', 'right', 'whole'])` in the prop types. Currently we have to go in to the styles to know what our options are (or the story to be fair!)

Very 💚 💚 💚 comment!

stories/components/Tabs.js
```diff
@@ -47,6 +48,9 @@ function TabsStory({ tabs }) {
                     name="cog"
                   />
                 </div>
+                {dragSide && dragIndex === index && (
```
altaudio, line 51: 👍  Nice

lib/TabsNew/styles.scss
AmeeMorris: perhaps we could use [components](https://tailwindcss.com/docs/extracting-components/) for our more custom fiddly stuff

DavidHollins6: > 🤔 It is becoming more and more apparent that `tailwind` is good for the generic stuff, but trying to do the nitty gritty stuff (like this) is a pain!

100%, sometimes we just need some really nice specific styles!

DavidHollins6: > perhaps we could use [components](https://tailwindcss.com/docs/extracting-components/) for our more custom fiddly stuff

Ooo I'll have a look

lib/TabsNew/TabsDragLine.js
DavidHollins6: I'm into the first suggestion, it's like a built in const

PR #782, opened 2020-06-12
Feature/repeatable buttons loader

petemcfarlane:
### 💬 Description
When someone repeats/moves/deletes a repeatable field in the content editor we should show a loading state on the button.

I've introduced a `useLoader(action, initialState = false)` hook as an abstraction over the isLoading state - this is where the magic happens... 
It basically is a wrapper around a `useState()` and `isLoading` boolean, and returns a tuple of `isLoading :bool` and `actionLoader: func`. It takes a callable `action` - this is what we want to be called when we trigger `actionLoader`.

When we first trigger `actionLoader`, say by passing it to a button onClick handler, we set the state of `isLoading` to `true` and call `action`.

Any further calls to `actionLoader` are ignored, for as long as `isLoading` is `true` - this prevents double/multiple clicking errors.

The return value from triggering `actionLoader` is a function, than when called will revert the state of `isLoading` back to `false`. See the storybook for a fuller example.

This could potentially be abstracted even further in the future into a `<ButtonWithLoader>` component, or composable `withLoader()` function... but don't want to prematurely do that. Not sure the best place for this hook to live, seems like it _may_ be a useful pattern in other places?

### 📹 GIF (optional)

Screen Recording 2020-06-12 at 06 55 pm

### 🚪 Start Points
The first couple of commits

### ✅ Checklist
- [ ] Tests written
- [x] Browser tested
- [ ] Added to documentation
- [x] Added to storybook

### DIFF
--- a/lib/Field/useLoader.js
+++ b/lib/Field/useLoader.js
@@ -1,0 +1,14 @@
+import { useState } from 'react';
+
+export function useLoader(action, isInitiallyLoading = false) {
+  const [isLoading, setIsLoading] = useState(isInitiallyLoading);
+
+  const actionLoader = () => {
+    if (isLoading) {
+      return;
+    }
+    setIsLoading(true);
+    action(() => setIsLoading(false));
+  };
+  return [isLoading, actionLoader];
+}
--- a/lib/Field/useLoader.js
+++ b/lib/Field/useLoader.js
@@ -3,12 +3,13 @@ import { useState } from 'react';
 export function useLoader(action, isInitiallyLoading = false) {
   const [isLoading, setIsLoading] = useState(isInitiallyLoading);
 
-  const actionLoader = () => {
+  const actionLoader = async () => {
     if (isLoading) {
       return;
     }
     setIsLoading(true);
-    action(() => setIsLoading(false));
+    await action();
+    setIsLoading(false);
   };
   return [isLoading, actionLoader];
 }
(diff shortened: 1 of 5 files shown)


lib/Field/useLoader.js
```diff
@@ -0,0 +1,14 @@
...
+export function useLoader(action, isInitiallyLoading = false) {
+  const [isLoading, setIsLoading] = useState(isInitiallyLoading);
+
+  const actionLoader = () => {
+    if (isLoading) {
+      return;
+    }
+    setIsLoading(true);
+    action(() => setIsLoading(false));
+  };
```
altaudio, line 12: 🚩 I suspect that the most likely use case for this function will be with us passing in a promise as the action, so it's worth setting this up to accept a promise.

I think we will probably only use promises with this, so unless you have a specific non-promise use, I probably wouldn't bother with trying to make it handle both cases!

Pretty sure all you need to do is 
```suggestion
  const async actionLoader = () => {
    if (isLoading) {
      return;
    }
    setIsLoading(true);
    await action(() => setIsLoading(false));
  };
```

💚 In this case, we can also probably not bother with passing the callback to our action call, and just do
```suggestion
  const async actionLoader = () => {
    if (isLoading) {
      return;
    }
    setIsLoading(true);
    await action();
    setIsLoading(false)
  };
```
this is more of an opinionated comment, but I think if I was picking this component up I would expect it to stop loading once the promise passed was resolved. We also don't have to manually set is loading to false.

petemcfarlane: Great suggestion, thanks!

PR #784, opened 2020-06-16
Only update state if component is still mounted

petemcfarlane:
### 💬 Description
`setIsLoading()` comes from `useState()` which will trigger a rerender when called. If the component has been unmounted already, this will throw a warning about memory leaks. We only really want to call `setIsLoading()` if the component is still mounted. So we'll `useRef` to a boolean to keep track of whether the component `isMounted`, and use a `useEffect` to update `isMounted` when the component is unmounted.

### ✅ Checklist
- [ ] Tests written
- [x] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook

### DIFF
--- a/lib/Field/useLoader.js
+++ b/lib/Field/useLoader.js
@@ -1,15 +1,30 @@
-import { useState } from 'react';
+import { useEffect, useRef, useState } from 'react';
 
 export function useLoader(action, isInitiallyLoading = false) {
   const [isLoading, setIsLoading] = useState(isInitiallyLoading);
+
+  const isMounted = useRef(true);
+
+  useEffect(
+    () => () => {
+      isMounted.current = false;
+    },
+    []
+  );
 
   const actionLoader = async () => {
     if (isLoading) {
       return;
     }
+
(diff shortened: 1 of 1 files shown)


lib/Field/useLoader.js
```diff
@@ -1,15 +1,30 @@
...
 
 export function useLoader(action, isInitiallyLoading = false) {
   const [isLoading, setIsLoading] = useState(isInitiallyLoading);
 
+  const isMounted = useRef(true);
+
+  useEffect(
+    () => () => {
+      isMounted.current = false;
+    },
```
altaudio, line 11: 💚 You could one liner this one with
```suggestion
    () => () => isMounted.current = false,
```
but obvs not blocking as stylistic!

petemcfarlane: I had that originally but my IDE was complaining, I think it's because the non-block arrow function returns expression, whereas it's just expecting a void return type. I think I'll leave it for now 🙂
Screenshot 2020-06-17 at 10 16 50

PR #796, opened 2020-06-26
Creating generic class for pseudo border

DavidHollins6:
### 💬 Description
There have been a few designs lately that have required both a 2px blue border (for focus) and a 1px grey border (for blur) on inputs. We have previously had workarounds that involved fun times with padding. I then remembered we fixed it for hierarchy many moons ago. This PR takes that fix and makes a generic class in helpers that can be used elsewhere!

It now also uses the tailwind theme for colours, wahay!

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook

### DIFF
--- a/styles/main.scss
+++ b/styles/main.scss
@@ -1,4 +1,4 @@
-@import "tailwind";
+@import 'tailwind';
 
 /* ==========================================================================
    @ Settings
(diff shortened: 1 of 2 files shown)


styles/main.scss
```diff
@@ -1,4 +1,4 @@
-@import "tailwind";
+@import 'tailwind';
```
altaudio, line 1: This reminds me, I think prettier can also format CSS files if we set it up to

PR #797, opened 2020-06-26
Added a new prop to handle the height of a layout

DavidHollins6:
### 💬 Description
Our layout was a bit too rigid, it always had the height set the same as the view height. It meant we couldn't use it in other places that have similar layouts.

I originally tried just passing `h-full` into the `className` prop, but this is tailwind and it don't override. Instead I've added a prop that defaults to `true` (So there's no other refactoring), that will set that height to full screen.

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook

### DIFF
--- a/lib/Layout/index.js
+++ b/lib/Layout/index.js
@@ -1,5 +1,6 @@
 import React from 'react';
-import { node, string } from 'prop-types';
+import { node, string, bool } from 'prop-types';
+import cx from 'classnames';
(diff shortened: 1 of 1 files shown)


DavidHollins6:
[Blank slate (in sidebar)](https://trello.com/c/ObbOvoEI/523-blank-slate-in-sidebar)

PR #798, opened 2020-06-26
ButtonBase children function

DavidHollins6:
### 💬 Description
ButtonBase now accepts a function as a child, and it will pass through the styles to that function. That means I could render an `<a>` with the `ButtonTertiary` styles and everything is hunky dory.

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook

### DIFF
--- a/lib/ButtonNew/ButtonBase.js
+++ b/lib/ButtonNew/ButtonBase.js
@@ -29,6 +29,10 @@ function ButtonBase({
     }
   );
 
+  if (typeof children === 'function') {
+    return children(classes);
+  }
+
   return (
     <button className={classes} {...rest}>
       {loading && !loaderRight && (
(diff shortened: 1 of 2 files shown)


DavidHollins6:
[Feedback footer](https://trello.com/c/g06jie44/530-feedback-footer)

lib/ButtonNew/ButtonBase.js
```diff
@@ -29,6 +29,10 @@ function ButtonBase({
     }
   );
 
+  if (typeof children === 'function') {
```
altaudio, line 32: 💚 I worry a little that we may be under considering what adding this means.

If we pass the a function as a child, the loader props all become useless, which feels like bad design.

I definitely think that we should be able to use different element types for buttons, but this reminds me of our old button component, which was built to do just about everything in a single component, and as a result ended up being a pain in the ass to maintain.

I'm not 100% on the answer here, so this is not a blocker!

altaudio: 💚 Should we add a story to show the button styles being applied to an anchor?

PR #812, opened 2020-07-08
Fixes build:watch command, renames it to dev and adds source maps

altaudio:
#### ⁉ Issue
We previously had a command called `build:watch`, which was supposed to watch our `lib/` dir and `tailwind.config.js` file and build them into the `/dist` directory whenever they changed. However, I broke it with the following:

```javascript
{
  "build": "babel lib/ --out-dir dist/lib && babel tailwind.config.js --out-dir dist",
  "build:watch": "concurrently \"yarn copy:styles:watch\" \"yarn build --watch\""
}
```
looking at `build:watch`, when we call `yarn build --watch`, it is actually only passing the `--watch` flag to the second babel call. Thus it was only recompiling the tailwindconfig when that changed.
#### ✅ Fix
I have created a new command `yarn dev` that actually watches and rebuilds both  `lib/*` and `tailwind.config.js`.

As a bonus, this command adds inline source maps, which means we can debug the actual GUI code when linking it to the App.

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,7 @@
     "lint:fix": "eslint -c .eslintrc.json --fix 'lib/**/*.js' 'tests/**/*.js' 'stories/webapp/**/*.js'",
     "lint:check": "eslint --print-config . | eslint-config-prettier-check",
     "copy:styles:watch": "nodemon --exec \"yarn copy-styles\" --watch lib --watch styles -e scss --watch tailwind.config.js",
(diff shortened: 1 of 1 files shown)


altaudio:
[GUI: yarn build:watch is knackered](https://trello.com/c/5kiWStC6/545-gui-yarn-buildwatch-is-knackered)

altaudio:
> I'd rather still have watch in the name but 🤷‍♀️

Yeah I think I will keep it as `build:watch` now, originally I changed it to `dev` for a reason that isn't valid anymore, so I'll switch it back!

PR #817, opened 2020-07-10
Hiding conversation blur

DavidHollins6:
### 💬 Description
We should just hide the BottomBlur of a conversation instead of re-rendering it. This was causing issues because it was rendering and pushing other elements around.

### 🚪 Start Points
`lib/Comment/CommentText.js`

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook


DavidHollins6:
[Blur for long text goes 0% 100% 0%](https://trello.com/c/KuigrOA0/553-blur-for-long-text-goes-0-100-0)

PR #819, opened 2020-07-13
Conversation failure states

DavidHollins6:
### 💬 Description
Our conversation components now handle failures in their promises. As a result, we now need to show some UI to tell the user that their action has just failed, in this PR we add these states to Deleting + subscribing.

There is one small change that snuck in, we have a max-width around the conversation story now to make it more realistic.

### 📹 GIF (optional)
https://share.getcloudapp.com/jkuQ010g

### 🚪 Start Points
`lib/ConfirmationOverlay/index.js` - Delete promise failure state
`lib/Comment/CommentSubscribeToggle.js` - Subscribe promise failure state

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook


DavidHollins6:
[[UI CHANGE] Add fail states for other api request. E.g. subscribing/deleting](https://trello.com/c/PitmUin8/613-ui-change-add-fail-states-for-other-api-request-eg-subscribing-deleting)

lib/ConfirmationOverlay/index.js
```diff
@@ -13,10 +13,24 @@ const ConfirmationOverlay = ({
...
-  const [isSubmitting, handleSubmitWithLoader] = useLoader(confirm);
+  const [hasFailed, setHasFailed] = useState(false);
+  const setHasFailedTimeout = useRef(null);
+
+  const handleConfirm = async () => {
+    try {
+      await confirm();
+    } catch (error) {
+      setHasFailed(true);
+      setHasFailedTimeout.current = setTimeout(() => setHasFailed(false), 2000);
```
altaudio, line 28: 🚩 Do we need to handle canceling the timeout when this component unmounts too?

I wonder if there's a hook that could be abstracted here, something like `useTimeout`, which you pass a function and a delay, and it handles the canceling logic for you. Maybe you could also pass it multiple functions to call.

PR #821, opened 2020-07-14
Feature/dropdown new context

AmeeMorris:
### 💬 Description
This is switching our Dropdown component over to the new(ish) context api. 

Functionally it should all work the same and I havent done much to the components bar make them use the new context. One big change is I moved a lot of stuff from `Dropdown/index` over to `Dropdown/DropdownProvider` since we were getting some circular dependancy issues.

The other is that I've ditched a load of tests that broke as a result of this, I've rewritten those that seemed useful, but if you think I've been harsh with any of them let me know. 
### 🚪 Start Points
`lib/Dropdown/index.js`

### ✅ Checklist
- [x] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook


AmeeMorris:
[Refactor Dropdown to use new context](https://trello.com/c/47CjaVKx/623-refactor-dropdown-to-use-new-context)

PR #822, opened 2020-07-14
Counter Component

DavidHollins6:
### 💬 Description
This PR creates a `Counter` component. It's a lovely little blue counter that has a numerical value. I have also added the children prop to `ButtonIcon` so we can add in this Counter as a child!

THIS PR NEEDS MORE WORK - I need some tailwind/css guidance with this. I hoped there'd be a `group-active`.

### 📹 GIF (optional)
https://share.getcloudapp.com/rRulKbbg

### 🚪 Start Points
`lib/Counter/index.js`

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook


DavidHollins6:
[Add comment count to general comments icon button](https://trello.com/c/er5ofu9Q/609-add-comment-count-to-general-comments-icon-button)

lib/Counter/stories/Counter.js
```diff
@@ -0,0 +1,32 @@
...
+      <StoryItem title="Counter on it's own">
+        <Counter>{count}</Counter>
+      </StoryItem>
+      <StoryItem title="Counter on ButtonIcon">
+        <ButtonIcon
+          className="relative group"
+          name="commentsMulti16"
+          size={ButtonIcon.sizes.md}
+        >
+          <Counter className="absolute top-0 right-0 mt mr border-solid border-2 border-white group-hover:border-neutral-95 active:border-blue-90">
```
kyleharper, line 25: There seems to be quite a lot of classes here. Every time I need to use a counter in a ButtonIcon I'll have to remember to copy this from somewhere.

If we update this too we'll have to make sure we update it everywhere (not a great pattern for "re-usable" components IMO).

I don't think I'd mind if it was just `absolute top-0` but looking at the array of classes we need it feels like it should be a pattern described by the component itself.

---

This feedback also links with your need for tailwind support because `group-active` isn't a thing. But I think you've answered your own question, it isn't a thing because it shouldn't be.

If you want to style things based on an active state then move that thing into the component and give it a place there. There are plenty of ways to do this but maybe;

- add a prop to `ButtonIcon` ... this can be `Counter` and then you can decorate that with classes when the button is active. If I've misunderstood your active question, let me know 👍 (I'm assuming not though cause you wouldn't want to do anything with group active pseudo, right?)

lib/Counter/styles.scss
```diff
@@ -0,0 +1,5 @@
+.counter {
+  font-size: 10px;
+  line-height: 9px;
+  min-width: 14px;
```
kyleharper, line 4: How come we have fixed values like this?

lib/Counter/stories/Counter.js
AmeeMorris: I agree this is a lot of classes to remember to put in for each button icon we want to use this with. This could be an opportunity for a tailwind component https://tailwindcss.com/docs/extracting-components/ 

As for group active we could use https://github.com/benface/tailwindcss-interaction-variants

DavidHollins6: Kyle and I have discussed this further. I was trying to avoid the trap of coupling `Counter` and `ButtonIcon` with a prop. Kyle had the great idea of making this a compound pattern. `Counter` still exists on its own, but there is a variant called `ButtonIcon.Counter` which brings with it all of the specific styles to render that button at the top right.

lib/Counter/styles.scss
altaudio: We should try getting these to the nearest REM value if they look ok. Otherwise we maybe need to align GCDS and the designs better?

lib/ButtonNew/stories/ButtonIcon/ButtonIcon.js
```diff
@@ -22,6 +22,12 @@ storiesOf('Components', module).add('ButtonIcon', () => (
       <ButtonStoryItem title="Active">
         <ButtonIcon name="caption16" active />
       </ButtonStoryItem>
+
+      <ButtonStoryItem title="Counter">
+        <ButtonIcon name="caption16">
+          <ButtonIcon.Counter>9</ButtonIcon.Counter>
```
altaudio, line 28: Like this 👍

lib/ButtonNew/ButtonIcon/ButtonIcon.js
```diff
@@ -16,10 +17,11 @@ function ButtonIcon({
   active,
   size,
   iconTitle,
+  children,
   ...rest
 }) {
   const nonDisabledClasses = cx(
-    'bg-transparent border-transparent',
+    'bg-transparent border-transparent relative',
```
altaudio, line 24: 💚 Maybe all buttons should be relative, and we can apply this to `ButtonBase`?

lib/Counter/stories/Counter.js
DavidHollins6: UPDATE:

I hadn't seen Amees comment when I was replying to this!

I have ended up doing a lot of this in CSS. As stated above I started going down the compound component route, but all that compound component was doing was adding some extra styles, which I could just do in CSS. I started thinking that way and a lot of my other issues started becoming easier to fix. I could apply active styling to the counter now!

lib/Counter/styles.scss
DavidHollins6: I've converted these to REM. They're fixed non-tailwind values because tailwind doesn't let me be really specific. This is such a tiny component (like 20px max) that there are just no useful helpers that give me that precision. Everything else was fine except for these values.

lib/Counter/styles.scss
```diff
@@ -0,0 +1,24 @@
+.counter {
+  font-size: 0.625rem;
+  line-height: 0.5625rem;
+  min-width: 0.875rem;
+}
+
+.button__icon {
```
altaudio, line 7: 🚩 I feel slightly resistant to this, as this component is now tightly coupled with `<ButtonIcon />`, which I feel like will make maintenance and updating more complex in the long run. I think in the perfect decoupled world, this counter component should not know anything about `<ButtonIcon />`.

I think we could make things less complex if we made the styles component agnostic.

We could add a `border={true/false}` prop to `<Counter />`, which would render the border and apply the border active styles.

I don't feel like the positioning styles should live here ` absolute top-0 right-0 mt-px mr-px`. Maybe we create a new component `ButtonIconCounter` that extends this and applies the specific positioning formatting?

DavidHollins6: I'm going to reference this PR in our frontend tailwind discussion card. I feel like it'll be really relevant there.

PR #829, opened 2020-07-17
add a useShiftSelect hook

AmeeMorris:
### 💬 Description
This is adding a lovely hook to allow shift selection with the SelectionProvider.

We already have this code in a few places in the app, this is just that code but made a little more vague so we can hopefully use it for all the current shift select purposes! There's an upcoming app pr which makes the content hub active/trashed items & hierachy use this hook instead.

### ✅ Checklist
- [x] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook


AmeeMorris:
[Let's just use the one selection provider](https://trello.com/c/YeREhxCH/614-lets-just-use-the-one-selection-provider)

PR #830, opened 2020-07-21
Makes check toggle label a label element

altaudio:
### 💬 Description
Updates the `<CheckToggle />` label to be a `<label />` tag rather than a `<p />` tag.

Turns out an input can have multiple labels (you'll notice in the file we also have another label that is the actual toggle.

This update was added for targeting `<CheckToggle />` in tests.

> Related to https://github.com/gathercontent/app/pull/3154

### DIFF
--- a/lib/CheckToggle/index.js
+++ b/lib/CheckToggle/index.js
@@ -85,7 +85,11 @@ class CheckToggle extends Component {
 
     return (
       <div className={wrapperClasses}>
-        {labelLeft && <p className={wrapperLabelClasses}>{labelLeft}</p>}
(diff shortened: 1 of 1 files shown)


altaudio:
[Adds e2e test for plain text field formatting](https://trello.com/c/sGINsOrJ/644-adds-e2e-test-for-plain-text-field-formatting)

altaudio:
> bits of debt like this always make me cringe, why werent they a label to begin with! 🤦‍♀️

IKR! I think I put these in, and at the time I thought "we are already using a label, we ant have another!" which was a wrong assumption!

PR #833, opened 2020-07-29
Add min height for new fields

kyleharper:
Simples 😄 

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook

### DIFF
--- a/lib/Figure/FigurePreview.js
+++ b/lib/Figure/FigurePreview.js
@@ -20,7 +20,7 @@ const FigurePreview = ({
   if (previewBeingProcessed) {
     return (
       <div className="figure__title">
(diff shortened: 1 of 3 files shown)


lib/Figure/FigurePreview.js
```diff
@@ -20,7 +20,7 @@ const FigurePreview = ({
   if (previewBeingProcessed) {
     return (
       <div className="figure__title">
-        <div className="figure__loader-wrapper">
+        <div className="figure__loader-wrapper mt-10">
```
kyleharper, line 23: I bolted this on top as the margin was being cleared by field styling.

PR #836, opened 2020-07-29
Changing comment error message

DavidHollins6:
### 💬 Description
We never changed the content of the error message when submitting a comment. This PR does that!

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook

### DIFF
--- a/lib/Comment/CommentForm.js
+++ b/lib/Comment/CommentForm.js
@@ -166,7 +166,9 @@ function CommentForm({
         </>
       </div>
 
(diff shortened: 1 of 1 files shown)


DavidHollins6:
[We haven't set the error text for comments, so it just says "This is an error message"](https://trello.com/c/wAacGYFx/659-we-havent-set-the-error-text-for-comments-so-it-just-says-this-is-an-error-message)

PR #840, opened 2020-08-11
Updates Repeatable fields control button to not flicker on hover

altaudio:
### 💬 Description
The repeatable field controls up and down buttons were flickering on hover.

This was being caused by an animation class being inherited by the button from its parent, so on hover it was animating adding the background color, causing the flicker!

This PR updates the parents class name to only animate opacity, meaning the background color is no longer animated 🙏


altaudio:
[Content editor arrows are flashing when they are hovered over.](https://trello.com/c/cAjETAl2/464-content-editor-arrows-are-flashing-when-they-are-hovered-over)

PR #842, opened 2020-08-18
Remove hard coded group color fill on plus circle

petemcfarlane:
### 💬 Description
so we can control this with CSS

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook


altaudio (approved):
Change is 👍 

Probably best to go through the app and check that removing this does not affect any of the current uses of the plus-circle icon, unfortunately we don't have UI change testing yet!

AmeeMorris (approved):

kyleharper (approved):

DavidHollins6 (approved):

PR #843, opened 2020-09-01
Feature/col field

AmeeMorris:
### 💬 Description
Who doesn't want even more versions of a Field component? Here's another one! Since `Field` and `FieldNew` is already taken and in use I've decided on `ColField`. Once this is in the app `Field` will no longer be used so can be replaced/deleted, but to prevent a breaking version of gather-ui for a bit its remaining until thats done.

### 🚪 Start Points
check `lib/ColField/stories/ColFieldStory.js` which gives an outline of all the new `ColField` components, then comments for any other random changes.

### ✅ Checklist
- [x] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook


lib/ButtonNew/ButtonIcon/ButtonIconDanger.js
```diff
@@ -52,7 +52,7 @@ function ButtonIconDanger({
 
   return (
     <ButtonBase className={classes} disabled={disabled} size={false} {...rest}>
-      <Icon name={name} types={iconTypes} />
+      <Icon name={name} types={iconTypes} defaultActiveColor={false} />
```
AmeeMorris, line 55: Icon gets stuck as blue, even though it should be red when used as a dropdown trigger without this.

lib/CheckToggle/index.js
```diff
@@ -90,7 +93,7 @@ class CheckToggle extends Component {
             {labelLeft}
           </label>
         )}
-        <span className="toggle-wrapper__inner">
+        <span className={`toggle-wrapper__inner ${toggleClasses}`}>
```
AmeeMorris, line 96: Pretty self explanatory, needed to target the toggle itself with styling.

lib/ColField/index.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import { propTypes, defaultProps } from './common';
+import ColFieldHeader from './ColFieldHeader';
+import ColFieldLabel from './ColFieldLabel';
+import ColFieldBody from './ColFieldBody';
+import ColFieldFooter from './ColFieldFooter';
+import ColFieldInstructions from './ColFieldInstructions';
+
+function ColField({ children, className, ...props }) {
```
altaudio, line 9: 💚 I think we chatted about not using the `/index.js` filename, as a `ColField/ColField.js` is easier for IDEs to find.

We may not have gone down to the wire and discussed this yet though, so I'll raise it as a FE retro point!

lib/ColField/stories/ColFieldStory.js
```diff
@@ -0,0 +1,35 @@
...
+storiesOf('Components', module).add('ColField', () => {
+  return (
+    <StoryItem title="ColField" description="stuff">
+      <Row>
+        <Col xs={12} sm={8} smOffset={2}>
+          <ColField>
+            <ColField.Header>
+              <ColField.Label editable onChange={() => {}}>
+                Howdy
+              </ColField.Label>
```
altaudio, line 18: 💚 What do you think to passing `children` as a `value={}` prop instead?

I find it confusing that the children set the value, as it's not a pattern we commonly use. I think it may not be instantly obvious that this is the case when using this component.

I can see that this pattern has arisen from `<EditableTextWrapper />`

altaudio (approved):
Looks good, like the component! Such a nice composable pattern 💯

I've added a couple of green-y comments. I think the one around the use of `children` as a control value is the most weighted.

lib/ColField/stories/ColFieldStory.js
AmeeMorris: I kept it as children since value feels a bit weird when its not 'editable', value doesn't quite make sense. Maybe instead of value, having a 'label' prop instead, that feels right for both cases then?

lib/ColField/index.js
AmeeMorris: This is something that was raised! I just forgot (I think its because we have way more offending index files in GUI)

lib/ColField/stories/ColFieldStory.js
altaudio: I would say value is ok though, although it's not editable to the user, we are still controlling it with a value

PR #850, opened 2020-09-08
Feature/button icon enabled

AmeeMorris:
### 💬 Description
Adding a lovely green enabled state to ButtonIcon.

The diff for the classes has gone a bit weird, we're adding bits for the new green styles, but to complicate things we don't want that to override any active styles. 

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook

### DIFF
--- a/lib/ButtonNew/ButtonIcon/ButtonIcon.js
+++ b/lib/ButtonNew/ButtonIcon/ButtonIcon.js
@@ -20,15 +20,15 @@ function ButtonIcon({
   enabled,
   ...rest
 }) {
-  const nonDisabledClasses = cx(
-    'bg-transparent border-transparent',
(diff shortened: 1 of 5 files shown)


DavidHollins6 (approved):

DavidHollins6 (commented):
A non-blocking thought. The usual HTML patter would be to pass `disabled` to an element. Is. there a reason `enabled` here?

AmeeMorris:
> A non-blocking thought. The usual HTML patter would be to pass `disabled` to an element. Is. there a reason `enabled` here?

@DavidHollins6 here's a cool gif http://g.recordit.co/tgWmXjItjJ.gif basically we had a button that had an active state for a dropdown being active, but we also needed another kind of positive state to show when the setting inside the dropdown was enabled.

PR #852, opened 2020-09-10
Add styles for field editor, refactor Guideline

AmeeMorris:
### 💬 Description
Turns out we need all these custom styles for quill editors within fields. I was umming and ahhing on whether this belonged in the app or here, since they're quite app specific. These live in GUI in the Field styles, so i thought I'd just follow that.

I also needed to pass className over to Guideline and thought I'd do a cheeky refactor while I was there.

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook

### DIFF
--- a/lib/ColField/styles.scss
+++ b/lib/ColField/styles.scss
@@ -1,0 +1,87 @@
+.col-field__body .ql-editor {
+  *:nth-child(1) {
+    @apply mt-0;
+  }
+  h1:nth-child(1) {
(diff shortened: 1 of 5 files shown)


AmeeMorris:
[Make item editor use ColField](https://trello.com/c/77gAA7qg/552-make-item-editor-use-colfield)

altaudio (approved):

PR #863, opened 2020-09-22
Removing the margin from figure

DavidHollins6:
### 💬 Description
The figure in the asset field now is separated using the css grid gap, meaning we no longer need the margin!

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook

### DIFF
--- a/lib/Figure/styles.scss
+++ b/lib/Figure/styles.scss
@@ -30,7 +30,6 @@ $figure-border-radius: 3px !default;
   border-radius: $figure-border-radius;
   width: $figure-size;
   padding-bottom: $figure-size;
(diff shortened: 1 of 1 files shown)


petemcfarlane (approved):
Ah nice one! For some reason I thought the Figure was a gather-ui thing so not easy to fix!

DavidHollins6:
[Fix height of upload button when its on a row by itself.](https://trello.com/c/mXUx5KS4/805-fix-height-of-upload-button-when-its-on-a-row-by-itself)

AmeeMorris (approved):

PR #866, opened 2020-09-24
Modal Header with navigation

DavidHollins6:
### 💬 Description
A new component for our modals so we can have tabs in the header! Looking a little bit like this: https://share.getcloudapp.com/llu2j19J

Prettier had fun with the story file. I'll comment where the actual new story is.

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook

### DIFF
--- a/stories/components/Modals.js
+++ b/stories/components/Modals.js
@@ -6,6 +6,7 @@ import Col from 'react-bootstrap/lib/Col';
 import FormControl from 'react-bootstrap/lib/FormControl';
 import FormGroup from 'react-bootstrap/lib/FormGroup';
 import ControlLabel from 'react-bootstrap/lib/ControlLabel';
+import MenuItem from "react-bootstrap/lib/MenuItem";
 import {
   Modal,
   Button,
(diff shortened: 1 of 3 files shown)


stories/components/Modals.js
```diff
@@ -475,24 +477,62 @@ storiesOf('Components', module).add('Modals', () => (
...
+            title="Modal with ModalHeaderNavigation"
+            description="Modal with tabs in the header."
+        >
+            <ModalTrigger>
+                <Modal.Container size="large">
+                    <Modal.HeaderNavigation title="Title!">
+                        <Navigation tabs>
+                            <MenuItem active onClick={action('clickHandler')}>
+                                {/* eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex */}
+                                <div className="h-full w-full" role="menuitem" tabIndex={0} >
```
petemcfarlane: What do you think to making these buttons rather than divs? Or is a MenuItem already a button? Is this purely for the tabindex?

petemcfarlane (approved):

stories/components/Modals.js
DavidHollins6: This is purely for a tabindex, `MenuItem` is an anchor that doesn't have a href, and we only need the `onClick` for it.

DavidHollins6:
[Tabs in FilePickerModal](https://trello.com/c/QbeULpVe/814-tabs-in-filepickermodal)

PR #869, opened 2020-09-24
Add yellow fill helper

kyleharper:
We need to fill icons in yellow sometimes. Just following our existing fill helper pattern.


AmeeMorris:
We seem to have a few different ways of changing an icons fill colour at the moment:

We have these helpers
We have the 'types' prop on the icon
We also have a `.inherit-color-icon` class which makes the icon path match whatever the current text colour is. 

In my mind for any widely used colours we would add a type, and for one offs we would use the inherit prop. 

What do you think?

PR #870, opened 2020-09-29
Stop SearchInputs from spanning 100% width

AmeeMorris:
### 💬 Description
Like the title says, we don't want this to happen. 
Screenshot 2020-09-29 at 09 45 52
If we use inline-flex instead of flex the element doesnt take up the entire width it can. I've also added the ability to add classnames, since this always is needed at some point

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook

### DIFF
--- a/lib/SearchInput/index.js
+++ b/lib/SearchInput/index.js
@@ -10,7 +10,9 @@ function SearchInputField({
   hideIcon,
   value,
   label,
-  id
+  id,
+  className,
+  inputClassName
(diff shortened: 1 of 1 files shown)


DavidHollins6 (approved):

petemcfarlane (approved):
Whoops, thanks @AmeeMorris!

PR #871, opened 2020-10-02
Add SelectionModal component

AmeeMorris:
### 💬 Description
We have some designs for a new modal type (see figma), which I've decided to call a SelectionModal. We now have a new SelectionModal component that can be used with and in place of other Modal components, I think the new story gives a pretty clear idea on how it works.
### 🔗 Links
https://www.figma.com/file/HWmdpuNLjhxlKUhzYQJWPW/Components?node-id=248%3A783
### 🚪 Start Points
`stories/components/Modals.js`

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook

### DIFF
--- a/lib/SelectionModal/SelectionModalCategory.js
+++ b/lib/SelectionModal/SelectionModalCategory.js
@@ -1,0 +1,56 @@
+import React from 'react';
+import { bool, string, number } from 'prop-types';
+import cx from 'classnames';
+
+function SelectionModalCategory({
+  isSelected,
+  className,
+  name,
+  subText,
+  counter,
+  ...rest
+}) {
+  const buttonClasses = cx(
+    `bg-white relative border-solid flex w-full rounded outline-none px-3 py-2 border border-neutral-90 items-center ${className}`,
+    {
+      'text-input-border-focus': isSelected,
+      'hover:border-blue-primary': !isSelected
+    }
+  );
+
+  const nameClasses = cx('m-0 mb-1', {
+    'text-neutral-20': !isSelected,
+    'text-blue-primary': isSelected
+  });
+
+  return (
+    <button className={buttonClasses} type="button" {...rest}>
+      <div className="text-left">
+        <p className={nameClasses}>{name}</p>
+        {subText && (
+          <p className="text-sm text-neutral-primary m-0">{subText}</p>
+        )}
+      </div>
+      {counter !== 0 && (
+        <div className="text-blue-primary ml-auto">{counter}</div>
+      )}
+    </button>
+  );
+}
+
+SelectionModalCategory.propTypes = {
+  isSelected: bool,
+  className: string,
+  name: string.isRequired,
+  subText: string,
+  counter: number
+};
+
+SelectionModalCategory.defaultProps = {
+  isSelected: false,
+  className: '',
+  subText: '',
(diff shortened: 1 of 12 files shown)


stories/components/Modals.js
```diff
@@ -541,5 +543,78 @@ storiesOf('Components', module).add('Modals', () => (
...
+        >
+          <ModalTrigger>
+            <SelectionModal size="large">
+              <Modal.Header useTitle={false}>
+                <h3 className="text-lg mt-2 mb-3">Import templates</h3>
+                <p className="m-0 leading-6">Imported templates will be duplicated, meaning changes made to the originals won’t be reflected here. Any components will also be imported in the process.</p>
+              </Modal.Header>
+              <SelectionModal.Body>
+                <SelectionModal.Column isHighlight className="w-3/6 flex-shrink-0 px-4 pb-8">
+                  <SelectionModal.ColumnHeader className="mb-3">PROJECTS</SelectionModal.ColumnHeader>
```
altaudio, line 558: 💚 Should any of the classes we have added via the `className` prop be baked into the component?

They may not, but it's worth I thought! It would be cool if we just put the components together and they are styled without external jiggery - that said, this may not be sensible!

lib/SelectionModal/SelectionModalCategory.js
```diff
@@ -0,0 +1,56 @@
...
+  counter,
+  ...rest
+}) {
+  const buttonClasses = cx(
+    `bg-white relative border-solid flex w-full rounded outline-none px-3 py-2 border border-neutral-90 items-center ${className}`,
+    {
+      'text-input-border-focus': isSelected,
+      'hover:border-blue-primary': !isSelected
+    }
+  );
```
altaudio, line 19: 🚩 At this point it's probably worth going back to cheeky Peaky and seeing if this button type is going to be used elsewhere, and if so get it into GCDS and a new button type. It looks pretty useful to me!

altaudio (changes requested):
Looking good so far!

Only one 🚩 on whether the category button should be a new button type - one for check Peaky 🖼

stories/components/Modals.js
AmeeMorris: I went back and forth on this and landed on no because in the future we may want the columns a different size.

lib/SelectionModal/SelectionModalCategory.js
AmeeMorris: This is another weird one, this is as much a new button type as the links in the finder panel sidebar is. Basically I don't think this is a new button type, more its own slightly more complex new thing. 

If it ends up being needed elsewhere then it would be pretty trivial to split out when needed, then we can decide where it lives/what its called based on both usecases.

stories/components/Modals.js
altaudio: Great - sounds good!

lib/SelectionModal/SelectionModalCategory.js
altaudio: Sounds good - let's keep an eye out for future uses 👀

altaudio (approved):
Thanks for the responses, LGTM 🚀

PR #873, opened 2020-10-07
Add an InputConfirmationModal component

AmeeMorris:
### 💬 Description
This adds a new modal for making people type in a keyword to confirm they do definitely want to do a thing. 
### 📹 GIF (optional)
http://g.recordit.co/oStjH8rurV.gif

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook

### DIFF
--- a/lib/InputConfirmationModal/InputConfirmationModal.js
+++ b/lib/InputConfirmationModal/InputConfirmationModal.js
@@ -1,0 +1,99 @@
+import React, { useState } from 'react';
+import { node, string, bool, func, oneOfType } from 'prop-types';
+import {
+  Modal,
+  FormInput,
+  ButtonPrimaryDanger,
+  ButtonTertiary,
+  ButtonSecondary
+} from 'lib';
+
+function InputConfirmationModal({
+  onHide,
+  show,
+  introTitle,
+  confirmTitle,
+  introBody,
+  confirmBody,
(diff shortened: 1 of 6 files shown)


lib/InputConfirmationModal/InputConfirmationModal.js
```diff
@@ -0,0 +1,99 @@
...
+import { node, string, bool, func, oneOfType } from 'prop-types';
+import {
+  Modal,
+  FormInput,
+  ButtonPrimaryDanger,
+  ButtonTertiary,
+  ButtonSecondary
+} from 'lib';
+
+function InputConfirmationModal({
```
altaudio, line 11: 💚💚💚 Very subjective comment here, as this code is perf as is 👌

Although it is less verbose to have a single component that acts as both the intro and confirm page, I think it is potentially more difficult to read than if we had separate `<Intro />` and `<Confirm />` components.

The only reason I really pick up on this is that was always on a campaign for DRYer and less verbose code, at any cost. I think I have since realized that my mentality was wrong, and often more verbose and less DRY makes things much more readable!

So this comment is an attempt to counter a lot of the "this could by DRYer" comments that I left before 😂

altaudio (approved):

PR #876, opened 2020-10-13
Updates useDeferredList

altaudio:
### 💬 Description
Updates the `useDeferredList` hook so that the `deferredList` is updated if the `initialItemCount` argument is changed.

If the `initialItemCount` is updated to a value that is higher than the number of items currently shown, we update the number of items shown to match `initialItemCount`.

This is so we can handle changing `initialItemCount` values.

### DIFF
--- a/lib/useDeferredList/specs/useDeferredList.spec.js
+++ b/lib/useDeferredList/specs/useDeferredList.spec.js
@@ -20,10 +20,23 @@ describe('useDeferredList', () => {
     act(() => result.current.loadItems());
 
     expect(result.current.deferredList).toEqual(times(15, index => index));
+  });
 
-    act(() => result.current.loadItems());
+  it('should load more items if the initialItems argument is updated to a higher number than the current item count', () => {
+    let initialItemCount = 10;
 
-    expect(result.current.deferredList).toEqual(times(20, index => index));
+    const { result, rerender } = renderHook(
+      () => useDeferredList({ initialItemCount, list, numberOfItemsToLoad: 5 }),
+      {}
+    );
+
+    expect(result.current.deferredList).toEqual(times(10, index => index));
+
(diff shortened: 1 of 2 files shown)


lib/useDeferredList/specs/useDeferredList.spec.js
```diff
@@ -20,10 +20,23 @@ describe('useDeferredList', () => {
...
+      () => useDeferredList({ initialItemCount, list, numberOfItemsToLoad: 5 }),
+      {}
+    );
 
-    expect(result.current.deferredList).toEqual(times(20, index => index));
+    expect(result.current.deferredList).toEqual(times(10, index => index));
+
+    initialItemCount = 15;
+    rerender();
+    rerender();
```
altaudio, line 37: Why the double rerender here? I don't know, but the test doesn't work without it.

Once to update the `initialCount` value, but why the second time?

altaudio:
[Fix navigate to comment to work with waypoint fields](https://trello.com/c/tJMS82uU/643-fix-navigate-to-comment-to-work-with-waypoint-fields)

AmeeMorris (approved):

kyleharper (approved):

lib/useDeferredList/specs/useDeferredList.spec.js
kyleharper: Has a timeout or event been introduced maybe? 🤔

altaudio: > Has a timeout or event been introduced maybe? 🤔
No timeouts or events, purely setting state and useEffects!
> This is weird, what happens without it?
With just one `rerender` it doesn't perform the final state update, and the list remains at it's previous value (and the test fails)

I'll add this to a debt card in the release, as it does bug me!

PR #879, opened 2020-10-15
Removes check for if count equals list length. This was stopping us l…

altaudio:
…oading the final items.

### 💬 Description
_What the PR does with as much context as possible. All the changes in your PR should be described here._
### 🔗 Links
### 📹 GIF (optional)
### 🚪 Start Points
### 👫 Related PRs (optional)

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook

### DIFF
--- a/lib/useDeferredList/useDeferredList.js
+++ b/lib/useDeferredList/useDeferredList.js
@@ -8,10 +8,6 @@ function useDeferredList({
   const [itemCount, setItemCount] = useState(initialItemCount);
 
   const setItemCountWithinLimits = count => {
-    if (count === listLength) {
-      return;
-    }
-
     setItemCount(Math.min(listLength, count));
   };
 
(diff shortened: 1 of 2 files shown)


lib/useDeferredList/useDeferredList.js
```diff
@@ -8,10 +8,6 @@ function useDeferredList({
   const [itemCount, setItemCount] = useState(initialItemCount);
 
   const setItemCountWithinLimits = count => {
-    if (count === listLength) {
```
altaudio, line 11: This was stopping us updating our list to the final count value!

altaudio: Why two rerenders meant that it did update, I don't know! But we don't need it anymore!

PR #884, opened 2020-10-19
Add FileCard component and modules from GCDS

kyleharper:
### 💬  Description
In our recent mob, we expressed the need to align our language with GCDS to improve multiple aspects of our UI development. To do this we now have a `modules` DIR, which can be used to house all of the modules from GCDS.

In this PR we add the component `src/components/FileCard` which utilises the following;

- `src/modules/meta`
- `src/modules/controls`
- `src/modules/thumb`

The new component also uses the existing `Card` component but that didn't require any refactoring so it's largely unchanged. If you're interested in the component itself I recommend spinning up Storybook and playing around but for the most part if's nothing complex: https://share.getcloudapp.com/DOu9Bvpw

---

I'm going to be working with @p34ks to get the README filled out today so I won't merge this in until that is done 👍 

### ✅  Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook


src/components/README.md
```diff
@@ -0,0 +1,5 @@
+## What are components?
+
+- ?
+- ??
+- ???
```
kyleharper, line 5: @p34ks lets work together to flesh this out 👍

src/modules/README.md
```diff
@@ -0,0 +1,5 @@
+## What are modules?
+
+- ?
+- ??
+- ???
```
kyleharper, line 5: @p34ks lets work together to flesh this out 👍

src/hooks/useClassNames.js
```diff
@@ -0,0 +1,7 @@
+import cx from 'classnames';
+
+function useClassNames(baseClassName, props = {}, conditionalClassNames = {}) {
+  return cx(baseClassName, props.className, conditionalClassNames);
+}
+
+export { useClassNames };
```
kyleharper, line 7: 🔴  this shouldn't be a hook, it's just a helper function to make the common className prop extending easier vs all the bloat of add prop types and the same code everywhere.

I will move this into a different DIR.

src/components/files/fileCard/FileCard.js
```diff
@@ -0,0 +1,41 @@
...
+  meta: func.isRequired,
+  thumb: func.isRequired,
+  insetMeta: bool,
+};
+
+FileCard.defaultProps = {
+  insetMeta: false,
+}
+
+export { FileCard };
```
kyleharper, line 41: `FileCard` was tricky. Where do we draw the line on something being too complex with props and children and how do we detail standards from GCDS?

The final implementation here feels about right to me, it describes what we expect for `FileCard` and because it expects components you can compose whatever you want 👍

src/modules/controls/Controls.js
```diff
@@ -0,0 +1,22 @@
...
+  )
+}
+
+Controls.Control = Control;
+
+Controls.propTypes = {
+  children: node.isRequired,
+};
+
+export { Controls };
```
kyleharper, line 22: 🟢  by default controls are hidden until the group is hovered. I should change this to be based on a prop to animate the controls or even move the animation logic to the `FileCard` component. I'll give it a think.

src/modules/thumb/Thumb.js
```diff
@@ -0,0 +1,82 @@
+import React, { useEffect, useState } from "react";
+import { node, string } from "prop-types";
+import { useImage } from 'react-image'
+import { useClassNames } from "src/hooks/useClassNames";
+import Icon from "lib/Icon";
+import bolt from "../../../assets/icons/bolt.svg";
+
+function Thumb({ src, altText, title, fallback, children, ...props }) {
```
kyleharper, line 8: `Thumb` is pretty cool out the box it;

- handles the loading and failures of showing a thumbnail
- it'll wait a minimal amount of time before showing the image to reduce UI movement and weird whitespace UI issues that you can get with caching and things actually rendering in the browser
- For now the default is fallback pants but this will be updated very soon and in the consuming app it'll also be set anyways

styles/helpers/_text.scss
```diff
@@ -1,3 +1,9 @@
+.text-inherit {
```
kyleharper, line 1: This new helper is handy when I want to set standards on a container element e.g. `Meta.Heading` and I don't want to have to set the standards every time I want to use a `h2` or `h3` etc. See https://github.com/gathercontent/gather-ui/pull/884/files#diff-68ce1da9d41e727fc2bf0006559c0562d0a722bcd741b166f852187a751a8d26R36 for a example.

lib/ButtonNew/ButtonIcon/ButtonContainedIcon.js
```diff
@@ -0,0 +1,29 @@
...
+    'border border-solid border-neutral-90 bg-white',
+    'hover:bg-blue-90 hover:border-blue-primary'
+  ]);
+
+  return (
+    <ButtonIcon
+      onMouseOver={() => setActive(true)}
+      onFocus={() => setActive(true)}
+      onMouseOut={() => setActive(false)}
+      onBlur={() => setActive(false)}
```
AmeeMorris, line 18: 🚩 This feels weird, especially if its just for hover and focus styling. This component feels like instead of using ButtonIcon as a base you might have been better just using ButtonBase and building from the ground up. 

Happy to pair on this if im just missing some obvious issue that this solves!

lib/src/components/README.md
```diff
@@ -0,0 +1,5 @@
+## What are components?
+
+- ?
+- ??
+- ???
```
AmeeMorris, line 5: 😂

lib/src/modules/thumb/Thumb.js
```diff
@@ -0,0 +1,84 @@
+import React, { useEffect, useState } from 'react';
+import { node, string } from 'prop-types';
+import { useImage } from 'react-image';
```
AmeeMorris, line 3: oooh what does this give us?

lib/src/components/files/fileCard/FileCard.js
```diff
@@ -0,0 +1,44 @@
...
+
+  return (
+    <Card
+      {...cardProps}
+      className={cardClassNames}
+      innerClassNames="relative overflow-hidden rounded"
+    >
+      <Card.Content collapse>{thumb}</Card.Content>
+
+      <Card.Content className={metaContainerClassNames}>{meta}</Card.Content>
```
AmeeMorris, line 28: 💚 opinion time: i don't really like passing components as props. is there a benefit of doing that here?

AmeeMorris (changes requested):
Nice work on this!

One 🚩 and a few questions

lib/src/components/files/fileCard/FileCard.js
kyleharper: It might be easier to understand why you don't like passing components as props 😕 

Seeing as in reality there's little difference in passing a component vs a function I don't quite understand your opinion. Components are sugared implementations of `Object.create` so what's the issue?

kyleharper: Benefits are;

- You can communicate component UI requirements better.

- Components are functions so you have infinite scope.

- `children` is good but provides no guidance for intent making GCDS language almost impossible to declare in code alone. You're reliant on documentation doing all the hard lifting and that doesn't work.

- Using `children` alone ends up with the dev building consuming app components to provide better standards throughout their app. This ends up either following the pattern here (passing functions) or overpopulation of props which we know doesn't work (e.g. conversations).

lib/src/modules/thumb/Thumb.js
kyleharper: Our main benefit is not managing the logic for everything load-related with images. E.g. `useImage` with an src gives us a loading state which we can then use in the UI 👍 (includes failures)

lib/ButtonNew/ButtonIcon/ButtonContainedIcon.js
kyleharper: From GCDS icon buttons should have always had a "contained" pattern. 

This was missed out (probably due to speed and getting `ButtonIcon` out the door) and so without this abstraction, it becomes a task of refactoring the base component (maybe even the entire NEW button family) and testing our entire usage of said component. 🤦 

Now, I'm not sure why but we HAVE to pass `active` as a prop to get the icon blue 🤷 So this abstraction is following the patterns set by the original (and quite recent) base component `ButtonIcon`.

----

I'm happy to pair but we can't create a brand new component for this as that won't map to GCDS. 

Maybe for `ButtonIcon`, we should use a component stylesheet cause it's a mess of tailwind restricts I feel atm, for example, [this](https://share.getcloudapp.com/ApuL9bnN) isn't a nice implementation to work with vs using an `scss` file to create a clear direction for states was varients.

lib/src/helpers/createClassNames.js
```diff
@@ -0,0 +1,12 @@
+import cx from 'classnames';
+
+function createClassNames(
```
altaudio, line 3: 💚 I'm not 100% this new function is that helpful, isn't it essentially the same as using `cx`, but less flexible?

altaudio (commented):
Looks good so far! I've left comments on the code for the file card changes, and left the button icon stuff as I'm guessing that is WIP stuff yourself and Amee are working on!

Hit me up with the re-review whenever these are split and ready to go again 🙏

lib/src/modules/controls/Controls.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import { bool, node } from 'prop-types';
+import { createClassNames } from 'helpers/createClassNames';
+import { Control } from './Control';
+
+function Controls({ children, animateFromTop, ...props }) {
+  const classNames = createClassNames('controls relative flex', props, {
```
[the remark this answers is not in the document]
kyleharper: In this case, it's purely semantic. 

When reviewing the DOM in dev tools it helps to have human names for these things.

lib/src/modules/thumb/Thumb.js
```diff
@@ -0,0 +1,84 @@
...
+    useSuspense: false
+  });
+
+  const [
+    minimumLoadingTimeSurpassed,
+    setMinimumLoadingTimeSurpassed
+  ] = useState(true);
+  useEffect(() => {
+    setTimeout(() => setMinimumLoadingTimeSurpassed(true), 250);
+  }, []);
```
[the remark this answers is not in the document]
kyleharper: When using the module, it helps if the user has a consistent experience so;

- the user sees a loader (something must be coming)
- the user then sees image appear after it's loaded
- win 🥳 

Without this we get some instances where the image has "loaded" but it takes X time to render on the page which gives us strange whitespace which causes the UI to look odd.

After some testing especially in a collection of thumbs, this is even more apparent and this minimal wait time really helps to give us that consistent experience with loading images 😄

---

Saying all of that! This was implemented when the loaded was unmounted which isn't a thing SOOOOO it might not be needed. I'll re-test this and it's very possible that it isn't required anymore.

I'll update and resolve if I remove it 👍

lib/src/helpers/createClassNames.js
kyleharper: My thinking was that pretty much every component should allow us to extend className.

Every time I add className as a prop I need to manually add propTypes, so what's the difference?

- using classnames: https://gist.github.com/kyleharper/c2d96476fdb4c94822e29682b31ef2e1

- using createClassNames: https://gist.github.com/kyleharper/1dacb3fecc1d931e6cb0f91655037f02

It's nearly a 50% reduction and if you consider that there are 100's of exported components in this library well let's do some math;

- 9 * 100 = 900 lines 
- 900 * 2 = 1800 lines
- 900 * 3 = 2700 lines

It actually saves quite a bit 👍 

-----------

The only question really is, how can we make the implementation better. I'm down for pairing on this if it makes it easier now that I have hopefully explained what I am trying to do here. 

Maybe the name could make it clear what we're saving on;

- `extendClassNameFromProps()`
- `extendCommonUseOfClassNames()`
- `combineClassNamesFromProps()`
- `createComponentClassNames()` (I quite like this one)

lib/src/modules/thumb/Thumb.js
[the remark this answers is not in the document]
kyleharper: This has been removed, this implementation is a hack to try and get around the issue of us overusing inline SVGs especially in the case where said SVG is animated.

lib/src/helpers/createClassNames.js
altaudio: Thanks for the in-depth explanation, I see the advantage!

The "using classnames: https://gist.github.com/kyleharper/c2d96476fdb4c94822e29682b31ef2e1" example is misleading in how much code is removed, as this example doesn't need `propTypes`.

It's also misleading to say that the two examples do different things with and without `defaultProps` - in both cases we pass `props.className` as undefined to `cx`. If we defaulted `props.className` to an empty string however, then I think that would be an advantage! But at that point, it is the only advantage of the function, and `cx` doesn't seem to mind being passed `cx` anyway.

My thinking is that both the examples below pass the same thing to `cx`, what's the advantage of the new function?

```javascript
import { createClassNames } from 'src/...';

function Component(props) {
  const classNames = createClassNames('component', props, {});
  
  return (
    <div clasName={classNames} />
  );
};

export { Component };
```

```javascript
import cx from 'classnames';

function Component(props) {
  const classNames = cx('component', props.className, {});
  
  return (
    <div clasName={classNames} />
  );
};

export { Component };
```

kyleharper: Gone for `createComponentClassNames` and refined the names of the arguments too so it's easier to understand.

lib/ButtonNew/ButtonIcon/ButtonIcon.js
```diff
@@ -19,48 +19,28 @@ function ButtonIcon({
   iconTitle,
   children,
   enabled,
+  contained,
```
AmeeMorris, line 22: 🚩 Do we still need this? I can't see it being used in the component apart from proptypes 👁️

lib/ButtonNew/ButtonIcon/ButtonIconContained.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import { ButtonIcon } from 'lib';
+import { createComponentClassNames } from 'helpers/createComponentClassNames';
+import { buttonIconDefaultProps } from '../common';
+
+function ButtonIconContained(props) {
+  const classNames = createComponentClassNames('button-icon-contained', props);
+
+  return <ButtonIcon className={classNames} contained {...props} />;
```
AmeeMorris, line 9: 🚩 Same as above, it doesn't look like we need to pass that prop now

lib/src/helpers/createComponentClassNames.js
```diff
@@ -0,0 +1,17 @@
+import cx from 'classnames';
+
+function createComponentClassNames(
```
AmeeMorris, line 3: 💚  I tried to unresolve and leave a comment on https://github.com/gathercontent/gather-ui/pull/884#discussion_r509112386 but it wouldnt let me so re-resolved it 😓 

But I agree with @altaudio I'm not sure if I do see the benefit of this, I find the old way a bit clearer to digest (though admittedly that could just be due to familarity)

AmeeMorris (changes requested):
A few really minor clean up 🚩s and one juicy 💚

lib/src/helpers/createComponentClassNames.js
kyleharper: I'll revert to use `cx` and see if I can ignore `className` for propTypes.

kyleharper: I've removed this and added to our eslint config 👍  (thanks for challenging this)

AmeeMorris (approved):

PR #885, opened 2020-10-19
Add ButtonIcon for darker backgrounds

AmeeMorris:
### 💬 Description
A lovely new Button type, this time to work with darker backgrounds. 

I've also added pointer events none to disabled Buttons/DropdownActions, this is a work around for tooltips to work on disabled elements that I generally have to add in. 
### 📹 GIF (optional)
Screenshot 2020-10-19 at 17 20 12

### 🚪 Start Points
lib/ButtonNew/ButtonIcon/ButtonIconDark.js

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [x] Added to storybook

### DIFF
--- a/lib/ButtonNew/ButtonBase.js
+++ b/lib/ButtonNew/ButtonBase.js
@@ -26,7 +26,8 @@ function ButtonBase({
 
       'text-base py-3 px-4': size && size === sizes.md,
       'text-base p-2': size && size === sizes.sm,
-      'text-sm p-05 px-2 rounded-small': size && size === sizes.xs
+      'text-sm p-05 px-2 rounded-small': size && size === sizes.xs,
+      'pointer-events-none': disabled
     }
   );
 
--- a/lib/ButtonNew/ButtonIcon/ButtonIconDark.js
+++ b/lib/ButtonNew/ButtonIcon/ButtonIconDark.js
@@ -1,0 +1,71 @@
+import React from 'react';
+import cx from 'classnames';
+import { Icon } from 'lib';
+import { ButtonBase } from '../ButtonBase';
+import {
+  sizes,
+  getSizeClasses,
+  buttonIconPropTypes,
+  buttonIconDefaultProps
+} from '../common';
+
+function ButtonIconDark({
+  className,
+  disabled,
+  name,
+  active,
+  size,
+  iconTitle,
+  children,
+  ...rest
+}) {
+  const nonDisabledClasses = cx(
+    'bg-transparent border-transparent focus:bg-neutral-40 focus:border-white focus:shadow-button-icon-dark-focus',
+    {
+      'active:bg-neutral-40 active:border-neutral-40 active:shadow-none hover:bg-neutral-30 hover:border-neutral-30': !active,
+      'button__icon-active bg-neutral-40 border-neutral-40 hover:bg-neutral-30': active
+    }
+  );
+
+  const classes = cx(
+    className,
+    'button__icon',
+    'p-0',
+    'items-center justify-center',
+    {
+      [nonDisabledClasses]: !disabled,
+      'opacity-30 border-transparent bg-transparent': disabled,
+      'rounded-small': size === sizes.sm
+    },
+    getSizeClasses(size)
+  );
+
+  const iconTypes = [];
+
+  if (!disabled) {
+    iconTypes.push('white-on-button-parent-active');
+  }
+
+  if (!active) {
+    iconTypes.push('neutral-70');
+  }
+
+  if (active) {
+    iconTypes.push('white');
+  }
+
+  return (
+    <ButtonBase className={classes} disabled={disabled} size={false} {...rest}>
+      <Icon name={name} types={iconTypes} title={iconTitle} />
+      {children}
+    </ButtonBase>
+  );
+}
+
+ButtonIconDark.sizes = sizes;
+
+ButtonIconDark.propTypes = buttonIconPropTypes;
+
+ButtonIconDark.defaultProps = buttonIconDefaultProps;
+
+export { ButtonIconDark };
--- a/lib/ButtonNew/ButtonIcon/ButtonIconDark.js
+++ b/lib/ButtonNew/ButtonIcon/ButtonIconDark.js
@@ -56,7 +56,13 @@ function ButtonIconDark({
 
   return (
     <ButtonBase className={classes} disabled={disabled} size={false} {...rest}>
-      <Icon name={name} types={iconTypes} title={iconTitle} />
+      <Icon
+        name={name}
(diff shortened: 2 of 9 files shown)


lib/ButtonNew/ButtonBase.js
```diff
@@ -26,7 +26,8 @@ function ButtonBase({
 
       'text-base py-3 px-4': size && size === sizes.md,
       'text-base p-2': size && size === sizes.sm,
-      'text-sm p-05 px-2 rounded-small': size && size === sizes.xs
+      'text-sm p-05 px-2 rounded-small': size && size === sizes.xs,
+      'pointer-events-none': disabled
```
kyleharper, line 30: 🔴  When a button is disabled, isn't it disabled so why do we need to do this? 🤔

lib/ButtonNew/ButtonIcon/ButtonIconDark.js
```diff
@@ -0,0 +1,77 @@
...
+    'bg-transparent border-transparent focus:bg-neutral-40 focus:border-white focus:shadow-button-icon-dark-focus',
+    {
+      'active:bg-neutral-40 active:border-neutral-40 active:shadow-none hover:bg-neutral-30 hover:border-neutral-30': !active,
+      'button__icon-active bg-neutral-40 border-neutral-40 hover:bg-neutral-30': active
+    }
+  );
+
+  const classes = cx(
+    className,
+    'button__icon',
```
kyleharper, line 32: What does this do? 🤔  (can tailwind & BEM live together? I'm sure you just copied this though so maybe there isn't an answer)

lib/ButtonNew/ButtonIcon/ButtonIconDark.js
```diff
@@ -0,0 +1,77 @@
...
+  className,
+  disabled,
+  name,
+  active,
+  size,
+  iconTitle,
+  children,
+  ...rest
+}) {
+  const nonDisabledClasses = cx(
```
kyleharper, line 22: 🔴  I've mentioned this in #884 but I'm not digging how we're implementing tailwind in these scenarios.

After come back and trying to work with `ButtonIcon` it's almost impossible to understand what is what. This ends up with whoever makes future changes, either bolting on another prop or causing bugs because they forget something like the focus state.

Here are my main issues;

- The implementation isn't easy to digest for those who didn't write it.
- It's CSS which is the most readable programming syntax (for the most part) and this isn't so why is that?

Especially with how `classnames` works, you specify a string for classes and then the predicate after so I have to parse an entire long string before I can understand when it is applied.

And that's before prettier comes along and starts breaking things up making it even worse.

----

I believe you've suggested we pair on #884 which I'm down for.

lib/ButtonNew/ButtonIcon/ButtonIconDark.js
```diff
@@ -0,0 +1,77 @@
...
+  if (!disabled) {
+    iconTypes.push('white-on-button-parent-active');
+  }
+
+  if (!active) {
+    iconTypes.push('neutral-70');
+  }
+
+  if (active) {
+    iconTypes.push('white');
```
kyleharper, line 54: 🔴  I'd check GCDS but I believe we're missing a contained variant of the `ButtonIcon` here, which I've had to add in #884 which adds some hovering handlers to set the "active" state on hover.

How can we do this without the state setting I'm having to do in #884?

lib/ButtonNew/stories/ButtonIcon/ButtonIconDark.js
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import { ButtonIconDark } from '../../ButtonIcon/ButtonIconDark';
+import StoryItem from '../../../../stories/styleguide/StoryItem';
+import { ButtonStoryItem } from '../ButtonStoryItem';
+import { Counter } from '../../../Counter';
+
+storiesOf('Components', module).add('ButtonIconDark', () => (
```
kyleharper, line 8: 🟢  What do you think to have all the `ButtonIcon`s in one story?

kyleharper (changes requested):

lib/ButtonNew/ButtonBase.js
AmeeMorris: I mentioned it in the description, this needs to be here in order for tooltips to work on disabled buttons and something I have to add in quite often so makes sense for gui!

lib/ButtonNew/ButtonIcon/ButtonIconDark.js
AmeeMorris: Copied! 

Depending on what its doing I don't see a problem with this. I also quite like adding semantic/bem classes so I can see what is what when scanning the dom.

lib/ButtonNew/ButtonBase.js
kyleharper: The image captivated me and I missed the detail about the workaround 😄

lib/ButtonNew/ButtonIcon/ButtonIconDark.js
kyleharper: I agree with adding semantic classes but maybe not BEM just because it confuses what we use to do with how we're trying to do things now. I've been adding semantic classes in hyphen format e.g. `button-icon-dark` which as you say helps identify them in the browser 👍 

I imagine this is doing something, making the component even more difficult to work with. Your call if you wanna make `ButtonIcon` easier to work with.

lib/ButtonNew/ButtonIcon/ButtonIconDark.js
AmeeMorris: Let's pair! Perhaps we can bring this up in front end catch up today too, while I don't have a problem with this implementation, there is a card for making our button styles into tailwind components that might solve your problem.

lib/ButtonNew/ButtonIcon/ButtonIconDark.js
AmeeMorris: I would say if its on hover then its not the active state :D

lib/ButtonNew/stories/ButtonIcon/ButtonIconDark.js
AmeeMorris: Yeah this could be nice!

lib/ButtonNew/ButtonIcon/ButtonIconDark.js
kyleharper: Let's get the contained variant from GCDS in here and you'll see what I mean.

PR #888, opened 2020-10-23
Improve loader & add FileCard thumbnail example with improved fallback padding.

kyleharper:
### 💬  Description
The loader module can have progress and heading text passed through to it but the styling wasn't there. It was also missing a story to show the states of the new module.

loader

-------

The `FileCard` component uses the `PreviewImage` module which when used in a thumbnail scenario (e.g. in rows) had issues with padding and the fallback touching the edges. This PR adds padding which suits all scenarios. I also added a thumbnail example to the story.

thumbnail

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook

### DIFF
--- a/lib/src/components/files/fileCard/stories/FileCardStory.js
+++ b/lib/src/components/files/fileCard/stories/FileCardStory.js
@@ -25,13 +25,13 @@ stories.add('FileCard', () => {
     ? 'http://fail'
     : 'http://placeimg.com/500/500/animals';
 
-  const preview = (
+  const Preview = ({ children }) => (
     <PreviewImage
       src={previewSrc}
       altText="preview image"
       title="preview image"
     >
-      {controls}
+      {children}
     </PreviewImage>
   );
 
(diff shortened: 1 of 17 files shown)


lib/src/components/files/fileCard/stories/FileCardStory.js
```diff
@@ -59,15 +60,52 @@ stories.add('FileCard', () => {
...
+        </ul>
+      </StoryItem>
+
+      <StoryItem
+        title="FileCard (thumbnail usage)"
+        description="A card component which utilises the thumb, meta and control modules to display a file."
+      >
+        <FileCard
+          preview={<Preview loader={<Loader size="sm" />} />}
+          className="w-10"
```
altaudio, line 106: 💚 Do you feel we should have a `thumbnail` prop or something similar that sets this width, rather than doing it on the fly?

altaudio (approved):
👍

AmeeMorris (approved):

lib/src/components/files/fileCard/stories/FileCardStory.js
kyleharper: We could do, I'll take a look 👍

kyleharper: @altaudio I don't think having a prop for a "thumbnail" size isn't ideal for `FileCard`. 

We'll end up setting props to parents just for children as `Loader` and `PreviewImage` have a number of props individually, which leads us to smell patterns in the future e.g. `props.loaderHeadingText`, `props.loaderProgressText`, `props.loaderSize` 🤮 

It also doesn't help that the size of the image preview is set outside of the component so setting a thumbnail prop will mean we'll have some implementation that is internally managed and then some that are externally managed. Making is harder to digest.

PR #890, opened 2020-10-27
Move Controls animation to @apply styling

kyleharper:
### 💬 Description
Controls are used in `FileCard` components which are placed within `Field` components. In our consuming app fields have a `group` class applied, triggering lower-level components to animate on hover before they should.

This highlight an issue with using `group` as we are so I'm moving the controls animation styling to an `scss` file rather than using inline utility classes.

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook


lib/src/modules/controls/Control.js
```diff
@@ -1,9 +1,10 @@
 import React from 'react';
 import { func, string } from 'prop-types';
+import cx from 'classnames';
 import { ButtonIconContained } from 'lib';
 
 function Control({ onClick, iconName, children, className }) {
-  const classNames = `control flex mr-2 ${className}`;
```
kyleharper, line 6: Fixes `undefined` className being rendered in the DOM.

lib/src/modules/preview/previewImage.scss
```diff
@@ -6,4 +6,18 @@
 
 .preview-image-fallback {
   padding: 10%;
-}
\ No newline at end of file
+}
+
+.preview-children {
```
altaudio, line 11: 💚 One thing to note is that this change means that the only thing that can be passed as children to `<PreviewImage />` is now `<Controls />`, as the styles applied here are specific to controls.

This is not really a problem, but maybe it could be reflected in the API - instead of `<PreviewImage />` rendering  `children`, it could be passed a `controls` component prop. That way we're telling people "hey, this component expects controls and nothing else!"

altaudio (approved):
LGTM 👍 🚀

One comment on the change in the API that is worth a look!

lib/src/modules/preview/previewImage.scss
kyleharper: I understand what you're saying here but in truth, anything can be added here and it'll be positioned correctly.

PR #894, opened 2020-10-27
Card GCDS refactor

kyleharper:
### 💬  Description
The recently added `FileCard` component doesn't show some key states. States, like selected, highlighted, added and removed, are all currently handled in various implementations so I thought it would be helpful to refactor the card component to become the key component for these states.

It is then used by `FileCard` to inherit the states so we can show these states for files 👍 

-----

**The base card component**

card

-----

**Which is then used in FileCard**

card component

----

Here's a link showing the disabled state: https://share.getcloudapp.com/qGuvq7X9

### ✅  Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook


lib/GuideCard/index.js
```diff
@@ -1,13 +1,13 @@
...
-import Card from '../Card';
+import { Card } from '../src/components/card/Card';
 import Icon from '../Icon';
 import Body from './Body';
 import CTA from './CTA';
 
 const GuideCard = ({ children, iconName }) => (
   <Card className="guide-card__container">
-    <Card.Content paddingLarge className="guide-card__content">
+    <Card.Content className="guide-card__content">
```
kyleharper, line 10: `paddingLarge` doesn't do anything 🤷

lib/GuideCard/styles.scss
```diff
@@ -6,6 +6,7 @@
   display: flex;
   flex-direction: row;
   font-size: $typo-size-base;
+  color: $neutral-base;
```
kyleharper, line 9: This is the only thing I had to add to the guide card. `GuideCard` is not refactored in this PR.

lib/src/modules/meta/meta.scss
```diff
@@ -0,0 +1,3 @@
+.meta-text {
+  @apply truncate;
```
kyleharper, line 2: This is applied like this, so we can overwrite styles to allowing wrapping when required. `!important` and tailwind makes this tricky.

tailwind.config.js
```diff
@@ -58,6 +61,18 @@ module.exports = {
 
       'bottom-inset': 'inset 0px -4px 4px rgba(0, 0, 0, 0.06)',
       'top-inset': 'inset 0px 4px 4px rgba(0, 0, 0, 0.06)',
+
+      'yellow-primary-2px': `0px 0px 0px 2px ${yellowPrimary}`,
```
kyleharper, line 65: These new shadows have been very helpful for the cards states. Before we did a load of border re-jigging but now hovers for cards in specific states just use shadows 👍

lib/src/components/card/card.scss
```diff
@@ -0,0 +1,89 @@
...
+
+.card-disabled .card-content {
+  @supports (filter: blur()) {
+    @apply transition-all duration-200;
+    filter: blur(3px);
+  }
+}
+
+.card-controls {
+  @apply absolute top-0 p-2 -mt-10 transition-all duration-200;
```
kyleharper, line 77: The controls styling mainly fixes an issue with tailwinds `group` class clashing with other groups and animating before it should.

lib/src/components/files/fileCard/FileCard.js
```diff
@@ -1,44 +1,50 @@
...
     >
-      <Card.Content collapse>{preview}</Card.Content>
+      <Card.Content className="p-0">{preview}</Card.Content>
 
-      {meta && (
-        <Card.Content className={metaContainerClassNames}>{meta}</Card.Content>
-      )}
+      {meta && <Card.Content className="file-card-meta">{meta}</Card.Content>}
 
+      {controls && <Card.Controls>{controls}</Card.Controls>}
```
kyleharper, line 31: Some recent feedback from @altaudio was making it clearer about where "controls" live in the card / filecard namespace. This makes it easier to understand.

lib/src/modules/preview/PreviewImage.js
```diff
@@ -58,9 +59,7 @@ function PreviewImage({
           {imageHasFailed && fallback}
         </div>
 
-        <div className="absolute top-0 m-2">
-          <div className="preview-children">{children}</div>
-        </div>
+        {children}
```
kyleharper, line 62: controls are not named and handled in file cards. This is just good practice to render `children` still 👍

lib/src/modules/preview/previewImage.scss
```diff
@@ -7,17 +7,3 @@
 .preview-image-fallback {
   padding: 10%;
 }
-
-.preview-children {
```
kyleharper, line 11: Moved to file card styling.

lib/src/components/card/Card.js
```diff
@@ -0,0 +1,70 @@
...
+      onClick(e);
+    }
+  };
+
+  return (
+    <div // eslint-disable-line jsx-a11y/no-static-element-interactions
+      role={onClick ? 'button' : 'presentation'}
+      className={classNames}
+      onClick={onClick}
+      onKeyUp={handleEditKeyPress}
```
AmeeMorris, line 40: 🚩 A really soft red flag here just because this confused me!

What's this for? Just looking at card I don't know what "edit" is and why we have a keypress that fires onclick. Is it also possible that I wouldn't want enter to fire onClick?

lib/src/modules/meta/meta.scss
AmeeMorris: 💚 A sass file just for this is a little sad. Either add a prop that controls this classname within the component, or move everything over to here

tailwind.config.js
AmeeMorris: 💚 opinion time. I'm not a massive fan of using a shadow when it looks like and behaves like a border, we may want to add a shadow to this element later only to find out that we will be overriding the "border" by doing so. 

How bad was the juggling? David did a workaround using a pseudo element before 👀

AmeeMorris (changes requested):
Nice refactor! 

I've got a really minor 🚩 and a few 💚s dotted about. Feel free to ping me for a quick re-review.

lib/src/components/card/Card.js
kyleharper: ESLint requires it. Any "none" interactive element which takes an interactive role needs to handle keyUp events (which are only fired when you are focused on the element in the quest and press any key.

It's also worth noting that the original card component did this, I have just moved the card component into the new `src/components` DIR and refactored it to be functional 👍 

----

I will rename it though so it's just `handleKeyPress` again this was just copied so it kept the name 🤦

tailwind.config.js
kyleharper: The juggling was bad. We had fake transparent borders in places which then wouldn't work well with image content which needed to be flush and the implementation was much worse than this.

I don't feel the same way as you do though, shadows are just another tool which can be utilised to create a look. I recommend that once this is merged in you give it a go and see if you can get borders to function as it does here and then I can see how we could use borders instead 👍

lib/src/modules/meta/meta.scss
kyleharper: Let me see if I can still use `truncate` and when I pass `whitespace` whatever, it fixed the issue.

tailwind.config.js
kyleharper: I also tried using a pseudo-element and it looked horrible with clipping etc so if you wanna try that again you can, maybe I was missing something 👍  

(although I'd argue that your original comment about pretending to be a border is no different for a pseudo-element vs a shadow).

AmeeMorris: It's just a 💚  so no biggie. Its quite opinionated and if its saving a load of hassle then its no problem.

I would disagree with your argument though! The pseudo element is still using a border and not taking the place of any shadow.

AmeeMorris (approved):

PR #897, opened 2020-10-29
Update/add stucture editor layout

altaudio:
### 💬 Description
The component editor page of the app shares a lot in common with the template editor, as you can see from these designs https://www.figma.com/file/HWmdpuNLjhxlKUhzYQJWPW/Components?node-id=218%3A0

To make this as easy as possible, we want to make as much of the UI reusable as is sensible.

This PR creates the components:
```javascript
<StructureEditorLayout>
  <StructureEditorLayout.InventoryContainer>
  {...}
  </StructureEditorLayout.InventoryContainer>

  <StructureEditorLayout.FieldsContainer>
  {...}
  </StructureEditorLayout.FieldsContainer>  
</StructureEditorLayout>
```

you can see it in action in this PR https://github.com/gathercontent/app/pull/3457/files

This can now be reused for the component editor.

### DIFF
--- a/lib/src/components/layouts/StructureEditorLayout/InventoryContainer.js
+++ b/lib/src/components/layouts/StructureEditorLayout/InventoryContainer.js
@@ -1,0 +1,12 @@
+import React from 'react';
+import { Col } from 'react-bootstrap';
+
+function InventoryContainer({ children }) {
+  return (
+    <Col md={1} sm={2} className="structure-editor__inventory-container">
+      {children}
+    </Col>
+  );
+}
+
+export { InventoryContainer };
--- a/lib/index.js
+++ b/lib/index.js
@@ -161,3 +161,6 @@ export { Meta } from './src/modules/meta/Meta';
 export { PreviewImage } from './src/modules/preview/PreviewImage';
 export { Controls } from './src/modules/controls/Controls';
 export { Loader } from './src/modules/loader/Loader';
+export {
+  StructureEditorLayout
+} from './src/components/layouts/StructureEditorLayout/StructureEditorLayout';
--- a/lib/src/components/layouts/StructureEditorLayout/FieldsContainer.js
+++ b/lib/src/components/layouts/StructureEditorLayout/FieldsContainer.js
@@ -1,0 +1,12 @@
+import React from 'react';
+import { Col } from 'react-bootstrap';
+
+function FieldsContainer({ children }) {
+  return (
+    <Col md={9} xs={12} className="structure-editor__fields-container h-full">
+      {children}
+    </Col>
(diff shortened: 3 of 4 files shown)


lib/src/components/layouts/StructureEditorLayout/InventoryContainer.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import { Col } from 'react-bootstrap';
+
+function InventoryContainer({ children }) {
+  return (
+    <Col md={1} sm={2} className="structure-editor__inventory-container">
```
kyleharper, line 6: 🔴  We're moving away from BEM, if this is just semantic use "structure-editor-inventory-container" 👍

kyleharper (changes requested):
I've got one 🔴  and an overall question about these components;

GUI is meant to be a UI library that maps to our design system. At times we've found it beneficial to bring certain larger pieces of our Webapp into GUI to aid in a number of aspects of our dev process...although I'm tempted to delete the hierarchy one as it is out of date already and it was just an experiment.

My main question is - why do we need specific structure editor layout components in GUI?

Is this is GCDS and you're mapping it in here? Is it necessary for it to be GUI because it's composed in other components in GUI? Aren't layout components abstracted so can these new changes be named away from fields and inventory? What are those things in relation to layouts?

I don't believe GUI is the suitable location for this implementation in it's current state.

lib/src/components/layouts/StructureEditorLayout/InventoryContainer.js
altaudio: I didn't know we were moving away from BEM, happy to update this 👍

altaudio:
> I've got one 🔴 and an overall question about these components;
> 
> GUI is meant to be a UI library that maps to our design system. At times we've found it beneficial to bring certain larger pieces of our Webapp into GUI to aid in a number of aspects of our dev process...although I'm tempted to delete the hierarchy one as it is out of date already and it was just an experiment.
> 
> My main question is - why do we need specific structure editor layout components in GUI?
> 
> Is this from GCDS and you're mapping it in here? Is it necessary for it to be GUI because it's composed in other components in GUI? Aren't layout components abstracted so can these new changes be named away from fields and inventory?
> 
> Overall, I don't believe GUI is the suitable location for this implementation in its current state.

Both Amy and yourself have raised the same point and I agree with you both! I am going to move this into the App instead of GUI.

> Aren't layout components abstracted so can these new changes be named away from fields and inventory?
Normally yes, but the layout components for the fields and inventory are _very_ specific and cannot be made generic without a major refactor.

PR #902, opened 2020-11-05
Using react-dnd-cjs instead of the esm version

DavidHollins6:
### 💬 Description
This PR replaces any `react-dnd` packages with `react-dnd-cjs` packages. This is because NextJs does not play nicely with ESM modules, and neither does Jest!

When looking at their issues page, the solution they always give is to use the `-cjs` package instead:

https://github.com/react-dnd/react-dnd/issues/1414
https://github.com/react-dnd/react-dnd/issues/1514

Which is quite annoying but we gotta do it!

### 🚪 Start Points
`package.json`?

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook


altaudio (approved):
Changes look good!

Are we happy to use this package? I know there was some discussion yesterday about its suitability due to it not being updated regularly 🏚

kyleharper (approved):

PR #903, opened 2020-11-09
Add extra screen height utility classes

kyleharper:
### 💬  Description
Tailwind only provides us with `h-screen` but I want `h-screen70` or something like it so I've added some new ones which may be helpful.

---

As an added bonus I have also fixed a warning about `size` in button icons being `false` when it expects a string and also a missing spread operation for `ButtonIconContained ` in our `Controls.Control` module.

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook


tailwind.config.js
```diff
@@ -180,6 +180,36 @@ module.exports = {
...
+      '24': ' 6rem',
+      '30': ' 7.5rem',
+      '32': ' 8rem',
+      '40': ' 10rem',
+      '48': ' 12rem',
+      '56': ' 14rem',
+      '64': ' 16rem',
+      '128': ' 32rem',
+      auto: 'auto',
+      full: '100%',
```
AmeeMorris, line 205: 🚩  You should only need to include your added screen options here, rather than including options that we already get. We do it for maxHeight and maxWidth because there's not many options for those out of the box https://tailwindcss.com/docs/min-height#app 

As long as its in `extend` you're good just to add the missing ones 👍

lib/ButtonNew/common.js
```diff
@@ -25,7 +25,7 @@ export const propTypes = {
   type: string,
   loading: bool,
   loaderRight: bool,
-  size: string,
+  size: oneOfType([string, bool]),
```
AmeeMorris, line 28: 🙌

AmeeMorris (changes requested):
One very small 🚩

lib/src/modules/controls/Control.js
```diff
@@ -3,13 +3,18 @@ import { func, string } from 'prop-types';
...
 
   return (
     <div className={classNames}>
       {children || (
-        <ButtonIconContained name={iconName} onClick={onClick} size="sm" />
+        <ButtonIconContained
+          name={iconName}
+          onClick={onClick}
+          size="sm"
+          {...buttonProps}
```
altaudio, line 16: Nice naming, makes it clear why we are spreading 👍

altaudio (approved):
Looks good once Amee's comment is in 🙏

AmeeMorris (approved):

PR #908, opened 2020-11-12
Feature/inventory item

AmeeMorris:
### 💬 Description
In this we're making a lovely new InventoryItem component to replace some custom styling we have inside the app. It's quite a small simple component so there shouldn't be anything controversial! 

In order for us to have InventoryItem I've had to add a `lg` size to button (this is also reflected in gcds figma!), since InventoryItem is basically just a big ButtonIconContained with a few extra bits added on top.

There's also some new icons added!

### 🚪 Start Points
`lib/src/modules/inventoryItem/InventoryItem.js`
### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook


lib/ButtonNew/ButtonIcon/ButtonIconContained.js
```diff
@@ -1,10 +1,13 @@
 import React from 'react';
 import { ButtonIcon } from 'lib';
+import { sizes } from '../common';
 
-function ButtonIconContained(props) {
-  const classNames = `button-icon-contained ${props.className}`;
+function ButtonIconContained({ className, ...rest }) {
```
AmeeMorris, line 5: If we dont do this and pass a className to this component it overrides the className we set on the line below

lib/src/modules/inventoryItem/InventoryItem.js
```diff
@@ -0,0 +1,56 @@
+import React from 'react';
+import cx from 'classnames';
+import { bool, string } from 'prop-types';
+import { ButtonIconContained, TooltipWrapper } from 'lib';
+
+export function InventoryItem({
+  isDragging,
```
kyleharper, line 7: 🟢  Feels a little strange how we have the dragging props here. I believe it's just the name of them, is there a more abstracted naming convention we could use?

For example, `isDragPreview` could just be `preview` to say "this is the component when it's previewing the inventory" - I don't think we need to be specific about the interaction.

lib/src/modules/inventoryItem/InventoryItemStory.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import { boolean } from '@storybook/addon-knobs';
+import { InventoryItem } from 'lib';
+import StoryItem from 'stories/styleguide/StoryItem';
+
+storiesOf('Components', module).add('InventoryItem', () => {
+  const isDragging = boolean('isDragging', false);
```
kyleharper, line 8: 🟢  You could use the `<Draggable />` component here to make it interactive vs just using knobs.

lib/src/modules/inventoryItem/InventoryItem.js
```diff
@@ -0,0 +1,56 @@
...
+
+  if (isDragging) {
+    return <div className={classNames} />;
+  }
+
+  return (
+    <TooltipWrapper
+      id={`inventory-tooltip-${iconName}`}
+      tooltipText={tooltipText}
+      placement="right"
```
kyleharper, line 28: 🔴  What if I require a different placement?

lib/src/modules/inventoryItem/InventoryItem.js
```diff
@@ -0,0 +1,56 @@
...
+    'inventory-item-preview': isDragPreview
+  });
+
+  if (isDragging) {
+    return <div className={classNames} />;
+  }
+
+  return (
+    <TooltipWrapper
+      id={`inventory-tooltip-${iconName}`}
```
kyleharper, line 26: 🟢  Is there not a chance that we'll end up with 2 ids that are the same? Does this matter?

lib/src/modules/inventoryItem/InventoryItem.js
```diff
@@ -0,0 +1,56 @@
...
+    'inventory-item-dragging': isDragging,
+    'inventory-item-preview': isDragPreview
+  });
+
+  if (isDragging) {
+    return <div className={classNames} />;
+  }
+
+  return (
+    <TooltipWrapper
```
kyleharper, line 25: 🔴  Is there a reason why `TooltipWrapper` isn't composed outside of this?

This module could just focus on rendering an `InventoryItem` module, instead of worrying about every prop required for tooltips. Tooltips are so widely used (and easily composable) that I think in the long run it'd be much better if a lot of our component doesn't have props like `tooltipText`.

kyleharper (changes requested):

lib/src/modules/inventoryItem/InventoryItem.js
AmeeMorris: I get what you mean, but I think we should leave it in, everywhere we're going to use this will use a tooltip. However, you absolutely don't have to pass `tooltipText` so you can pretend it doesn't exist. I'm also pretty sure that if you don't pass `tooltipText` to a tooltip it'll just render its children and none of the tooltip stuff!

lib/src/modules/inventoryItem/InventoryItem.js
AmeeMorris: This is fair, but its a bit of a weird one that I went back and forth on when making this! While I know that this is a public repository, realistically we're the ones using it and those styles are used when those things are happening. And this is the name of what we pass down from a `Draggable` so it's not a completely alien concept to GUI.

lib/src/modules/inventoryItem/InventoryItemStory.js
AmeeMorris: Good idea!

lib/src/modules/inventoryItem/InventoryItem.js
AmeeMorris: I've changed `isDragPreview` to just `isPreview`, but I can't think of a vaguer name for the `isDragging` part! 

Maybe now we're using `Draggable` in the story that links the two together anymore so the prop name doesnt feel so out of place?

kyleharper (approved):

PR #917, opened 2020-12-07
Stopping the text fade from touching the shadow on tabs

DavidHollins6:
### 💬 Description
We have a shadow on tabs when you are able to scroll to indicate that there are more, the fade for a tab name clashes with it and makes things look a bit nasty. This PR just adds a z-index to the shadow so it is always about the fade, and not part of it.

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook

### DIFF
--- a/lib/TabsNew/TabsScrollShadow.js
+++ b/lib/TabsNew/TabsScrollShadow.js
@@ -23,7 +23,7 @@ export function TabsScrollShadow({ scrollPosition }) {
     opacity: shouldBottomShadowBeVisible ? 1 : 0
   });
 
-  const initialClassName = 'absolute w-full h-4 -mr-4 pointer-events-none';
+  const initialClassName = 'absolute w-full h-4 -mr-4 pointer-events-none z-50';
 
   return (
     <Fragment>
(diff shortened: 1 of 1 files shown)


DavidHollins6:
[A vein in Peaks' forehead will explode if the fade continues to mess with the shadow on tabs](https://trello.com/c/URPeCQGi/126-a-vein-in-peaks-forehead-will-explode-if-the-fade-continues-to-mess-with-the-shadow-on-tabs)

altaudio (approved):

kyleharper (approved):

lib/TabsNew/TabsScrollShadow.js
```diff
@@ -23,7 +23,7 @@ export function TabsScrollShadow({ scrollPosition }) {
     opacity: shouldBottomShadowBeVisible ? 1 : 0
   });
 
-  const initialClassName = 'absolute w-full h-4 -mr-4 pointer-events-none';
+  const initialClassName = 'absolute w-full h-4 -mr-4 pointer-events-none z-50';
```
AmeeMorris, line 26: 💚  is this the lowest the z-index can be? don't want the shadow going above something we aren't expecting. (tabs are usually fixed so shouldn't be an issue, I just find it good practice to go as low as poss for z-index)

AmeeMorris (approved):

lib/TabsNew/TabsScrollShadow.js
DavidHollins6: Makes sense!

PR #920, opened 2021-01-13
Feature/rename input

AmeeMorris:
### 💬 Description
This is our GCDS version of EditableTextWrapper, it looks and functions pretty much the same as renaming our new tab. Just like EditableTextWrapper it's some text you click on to turn into an input which you can then edit and hit enter to save. There's a story and a gif showing it off for details.

Unlike EditableTextWrapper but much like new tabs the input grows and shrinks to fit the content. This is mainly done with styling. Along with our input we have an element with either our input value or placeholder text, we hide this element but keep it in the flow to act as our sizer. As for the input we take it out of the flow with `absolute` and just set the width and height to 100%. 

### 📹 GIF (optional)
http://g.recordit.co/6Bz1DypKVo.gif
### 🚪 Start Points
lib/src/modules/renameInput/RenameInput.js

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook


lib/src/modules/renameInput/RenameInputForm.js
```diff
@@ -0,0 +1,60 @@
...
+  };
+
+  const handleEsc = e => {
+    const isEscKey = e.keyCode === 27;
+    const isEnterKey = e.keyCode === 13;
+    if (isEscKey) {
+      stopEditing();
+    }
+    if (isEnterKey) {
+      handleSubmit(e);
```
AmeeMorris, line 34: Even though we're in a form we still need to do this incase we're using a textarea

AmeeMorris:
[Update component label input to be inline with tabs](https://trello.com/c/R7xJsUS9/927-update-component-label-input-to-be-inline-with-tabs)

lib/src/modules/renameInput/RenameInputForm.js
altaudio: Do we need the `onSubmit` attribute on the form if it is handled here then?

lib/src/modules/renameInput/RenameInputForm.js
```diff
@@ -0,0 +1,62 @@
...
+      onRename(value);
+    }
+    stopEditing();
+  };
+
+  const handleOnChange = e => {
+    setValue(e.target.value);
+  };
+
+  const handleEsc = e => {
```
altaudio, line 28: 💚 I guess this could be `handleKeyDown`, as it checks for enter as well 👍

lib/src/modules/renameInput/RenameInputForm.js
```diff
@@ -0,0 +1,62 @@
...
+    const isEnterKey = e.keyCode === 13;
+    if (isEscKey) {
+      stopEditing();
+    }
+    if (isEnterKey) {
+      handleSubmit(e);
+    }
+  };
+
+  const InputType = multiline ? 'textarea' : 'input';
```
altaudio, line 39: This one is new to me! Who knew you could do this (obviously you did! 😉)

lib/src/modules/renameInput/RenameInputForm.js
```diff
@@ -0,0 +1,62 @@
...
+  return (
+    <form onSubmit={handleSubmit} className={`rename-input-form ${formClass}`}>
+      <label htmlFor={id} className="hidden">
+        {label}
+      </label>
+      <div className={`rename-input-size-helper ${classes}`} aria-hidden="true">
```
altaudio, line 45: This is a stroke of genius 🙏

altaudio (approved):
Great work! Definitely worth taking the consideration over this implementation, as it is super clear and perfectly accessible. Love the automatic width adjustment 🙇

lib/src/modules/renameInput/RenameInputForm.js
altaudio: 💚 My only concern with this PR is that without your comment, I might not have known that this div is the magic that resizes the parent. Maybe if it was extracted to a component like `<InvisibleParentSizer />` or something like that (probably better named)

altaudio (approved):

lib/src/modules/renameInput/RenameInputForm.js
kyleharper: As this is in a form we shouldn't need to do this.

kyleharper (changes requested):

lib/src/modules/renameInput/RenameInputForm.js
AmeeMorris: @kyleharper I didn't think we would either – it works great if its just a text input, but like I said it doesn't work if it is a textarea. If its a textarea it just adds new lines rather than submitting, something in textarea must override forms enter key handling!

PR #922, opened 2021-01-14
Add isOverOptions prop to Droppable component

altaudio:
### 💬 Description
Adds prop to `<Droppable />` so that we can pass options to the `monitor.isOver(...)` function.

This is so we can pass `{ shallow: true }` so that `isOver` is false if the dragged item is over a child of the target that is also droppable.


altaudio:
[Add `isOver` prop to `<Droppable />`](https://trello.com/c/xfDu6Doh/977-add-isover-prop-to-droppable)

AmeeMorris (approved):

PR #924, opened 2021-01-21
Creating the navigation module

DavidHollins6:
### 💬 Description
Welcome to my first module!!

This PR adds a new module to represent Navigation Tabs.

I have consulted some accessibility guidelines on web navigation when doing this to make sure we're all good. I'll paste some links below regarding that. There's some fancy aria & role stuff in here that seems important!

Styling-wise it's fairly bare bones, it serves my purpose in the uploader but will not do other things I imagine we would need from them (e.g. editable tab names). Those extra changes are for a conversation in the future. My opinion would be that this is strictly for navigation and nothing else though!

### 🔗 Links
https://www.w3.org/WAI/tutorials/menus/structure/
https://www.w3.org/TR/wai-aria-practices/examples/menubar/menubar-1/menubar-1.html#
https://www.figma.com/file/7uaL7O73bIuNhyBnfLs6lt/Upload-modal?node-id=1%3A4849

### 📹 GIF (optional)
https://share.getcloudapp.com/RBu9ggOb

### 🚪 Start Points
lib/src/modules/navigationItems/NavigationItemsStory.js

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook

### DIFF
--- a/lib/src/modules/navigationItems/NavigationItem.js
+++ b/lib/src/modules/navigationItems/NavigationItem.js
@@ -2,17 +2,22 @@ import React, { useContext } from 'react';
 import cx from 'classnames';
 import { NavigationContext } from './NavigationProvider';
 
-export function NavigationItem({ children, id, className }) {
+export function NavigationItem({ children, id, className, onClick, disabled }) {
   const { setActiveTabId, activeTabId } = useContext(NavigationContext);
   const isActive = activeTabId === id;
 
   const classNames = cx(className, 'navigation--item', {
-    active: isActive
+    active: isActive,
+    disabled
   });
   return (
     <li className={classNames} role="none">
       <button
-        onClick={() => setActiveTabId(id)}
+        onClick={() => {
+          onClick(id);
+          setActiveTabId(id);
+        }}
+        disabled={disabled}
         type="button"
         role="menuitem"
         tabIndex="0"
--- a/lib/src/modules/navigationItems/NavigationItem.js
+++ b/lib/src/modules/navigationItems/NavigationItem.js
@@ -1,0 +1,25 @@
+import React, { useContext } from 'react';
+import cx from 'classnames';
+import { NavigationContext } from './NavigationProvider';
+
+export function NavigationItem({ children, id, className }) {
+  const { setActiveTabId, activeTabId } = useContext(NavigationContext);
+  const isActive = activeTabId === id;
+
+  const classNames = cx(className, 'navigation--item', {
+    active: isActive
+  });
+  return (
+    <li className={classNames} role="none">
+      <button
+        onClick={() => setActiveTabId(id)}
+        type="button"
+        role="menuitem"
+        tabIndex="0"
+      >
+        {children}
+      </button>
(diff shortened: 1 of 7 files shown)


stories/index.js
```diff
@@ -88,4 +89,4 @@ import Sidebar from './components/Sidebar';
 import Layout from './components/Layout';
 import EditableChoiceInputStory from '../lib/EditableChoiceInput/stories/EditableChoiceInput';
 import ColField from '../lib/ColField/stories/ColFieldStory';
-import './experiments/virtualisedFields/dynamicHeightVirtualisedList'
+import './experiments/virtualisedFields/dynamicHeightVirtualisedList';
```
DavidHollins6, line 92: Thx prettier

lib/src/modules/navigationItems/NavigationItem.js
```diff
@@ -0,0 +1,25 @@
...
+  const { setActiveTabId, activeTabId } = useContext(NavigationContext);
+  const isActive = activeTabId === id;
+
+  const classNames = cx(className, 'navigation--item', {
+    active: isActive
+  });
+  return (
+    <li className={classNames} role="none">
+      <button
+        onClick={() => setActiveTabId(id)}
```
AmeeMorris, line 15: 🚩 Do we also want to pass our own `onClick` prop so we can control what it does? 😅  (unless i'm mistaking how we'll use this!)

AmeeMorris (changes requested):
One 🚩 which is hopefully minor or my mistake!

Also thanks for including all the accessibility links 🙏

lib/src/modules/navigationItems/NavigationItem.js
kyleharper: Good shout. 

I guess we could digest the active tab (from context) and `useEffect` on that but maybe that isn't clear? I'm easy.

lib/src/modules/navigationItems/NavigationItem.js
```diff
@@ -0,0 +1,25 @@
...
+        type="button"
+        role="menuitem"
+        tabIndex="0"
+      >
+        {children}
+      </button>
```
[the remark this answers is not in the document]
kyleharper: If it's not in GCDS maybe it shouldn't be. A `button` is just an interactive element.

lib/src/modules/navigationItems/NavigationItem.js
```diff
@@ -0,0 +1,25 @@
...
+export function NavigationItem({ children, id, className }) {
+  const { setActiveTabId, activeTabId } = useContext(NavigationContext);
+  const isActive = activeTabId === id;
+
+  const classNames = cx(className, 'navigation--item', {
+    active: isActive
+  });
+  return (
+    <li className={classNames} role="none">
+      <button
```
kyleharper, line 14: 🚩  is it a little strange how this is called `NavigationItem` but anchor tags aren't used?

kyleharper (changes requested):

PR #925, opened 2021-01-21
Escape Button Module

DavidHollins6:
### 💬 Description
This PR adds an escape button module. In an attempt to make it more plug and play, this module will slot itself at the top right corner of whatever element it's inside. I can't imagine this module being used in any other way! Other than that it's fairly simple.

### 📹 GIF (optional)
https://share.getcloudapp.com/JruqdP09

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook

### DIFF
--- a/lib/src/modules/escapeButton/EscapeButton.js
+++ b/lib/src/modules/escapeButton/EscapeButton.js
@@ -1,9 +1,9 @@
 import React from 'react';
 import { ButtonIcon } from 'lib';
 
-export function CancelButton({ onClick }) {
+export function EscapeButton({ onClick }) {
   return (
-    <span className="cancel-button">
+    <span className="escape-button">
       <ButtonIcon onClick={onClick} name="cross" />
     </span>
   );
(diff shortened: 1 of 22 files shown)


lib/src/modules/escapeButton/EscapeButton.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import { ButtonIcon } from 'lib';
+
+export function EscapeButton({ onClick }) {
+  return (
+    <span className="escape-button">
+      <ButtonIcon onClick={onClick} name="cross" />
+    </span>
+  );
+}
```
AmeeMorris, line 10: 💚 Could we add a cheeky bonus extra to this, and add an optional esc key handler 👀  the name might just be throwing me off with this, if thats not valid maybe 'CloseButton' might be better?

AmeeMorris (approved):
One little 💚  about a fun extra / renaming

kyleharper (approved):

PR #927, opened 2021-01-26
Modal Components

DavidHollins6:
### 💬 Description
This is the beginning of our modal components. At the moment they're very basic compound components. Most of the new code is the usage of a new modal library (`react-modal`). 

This PR focuses on ensuring the new modal looks/acts like our existing modals, so we shouldn't see any difference!

There is more to do around modals still, but this felt like a good time to get what I had into a PR.

Still to do:

- Flesh out the compound components
- Look at the props of the old modal and ensure we have that functionality built into the new ones.

### 🔗 Links
https://www.figma.com/file/66iSurBAx32DWrJymo9NKL/GCDS-%E2%80%93-Web-UI?node-id=6266%3A829
### 📹 GIF (optional)
https://share.getcloudapp.com/geu4xwwO
### 🚪 Start Points
`lib/src/components/modal/Modal.js` - Creation of the new modal
`lib/src/components/modal/modal.scss` - Styling the new modal

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook

### DIFF
--- a/lib/src/components/modal/modal.scss
+++ b/lib/src/components/modal/modal.scss
@@ -1,17 +1,30 @@
-.ReactModal__Overlay {
+.ReactModal__Content {
+  transform: translate(0, -25%);
+  opacity: 0;
+  transition: transform 0.3s ease-out, opacity 0.3s ease-out;
+}
+
+.ReactModal__Content--after-open {
+  opacity: 1;
+  transform: none;
+}
+
+.ReactModal__Content--before-close {
   opacity: 0;
   transform: translate(0, -25%);
-  transition: transform, opacity 0.3s ease-out;
+}
+
+.ReactModal__Overlay {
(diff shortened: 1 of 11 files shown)


lib/src/components/modal/modal.scss
```diff
@@ -0,0 +1,32 @@
+.ReactModal__Content {
```
kyleharper, line 1: How comes there are some fixed values in here? (bg colours etc)

kyleharper (changes requested):
One minor point but I wouldn't wanna see it remain in the long run so requesting changes.

lib/src/components/modal/modal.scss
DavidHollins6: Great spot, I was meant to convert these back to our hex variables

PR #930, opened 2021-01-27
Fixing modal styling and adding new props

DavidHollins6:
### 💬 Description
I have wanted to make the new Modal component has compatible with existing work that I could, so I have scoured app and found the most popular props we use and ensure they're usable in the new one. There ended up not being too many, with size being the most important one.

I also had a bit of a hassle with the theme colors and RGB mixins, so had to revert back to hardcoding (but with a comment so it should appear if we ever do a ctrl+f for the theme.

### 🚪 Start Points
`lib/src/components/modal/Modal.js`

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook

### DIFF
--- a/lib/src/components/modal/modal.scss
+++ b/lib/src/components/modal/modal.scss
@@ -1,30 +1,65 @@
+$modal-large-width: 780px !default;
+$modal-x-large-width: 900px !default;
+$modal-small-width: 400px !default;
+
 .ReactModal__Content {
   transform: translate(0, -25%);
-  opacity: 0;
-  transition: transform 0.3s ease-out, opacity 0.3s ease-out;
+  @apply opacity-0 p-0 transition-all ease-out duration-300;
 }
 
 .ReactModal__Content--after-open {
+  @apply opacity-100;
   opacity: 1;
   transform: none;
 }
 
 .ReactModal__Content--before-close {
-  opacity: 0;
+  @apply opacity-0;
   transform: translate(0, -25%);
 }
 
+.Modal {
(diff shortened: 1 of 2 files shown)


lib/src/components/modal/modal.scss
```diff
@@ -1,30 +1,65 @@
...
+
 .ReactModal__Overlay {
-  background: rgb(41,51,61, 0) !important; //Should be theme("colors.neutral.20") but it doesn't work for some reason and have spent too long on this!
-  transition: background 0.3s ease-out;
+  background: rgb(
+    41,
+    51,
+    61,
+    0
+  ) !important; //Should be theme("colors.neutral.20") but it doesn't work for some reason and have spent too long on this!
```
altaudio, line 43: Fair enough - nice idea with the comment 👍

altaudio (approved):
Nice work 🙏

lib/src/components/modal/modal.scss
```diff
@@ -1,30 +1,65 @@
...
 }
 
 .ReactModal__Overlay--after-open {
-  background: rgb(41,51,61, .2) !important; //Should be theme("colors.neutral.20") but it doesn't work for some reason and have spent too long on this!
+  background: rgb(
+    41,
+    51,
+    61,
+    0.2
+  ) !important; //Should be theme("colors.neutral.20") but it doesn't work for some reason and have spent too long on this!
```
AmeeMorris, line 53: Bizarre!

AmeeMorris (approved):

PR #933, opened 2021-02-01
ViewOptions Component

DavidHollins6:
### 💬 Description
A new component! ⚠️

This is the nice subheader that you'd see on a files & media/project library tab. It has a search input, a sorting dropdown, and a view switcher. There's also an option to add extra actions!

I had gone back and forth as to whether this should be split into more compound components, but I ended up feeling like it was too much of an over-abstraction. There aren't too many components here and it feels manageable.

### 🔗 Links
https://share.getcloudapp.com/jkueoBb4

### 🚪 Start Points
`lib/src/components/viewOptions/ViewOptions.js`

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook

### DIFF
--- a/lib/src/components/viewOptions/ViewOptions.js
+++ b/lib/src/components/viewOptions/ViewOptions.js
@@ -4,6 +4,11 @@ import { SearchInput } from '../../modules/searchInput/SearchInput';
 import Dropdown from '../../../Dropdown';
 import Icon from '../../../Icon';
 
+const viewTypes = {
+  grid: 'grid',
+  list: 'list'
+};
+
 export function ViewOptions({
   onChangeSearchHandler,
   onClearSearchHandler,
@@ -13,7 +18,7 @@ export function ViewOptions({
   sortingValues,
   actions
 }) {
-  const [view, setView] = useState('grid');
+  const [view, setView] = useState(viewTypes.grid);
   const [sort, setSort] = useState(sortingValues[0]);
 
   const onViewClick = value => {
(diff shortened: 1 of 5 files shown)


lib/src/components/viewOptions/ViewOptions.js
```diff
@@ -0,0 +1,84 @@
...
+export function ViewOptions({
+  onChangeSearchHandler,
+  onClearSearchHandler,
+  searchValue,
+  onViewChange,
+  onSortChange,
+  sortingValues,
+  actions
+}) {
+  const [view, setView] = useState('grid');
```
altaudio, line 16: 💚 It could be useful to attach the view types to the component, for use in the app
```javascript
ViewOptions.viewTypes = {
  grid: 'grid',
  list: 'list'
}
```
that way I don't have to open up the component to know what the view types are when I am implementing list and gird views in the app.

lib/src/components/viewOptions/ViewOptions.js
```diff
@@ -0,0 +1,84 @@
...
+          </Dropdown.Trigger>
+          <Dropdown.Content className="p-0">
+            {sortingValues.map(({ name, key }) => (
+              <Dropdown.Action
+                action={() => {
+                  onSortClick({ name, key });
+                }}
+              >
+                {name}
+              </Dropdown.Action>
```
altaudio, line 62: 💚 It's better for accessibility to render this content conditionally based on the `showContent` value that is passed to the children of `<Dropdown />` (checkout `gui/lib/Dropdown/DropdownProvider` for what I'm on about).

Otherwise this content is still rendered, just hidden.

lib/src/components/viewOptions/ViewOptions.js
```diff
@@ -0,0 +1,84 @@
...
+          active={view === 'grid'}
+        />
+        <ButtonIcon
+          size={ButtonIcon.sizes.sm}
+          name="list"
+          onClick={() => onViewClick('list')}
+          active={view === 'list'}
+        />
+      </div>
+      {actions && <div className="view-options__actions">{actions()}</div>}
```
altaudio, line 81: 💚 Do we need this? If not I'd be tempted to get rid of it, as it invites us to modify stuff app side rather than being strict with gui.

lib/src/components/viewOptions/ViewOptions.js
```diff
@@ -0,0 +1,84 @@
+import React, { useState } from 'react';
```
altaudio, line 1: 💚💚💚 Probably debatable, but I feel like the dir name should be capitalised to `ViewOptions/ViewOptions.js` - as it is the component name.

TBF I am happy with either, but maybe we should discuss in FE chat to get a standard. I'll create a card 🌄

altaudio (approved):
Love it! This is looking great!

I've left a few comments, but they are all green apples.

Nice work 🙏

lib/src/components/viewOptions/ViewOptions.js
altaudio: Here's an example https://github.com/gathercontent/app/pull/3722/files#diff-6aa60d91cb5544c121ca8c0cfe26d76ac51e9cb9975cf8f1cc018a3cce4cf24bR15-R33

lib/src/components/viewOptions/ViewOptions.js
kyleharper: DIRs shouldn't be capitalised (IMO). I've given a detailed reason in the past about why it's ideal not to capitalised folders but I'll have it dig it out.

PR #934, opened 2021-02-01
Creating a ModalFooterConfirm component

DavidHollins6:
### 💬 Description
A component for the ModalFooterConfirm modal! It's just a nice footer with some buttons and some text.

### 🔗 Links
### 📹 GIF (optional)
https://share.getcloudapp.com/NQuJzXBb

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook

### DIFF
--- a/lib/src/components/modal/Modal.js
+++ b/lib/src/components/modal/Modal.js
@@ -3,6 +3,7 @@ import ReactModal from 'react-modal';
 import { ModalFooter } from './ModalFooter';
 import { ModalHeader } from './ModalHeader';
 import { ModalBody } from './ModalBody';
+import { ModalFooterConfirm } from './ModalFooterConfirm';
 
 function Modal({ children, show, onHide, srContentLabel, size, keyboard }) {
   return (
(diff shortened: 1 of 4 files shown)


DavidHollins6:
[ModalFooterConfirm Component](https://trello.com/c/PZntc0g4/1019-modalfooterconfirm-component)

altaudio (changes requested):
This might be a wider question, but I was expecting the component to contain the primary and secondary buttons - not for them to be passed into the component at the app level.

To me a confirm footer has two buttons - "Confirm" and "Cancel" (not strictly that wording, but functionally).

If I can put any buttons in, what makes this a confirm footer?

I feel like we should be stricter on what these can do so we don't end up modifying these and losing the advantage of standardisation.

DavidHollins6:
> This might be a wider question, but I was expecting the component to contain the primary and secondary buttons - not for them to be passed into the component at the app level.
> 
> To me a confirm footer has two buttons - "Confirm" and "Cancel" (not strictly that wording, but functionally).
> 
> If I can put any buttons in, what makes this a confirm footer?
> 
> I feel like we should be stricter on what these can do so we don't end up modifying these and losing the advantage of standardization.
> 
> That said, everything else on here is looking really good so far 🚀

I agree with this. I think I over-engineered this yesterday, the buttons should definitely be in the component.

kyleharper (approved):

PR #935, opened 2021-02-01
Modal Header Tabs Component

DavidHollins6:
### 💬 Description
Here we create a new component to represent tabs in a modal header!
### 🔗 Links
https://share.getcloudapp.com/v1uNR8yb

### 🚪 Start Points
Top to bottom

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook

### DIFF
--- a/lib/src/components/modal/ModalHeaderTabs.js
+++ b/lib/src/components/modal/ModalHeaderTabs.js
@@ -1,0 +1,12 @@
+import React from 'react';
+import { CloseButton } from '../../modules/closeButton/CloseButton';
+
+export function ModalHeaderTabs({ children, headerText, onCloseClick }) {
+  return (
+    <div className="modal-header__tabs">
+      <div className="modal-header__tabs-text">{headerText}</div>
+      <div className="modal-header__tabs-tabs">{children}</div>
+      <CloseButton onClick={onCloseClick} />
+    </div>
+  );
+}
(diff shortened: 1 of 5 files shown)


lib/src/components/modal/ModalHeaderTabs.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import { CloseButton } from '../../modules/closeButton/CloseButton';
+
+export function ModalHeaderTabs({ children, headerText, onCloseClick }) {
+  return (
+    <div className="modal-header__tabs">
+      <div className="modal-header__tabs-text">{headerText}</div>
+      {children}
```
altaudio, line 8: 💚 I wonder if it's easier app side to include the `NavigationItem` code here rather than passing it in.

My concern is that doing it app side multiple times will result in variations in how it looks, and will take longer.

That said, it may not be sane to implement!

altaudio (approved):
Looks good, I've left one 💚 comment that may be worth checking out!

lib/src/components/modal/ModalHeaderTabs.js
kyleharper: Personally I like what `FileCard` does and set a prop called something like `navigationItem`. Take a look at that example and maybe it'll help.

kyleharper (approved):

PR #937, opened 2021-02-03
Add an option for the dropdown animation not to end in a transform

AmeeMorris:
### 💬 Description
This can cause an issue with nested dropdowns, so sometimes we want to ditch it

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook

### DIFF
--- a/lib/Dropdown/DropdownContent.js
+++ b/lib/Dropdown/DropdownContent.js
@@ -112,7 +112,8 @@ class DropdownContent extends Component {
       collapse,
       className,
       full,
-      noBorder
+      noBorder,
+      noTransform
     } = this.props;
 
     const { showContent } = this.context;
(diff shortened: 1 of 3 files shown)


altaudio (approved):
Code looks good! 🙏

I'm not 100% from the description or the prop name what the effects of toggling the prop will be - is there a more descriptive name in terms of what it does visually? Eg. `noEndBounce` (😆 I really don't know what it does so that suggestion is most likely crap!)

AmeeMorris:
> Code looks good! 🙏
> 
> I'm not 100% from the description or the prop name what the effects of toggling the prop will be - is there a more descriptive name in terms of what it does visually? Eg. `noEndBounce` (😆 I really don't know what it does so that suggestion is most likely crap!)

It still bounces and everything. There should be no visible difference, it just changes how fixed element child elements behave

PR #950, opened 2021-03-17
Form component

kyleharper:
### 💬 Description
As mentioned in #949 the form component is now here! As we have an existing `Form` component it's been exported as `TextForm` but in the near future we'll replace the existing `Form` with the new `Form` 😂 

It has some nice built in functionality like passing states to child proxy components to show loading and failure states 👍 

Here's a GIF showing the component: https://share.getcloudapp.com/wbu92y2v

### ✅  Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook


lib/src/components/form/FormHelper.js
```diff
@@ -0,0 +1,27 @@
...
+
+  if (hasFailed) {
+    hasFailedRef.current = true;
+  }
+
+  if (hasSucceeded) {
+    hasFailedRef.current = false;
+  }
+
+  return React.useMemo(
```
DavidHollins6, line 18: 🍏 

Would it be nicer to store `hasFailed` in the state, and have a `useEffect` triggering off a change in `status`. Which would then update the state? I suppose the actual question is, why does this have to be a ref?

lib/src/components/form/FormInput.js
```diff
@@ -0,0 +1,11 @@
+import * as React from 'react';
+import { Input } from 'lib';
+import { FormContext, statuses } from './Form';
+
+export function FormInput({ children, ...rest }) {
```
DavidHollins6, line 5: 🔴 

It would be nice to enforce a bit of consistency here. Every form input needs an `id` and a `name`. I hate to say this but maybe prop-types would help here to make sure that the input _always_ gets those.

DavidHollins6 (changes requested):

PR #953, opened 2021-03-22
Adding an innerclass prop to filecard to manipulate it a bit more

DavidHollins6:
### 💬 Description
That's it really. I'd like to be able to style the inner component of a `FileCard`. This is so I can make a nice loading filecard that you can see in the new story. 😎 

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook

### DIFF
--- a/lib/src/prefabs/files/fileCard/FileCard.js
+++ b/lib/src/prefabs/files/fileCard/FileCard.js
@@ -10,6 +10,7 @@ function FileCard({
   insetMeta,
   children,
   size,
+  innerClassName,
   ...cardProps
 }) {
   const cardClassNames = cx('file-card group', cardProps.className, {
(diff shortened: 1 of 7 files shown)


lib/src/prefabs/files/fileCard/FileCard.js
```diff
@@ -10,6 +10,7 @@ function FileCard({
   insetMeta,
   children,
   size,
+  innerClassName,
   ...cardProps
```
kyleharper, line 14: I'm not too sure if we need to add this prop for `fileCard`.

`...cardProps` is collected in this pattern because the `Card` component has the `innerClassName` prop. File card is just built on top so do we need to add this in?

If you're passing inner class name then it'l be spread to the card component too.

------

This is similar to other areas that do this like proxy components like `Input` and `Buttons`.

kyleharper (changes requested):
The spread logic here does this for you (see my feedback).

kyleharper (approved):

PR #957, opened 2021-03-26
Prevent ShortCutTrigger from triggering in text fields

DavidHollins6:
### 💬 Description
The title says it all. We check the event to see if it was from a text input, if it is we do nothing.

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook

### DIFF
--- a/lib/ShortcutTrigger/index.js
+++ b/lib/ShortcutTrigger/index.js
@@ -41,6 +41,10 @@ class ShortcutTrigger extends Component {
   }
 
   onKeyDown(event) {
+    if (event.target.type && event.target.type === 'text') {
+      return event;
+    }
+
     return this.shortcutHasBeenExecuted(event) && !event.repeat
       ? this.props.onShortcutTrigger(event)
       : event;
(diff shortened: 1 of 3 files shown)


lib/ShortcutTrigger/index.js
```diff
@@ -41,6 +41,10 @@ class ShortcutTrigger extends Component {
   }
 
   onKeyDown(event) {
+    if (event.target.type && event.target.type === 'text') {
```
altaudio, line 44: 💚
```suggestion
    if (event.target.type  === 'text') {
```
Can we just do this? If `event.target.type` is undefined, the equality check will still fail.

altaudio (approved):
Looks good 🚀 - one 💚 comment!

lib/ShortcutTrigger/index.js
DavidHollins6: Great point, that's merged!

PR #958, opened 2021-04-07
add isHovered prop to ColField and ComponentWrapper

AmeeMorris:
### 💬 Description
When hovering over repeat buttons for fields/components we needed a 1px blue border style, enter our new prop `isHovered`. We already had a `hovered` prop for some `ComponentWrapper` components, but as we only use that for dnd i've renamed it to `draggedOver`, having a `isHovered` and `hovered` prop felt a bit icky!

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook

### DIFF
--- a/lib/ColField/ColField.js
+++ b/lib/ColField/ColField.js
@@ -33,21 +33,20 @@ function ColField({
     config: { duration: 100 }
   });
 
-  const classes = cx(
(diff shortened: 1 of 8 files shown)


altaudio (approved):
👍  LGTM

PR #968, opened 2021-05-05
Update baseline Card component with styles

kyleharper:
### 💬  Description

Card components have a tried a tested abstraction around `FileCard` and we're starting to move the baseline `Card` component into a better state.

This PR adds some styling elements to the components whilst adding a few new elements to the component (e.g. `Card.Title`).

This PR also;

- removed `GuideCard` as it's deprecated
- improves the story navigation around cards (we know have a "cards" story nav item)

Screenshot 2021-05-05 at 08 03 13

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook


lib/src/components/card/card.scss
```diff
@@ -13,7 +13,19 @@
...
+.card-title {
+  @apply text-lg font-semi-bold mb-4;
+}
+
+.card-description {
+  @apply text-neutral-primary leading-6 mb-4;
+}
+
+.card-footer {
+  @apply flex justify-end;
```
petemcfarlane, line 28: Ahh cool, this now makes sense to me, thanks 👍

petemcfarlane (approved):
Nice one. Is it also linked to something in Figma? Might be useful to add a link to the PR do you think? Or is it best not as it's a private repo? Would just make it easier to compare/review is all.

PR #976, opened 2021-05-14
Refactor & fix race condition for auto positioned dropdown content

AmeeMorris:
### 💬 Description
When conditionally rendering the dropdown content a race condition happens when auto positioning the dropdown content. It tries to grab the element and its bounds before its rendered, stopping auto positioning from working and keeping the dropdown off screen. 

I've refactored the component and tweaked when we calculate styles to stop this from happening. 
### 🚪 Start Points
`lib/Dropdown/DropdownContent.js`

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook

### DIFF
--- a/lib/ConfirmationDropdown/ConfirmationDropdownContent.js
+++ b/lib/ConfirmationDropdown/ConfirmationDropdownContent.js
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import Button from '../Button';
 import Icon from '../Icon';
 import TooltipWrapper from '../TooltipWrapper';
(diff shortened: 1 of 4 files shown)


AmeeMorris:
[SPIKE: Dropdowns within dropdowns, how would it work? (0.5)](https://trello.com/c/JJFzWfOf/1441-spike-dropdowns-within-dropdowns-how-would-it-work-05)

DavidHollins6 (approved):

PR #977, opened 2021-05-17
Export ButtonSecondary styles to its own class

AmeeMorris:
### 💬 Description
Another one moved over!

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook

### DIFF
--- a/lib/ButtonNew/styles/buttonBase.scss
+++ b/lib/ButtonNew/styles/buttonBase.scss
@@ -1,5 +1,5 @@
 .button-base {
-  @apply border border-solid leading-4 rounded font-display font-semi-bold inline-flex items-center relative;
+  @apply border border-solid leading-4 rounded font-display font-semi-bold inline-flex items-center relative no-underline;
 
   &-md {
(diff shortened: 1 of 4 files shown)


petemcfarlane (approved):
Ah this is so much nicer to understand, thanks for walking through this with me!

lib/ButtonNew/styles/buttonBase.scss
```diff
@@ -1,5 +1,5 @@
 .button-base {
-  @apply border border-solid leading-4 rounded font-display font-semi-bold inline-flex items-center relative;
+  @apply border border-solid leading-4 rounded font-display font-semi-bold inline-flex items-center relative no-underline;
```
petemcfarlane, line 2: For context to others, when we were adding `button-base` styling to an `<a>` tag, it was still inheriting the browsers `text-decoration: underline.` This resolves that issue. Can't think of a case where we'd want a button to also be underlined.

PR #978, opened 2021-05-17
Pass className prop to modal-header-inner in ModalHeader

petemcfarlane:
### 💬 Description
Can't override the classes/styling provided in the `ModalHeader` because the className prop is just sent to Modal.Header, not the modal-header-inner.

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook

### DIFF
--- a/lib/Modal/ModalHeader.js
+++ b/lib/Modal/ModalHeader.js
@@ -2,9 +2,9 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import Modal from 'react-bootstrap/lib/Modal';
 
-const ModalHeader = ({ children, text, useTitle, ...props }) => (
+const ModalHeader = ({ children, text, useTitle, className, ...props }) => (
   <Modal.Header {...props}>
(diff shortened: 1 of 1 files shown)


kyleharper (approved):

AmeeMorris (approved):
Just make sure we don't rely on that class being there for anything else!

DavidHollins6 (approved):

petemcfarlane:
Good call on checking @AmeeMorris , there's two modals (delete file confirmation modal) which do remove the border from modal-header (and footer) so this would be a breaking change there. 
Screenshot 2021-05-20 at 09 44 22

I suppose perhaps since it's the only place we don't have borders on the modal then that could be considered the odd one out? 
Screenshot 2021-05-20 at 09 39 01

Seeing as a found another solution and don't need this to work anyway I think I'll close this for now, unless others feel strongly that we should be doing this? Feel free to reopen.

PR #981, opened 2021-05-20
Creating WorkflowStep component

DavidHollins6:
### 💬 Description
Here we have the `WorkflowStep` compound component! It handles layout AND animations. There is some interesting code in here, it was all to aid animations. It turns out animating on width/height is more difficult than it should be.

### 🔗 Links
https://www.figma.com/file/XETErklTPSe6g39mrInB2D/Workflow-2021?node-id=54%3A8290

### 📹 GIF (optional)
https://share.getcloudapp.com/p9ub0D1o

### 🚪 Start Points
`lib/src/components/workflowStep/WorkflowStepStory.js` - The story explains it's usage. 

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook


lib/src/components/workflowStep/workflowStep.js
```diff
@@ -0,0 +1,75 @@
...
+        setShowActions(false);
+      }}
+      onKeyUp={handleKeyUp}
+      onClick={onStepClick}
+      className={workflowStepClassName}
+      tabIndex={!isActive && !showBody ? 0 : -1}
+      role="menuitem"
+    >
+      {children}
+    </div>
```
DavidHollins6, line 66: Something felt wrong about making this a button, so I have made it a `div` with the `menuItem` role, since a workflow step feels like an item in the workflow menu.

We use the hover event to trigger whether we show actions, as well as focus for keyboard only users!

lib/src/components/workflowStep/workflowStepMeta.js
```diff
@@ -0,0 +1,29 @@
...
+    maxWidth: showActions ? maxWidth : '0px',
+    opacity: showActions ? 1 : 0,
+    config: { duration: showActions ? 100 : 300, easing: easings.easeCubic }
+  });
+
+  return (
+    <div className="workflow-step__meta-container">
+      <div className="invisible workflow-step__actions" ref={hiddenActionsRef}>
+        {actions}
+      </div>
```
DavidHollins6, line 24: To get the amount of width we need to animate, we need to have an invisible version of the div that is always at full width. We can then reference it to get the `offsetWidth`!

styles/helpers/_borders.scss
```diff
@@ -7,6 +7,14 @@
...
 }
 
+.pseudo-left-border-blue-2px {
+  @include before-left-border(
+    2px,
+    solid,
+    theme('colors.blue.primary')
+  );
+}
+
```
DavidHollins6, line 17: There are times (not including here) where we just want a left border of that primary blue. So I added in the helper for it!

DavidHollins6:
[Create `WorkflowStep` compound component (1)](https://trello.com/c/Oz6CUJMX/1448-create-workflowstep-compound-component-1)

AmeeMorris (approved):

PR #1004, opened 2021-06-29
Various fixes to the responsiveness of workflow dropdown

DavidHollins6:
### 💬 Description
A user reported an issue when using the new workflow dropdown on a smaller resolution. Peaks ended up giving some feedback on how we can improve the responsiveness, here are those changes!

This PR is mostly style updates. As well as a change to our collapsible logic to handle the changing of sizes better.

### 🔗 Links
https://trello.com/c/T6gWuxPx/1588-regular-the-new-workflow-ui-makes-it-difficult-to-leave-longer-notes

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook

### DIFF
--- a/lib/src/components/workflowStep/workflowStepCollapsibleBody.js
+++ b/lib/src/components/workflowStep/workflowStepCollapsibleBody.js
@@ -2,21 +2,23 @@ import React, { useContext, useEffect, useRef, useState } from 'react';
 import { useTransition, animated } from 'react-spring';
 import * as easings from 'd3-ease';
 import { WorkflowStepContext } from './workflowStepProvider';
+import { useSize } from '../../../helpers/useSize';
 
 export function WorkflowStepCollapsibleBody({ children }) {
   const { showBody } = useContext(WorkflowStepContext);
   const bodyRef = useRef(null);
-  const [shouldAllowOverflow, setShouldAllowOverflow] = useState(false);
+  const size = useSize(bodyRef);
+  const [bodyIsFullyOpen, setBodyIsFullyOpen] = useState(false);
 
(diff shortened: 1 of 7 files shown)


lib/src/components/workflowStep/workflowStepCollapsibleBody.js
```diff
@@ -2,21 +2,23 @@ import React, { useContext, useEffect, useRef, useState } from 'react';
 import { useTransition, animated } from 'react-spring';
 import * as easings from 'd3-ease';
 import { WorkflowStepContext } from './workflowStepProvider';
+import { useSize } from '../../../helpers/useSize';
 
 export function WorkflowStepCollapsibleBody({ children }) {
   const { showBody } = useContext(WorkflowStepContext);
   const bodyRef = useRef(null);
-  const [shouldAllowOverflow, setShouldAllowOverflow] = useState(false);
+  const size = useSize(bodyRef);
```
DavidHollins6, line 10: We need to have the correct up-to-date value on the height of our collapsible body. This is a hook I found on the internets that gives us that.

I tried writing one myself and it turns out it's not that simple!

AmeeMorris (approved):
I feel like there’s probably a simpler way of doing this, but i suppose animations need a fixed value to work that forces our hand a bit?

DavidHollins6:
> I feel like there’s probably a simpler way of doing this, but i suppose animations need a fixed value to work that forces our hand a bit?

Yee we need the exact values to know how much to animate. I have tried estimating but it makes the animations too long/short and it looks baaaad.

PR #1005, opened 2021-06-30
Creating a PersonSearch component

DavidHollins6:
### 💬 Description
This is a component used for searching people. We already have one called `UserList`, but it is:

- a class component (and a scary one at that)
- not using the new styles
- a big ol component for a list of people and an input

So I have decided to make my own! 

### 📹 GIF (optional)
https://share.getcloudapp.com/9ZudzNmE
### 🚪 Start Points
`lib/src/components/personSearch/PersonSearch.js` <- the good stuff

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook


lib/src/components/personSearch/PersonSearch.js
```diff
@@ -0,0 +1,75 @@
...
+                    onBlur={() => toggleShowContent(true)}
+                    onChange={e => setSearchValue(e.target.value)}
+                    value={searchValue}
+                  />
+                </>
+              )}
+            </Dropdown.Trigger>
+            {showContent && (
+              <Dropdown.Content noTransform collapse className="w-full">
+                <DropdownContent>
```
AmeeMorris, line 36: 💚 This could be split up a bit. There’s a few nested ternarys here that makes it hard to parse. 

Maybe something like `PersonSearchContent` `PersonSearchNoResults` `PersonSearchResults`?

AmeeMorris (approved):
A greeny for your consideration!

PR #1024, opened 2021-07-19
Replace DropdownSwitcher with Dropdown

AmeeMorris:
### 💬 Description
This is making a start removing bootstrap Dropdown from GUI - DropdownSwitcher used bootstrap Dropdown, but DropdownSwitcher was barely used. I’ve refactored this out and replaced usages with our own Dropdown component.

I’ve had to add some styling to MenuItem in this as well, since bootstrap would apply some styling when MenuItem was used inside Dropdown. 
### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook


AmeeMorris:
[Replace bootstrap `Dropdown` with own version](https://trello.com/c/Piii3xpk/1677-replace-bootstrap-dropdown-with-own-version)

lib/src/modules/menuItem/menu-item.scss
```diff
@@ -0,0 +1,29 @@
...
+  @apply text-neutral-20 text-base py-3 px-4 block;
+  text-decoration: none;
+  white-space: nowrap;
+  &:hover, &:focus {
+    @apply text-neutral-20 bg-neutral-98;
+  }
+}
+
+.dropdown__content .menu-item.disabled {
+  opacity: 0.5;
```
kyleharper, line 22: 🟢  a very minor thing but what is a menu item isn't in a dropdown?

Should it not use these styles?

kyleharper (approved):
Just 1 🟢  but 👍

lib/src/modules/menuItem/menu-item.scss
AmeeMorris: Ah we use it for navigation links mainly - its just sometimes used in dropdowns

PR #1028, opened 2021-07-20
Creating new colours for Pills

DavidHollins6:
### 💬 Description
There feels like a bit of debt accruing here. I would love to just pass in a bg colour and text colour to the pill component, but instead, I need to create whole new types each time I want a new combination of those colours. I'll add something to the frontend board for it.

Bonus added features:
- I added a new icon
- I exported the status indicator circle

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook


DavidHollins6:
[Whats going on with the `Current Revision` button?](https://trello.com/c/rWN0yyz3/1706-whats-going-on-with-the-current-revision-button)

styles/helpers/_borders.scss
```diff
@@ -7,12 +7,8 @@
   );
 }
 
-.pseudo-left-border-blue-2px {
-  @include before-left-border(
-    2px,
-    solid,
-    theme('colors.blue.primary')
-  );
+.pseudo-border-blue-2px-no-border-radius-on-focus {
```
DavidHollins6, line 10: I also need this for the revisions view. This pattern is starting to get a bit smelly.

AmeeMorris (approved):
> There feels like a bit of debt accruing here. I would love to just pass in a bg colour and text colour to the pill component

I kind of like that all our Pill types (like our buttons, notification types, etc) are defined in GUI and we don't have a bunch of random ones in the app. If this is becoming a problem maybe we just need less pill types?

That being said and we feel we want to be more loosey goosey with pill types, couldn't we just export `PillBase` and just pass background and text colour classes to it?

kyleharper (approved):
100% on the debt POV!

PR #1031, opened 2021-07-26
Adding a shortcut trigger for submitting an input with mentions

DavidHollins6:
### 💬 Description
The title says it all. We would like to submit the `InputWIthMentions` with cmd+enter.

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook

### DIFF
--- a/lib/src/modules/input/InputWithMentions.js
+++ b/lib/src/modules/input/InputWithMentions.js
@@ -1,19 +1,21 @@
 import React, { useState, useRef, useEffect } from 'react';
 import { Mention, MentionsInput } from 'react-mentions';
 import cx from 'classnames';
-import { Avatar, AvatarInformation } from 'lib';
+import { Avatar, AvatarInformation, ShortcutTrigger } from 'lib';
 
 export function InputWithMentions({
   value,
   users,
   onChange,
   onMention,
+  onSubmitShortcutTriggered,
   ...rest
 }) {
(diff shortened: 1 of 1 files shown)


DavidHollins6:
[CMD+ENTER should submit the workflow note form.](https://trello.com/c/5MgDp1gh/1719-cmdenter-should-submit-the-workflow-note-form)

lib/src/modules/input/InputWithMentions.js
```diff
@@ -94,6 +98,28 @@ export function InputWithMentions({
         />
       </MentionsInput>
       <div ref={mentionsPortalRef} />
+      <ShortcutTrigger
```
kyleharper, line 101: 🔴   why do we need 2 `ShortcutTrigger`s? 

Can't we just set `withCtrlKey` & `withMetaKey` on a single component? 🤔

kyleharper (changes requested):

lib/src/modules/input/InputWithMentions.js
DavidHollins6: That would require the user to hold both `ctrl` and `meta` and `Enter` to trigger the shortcut.

kyleharper: haha oh yea! 😂

kyleharper (approved):

PR #1035, opened 2021-07-27
Fix DropdownMenu not closing after clicking on item or clicking outside of dropdown

AmeeMorris:
### 💬 Description
`useEffect` does some caching and because event listener is only set up on mount while `showItems` is false, it then always thinks its false every time that event is fired. 

Changing it so the event listener gets set up when `showItems` is true solves the problem, and as a nice side effect we're not needlessly setting up event listeners before someones even interacted with the dropdown

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook

### DIFF
--- a/lib/DropdownMenu/index.js
+++ b/lib/DropdownMenu/index.js
@@ -1,5 +1,5 @@
 /* eslint-disable react/no-array-index-key */
-import React, { useState, useEffect } from 'react';
+import React, { useState, useEffect, useRef } from 'react';
 import cx from 'classnames';
 import Button from '../Button';
 import Icon from '../Icon';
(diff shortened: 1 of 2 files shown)


AmeeMorris:
[[REGULAR] The change template drop-down doesn't collapse when selecting a new template in the item editor](https://trello.com/c/K0IdwU5l/1740-regular-the-change-template-drop-down-doesnt-collapse-when-selecting-a-new-template-in-the-item-editor)

DavidHollins6 (approved):

kyleharper (approved):

PR #1046, opened 2021-09-01
Slight modal styling changes

kyleharper:
Modals and extend beyond the viewport height. This PR resolves that as well as removing a fixed height from the modal header.

### DIFF
--- a/lib/Modal/styles.scss
+++ b/lib/Modal/styles.scss
@@ -18,10 +18,10 @@ $modal-media-close-dimensions: 30px;
 
 .legacy-modal {
 	width: 600px;
-	margin: 30px auto;
+
 	.modal-header {
 		@apply border-0 border-b border-neutral-90 border-solid p-3 min-h-16;
-		height: auto;
+
 		.react-modal-header__text {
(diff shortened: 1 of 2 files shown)


lib/Modal/styles.scss
```diff
@@ -18,10 +18,10 @@ $modal-media-close-dimensions: 30px;
 
 .legacy-modal {
 	width: 600px;
-	margin: 30px auto;
```
kyleharper, line 21: I removed this as we central align the modal anyways. This was causing some strange visual in storybook where the certain modals weren't vertically centred 🤔

lib/src/components/modal/modal.scss
```diff
@@ -83,7 +83,7 @@ $modal-small-width: 400px !default;
 }
 
 .react-modal-header {
-  @apply flex flex-row w-full items-center relative h-16;
```
kyleharper, line 86: Fixed heights in UI components are never ideal. This is causing issues in various headers in the app when they overflow with the screen height.

AmeeMorris (approved):

PR #1048, opened 2021-09-10
Fix/gather UI

petemcfarlane:
### 💬 Description
I got a bit distracted when trying to make a simple fix in gather-ui - turns out that eslint/prettier hasn’t been run for quite a while! 

### 🚪 Start Points
The first commit I just added the package.json/package-lock.json files so it's easier to see which dependencies were changed.
The second commit is the result of running prettier and manually fixing the eslint errors.

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook


lib/index.js
```diff
@@ -1,7 +1,7 @@
 import '@babel/polyfill';
+import * as importedConstants from './constants';
 
-export * as constants from './constants';
-
```
petemcfarlane, line 4: ES lint did not like this syntax, so I've imported it with a temporary name "importedConstants" and then re-exported it as "constants". Does this make sense?

PR #1051, opened 2021-09-10
Add colon icon

petemcfarlane:
As the name suggests, adds a `:` icon (used in the sidebar for aspect ratio, e.g. `4  : 3`


petemcfarlane:
Just use a `:` character (in a `<span>` with a margin instead.

PR #1061, opened 2021-09-21
Fix issues with our pre-commit hook

kyleharper:
### 💬 Description
This has been causing issues with the pre-commit hook since #1045 

The assumption was to replace `yarn prettier` with `npm run prettier` but that isn't the case. I changed some files to see if this works and it does the trick 👍 

I'll create a separate PR off this branch, which runs the prettier and linting process on the entire library. 

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook


package.json
```diff
@@ -72,7 +72,7 @@
   },
   "lint-staged": {
     "{lib,tests,stories/webapp}/**/*.js": [
-      "npm run prettier --write",
+      "prettier --write",
```
petemcfarlane, line 75: I also fixed it in https://github.com/gathercontent/gather-ui/pull/1048/commits/24225d159b7bf6cb505844eec75a375100213e6c#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R75 by calling `npm run format:js`, which is similar to this approach but has some extra scoping based on file path - which do you think is the best approach?

PR #1064, opened 2021-09-22
Fixing fade cutting into option menu items

DavidHollins6:
### 💬 Description
We need to full width the text in an option menu item, this is so the fade is at the end of its container. This same fade issue occurs on both usages of `OptionMenuItem`.

### ✅ Checklist
- [ ] Tests written
- [ ] Browser tested
- [ ] Added to documentation
- [ ] Added to storybook


AmeeMorris (approved):
double check this doesn’t break current usages!

PR #1087, opened 2021-11-18
updating tailwind to the latest version

AmeeMorris:
### 💬 Description
See https://tailwindcss.com/docs/upgrading-to-v2 for details on why I’ve changed our config or any scss files.

Pretty much every css/scss package has been updated and the config to match. Tailwind and postcss are at the latest versions, but everything else is about a year old (still several major versions from what we had before though) Sadly we couldn’t update everything since a few things are dependant on webpack 5 for the latest version, and we cant use that until we update storybook 😱


AmeeMorris:
[Lets try updating tailwind](https://trello.com/c/zZvqputU/2294-lets-try-updating-tailwind)

micmcgrorty (approved):

DavidHollins6 (approved):

PR #1090, opened 2021-12-02
Update/storybook 6

AmeeMorris:
### 💬 Description
Hey look its storybook 6 all working, just in time for storybook 7 to come out.

Obviously this is very large so not feasible to actually review. I would recommend pulling this down, running `npm install` then `npm run storybook` and have a click about.

Some highlights:
Stories are now way different, but they're quite easy to get your head around. I'd take a little look through their docs https://storybook.js.org/docs/react/writing-stories/introduction then take a look at a couple here.

Knobs are no more, they are now 'controls' (boo), again take a look at the docs https://storybook.js.org/docs/react/essentials/controls they explain it all better than I can. 

How you set up storybook is also a lot different, check this bit out https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#from-version-52x-to-53x in the migration guide. 

Weirdness... There's one weird not so nice thing I've had to do. Storybook does not like you to have your own custom webpack file anymore. It does allow you to extend it in `main.js` (see https://storybook.js.org/docs/react/configure/webpack) but obviously they're not keen on you doing that since its not the neatest thing to do. 
Storybook wants you to use postcss7, but we need postcss8 in order for tailwind to work. It supposedly lets you use postcss8 if you follow directions here https://storybook.js.org/addons/@storybook/addon-postcss, however this just ignores any postcss plugins you have so it basically does nothing. 
So in our own `main.js` I've had to extend webpack, and also exclude their css rule since it clashes with ours in order for postcss8 to work properly.


AmeeMorris:
[Update storybook](https://trello.com/c/fEdUX5dW/2318-update-storybook)

DavidHollins6 (approved):

micmcgrorty (approved):

PR #1096, opened 2021-12-14
Revert "Add bootstrap styles back in"

kyleharper:
Reverts gathercontent/gather-ui#1095

@AmeeMorris I accidentally merged your PR in #1092 

There are a few issues with it, we can pair on it when you're free 👍


AmeeMorris:
@kyleharper
Screenshot 2022-03-01 at 12 38 02
seems ok to me? local on the left with no bootstrap, qa on the right

PR #1100, opened 2021-12-21
TopBar icon colours

micmcgrorty:
### 💬 Description
TopBar dark mode icons no longer have their colour set by CSS.

A new component, ButtonIconWhite has been added to allow us to use white button icons.
ButtonIconWhite has been added to the button icon story.
### 🔗 Links
(https://trello.com/c/8fhSwIVx/2379-regular-icon-colour-is-incorrect-on-disabled-button-states)
### 👫 Related PRs (optional)
https://github.com/gathercontent/app/pull/4509

### DIFF
--- a/lib/TopBar/styles.scss
+++ b/lib/TopBar/styles.scss
@@ -158,8 +158,7 @@ $top-bar-notification-height: 42px;
       border-color: $neutral-dark;
       color: white;
     }
-    .top-bar__action,
-    .icon:not(.structure__save-btn) path {
+    .top-bar__action {
       fill: white;
     }
     .top-bar__cell--bordered {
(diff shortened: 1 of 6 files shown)


micmcgrorty:
[[REGULAR] Icon colour is incorrect on disabled button states](https://trello.com/c/8fhSwIVx/2379-regular-icon-colour-is-incorrect-on-disabled-button-states)

lib/TopBar/styles.scss
```diff
@@ -152,7 +159,7 @@ $top-bar-notification-height: 42px;
       color: white;
     }
     .top-bar__action,
-    .icon path {
+    .icon:not(.structure__save-btn) path {
```
kyleharper, line 162: 🔴  where is `.structure__save-btn` from? 

Seems like it's from something outside of GUI and shouldn't be added to the styles specifically.

kyleharper (changes requested):

DavidHollins6 (approved):

lib/TopBar/styles.scss
micmcgrorty: Yeah that's a good point. Will need to figure out a different way.

kyleharper (approved):

PR #1118, opened 2022-03-22
update AvatarWithPopover component to use Tippy

micmcgrorty:
### 💬 Description
`AvatarWithPopover` has been updated to use Tippy rather than the `Dropdown` component.

With the `Dropdown` component, when hovering from the bottom of the trigger, the dropdown content overlays the trigger and causes the mouse to 'leave' the trigger, hiding the content. Using Tippy and a standard tooltip stops this from happening.

There is a slight drawback in that the avatar component loses some functionality to allow it to have links or menu items as the dropdown content, but this is never used inside the app and I think it's worth removing unused functionality to fix a bug. It also massively simplifies the component.
### 🔗 Links
(https://trello.com/c/CcQVR0my/2812-presence-avatar-tooltips-flash-and-disappear)
### 📹 GIF (optional)
avatarpopover

### DIFF
--- a/lib/Avatar/AvatarWithPopover.js
+++ b/lib/Avatar/AvatarWithPopover.js
@@ -17,9 +17,7 @@ import Avatar from '.';
 function AvatarWithPopover({
   email,
   name,
-  popoverClass,
   children,
-  theme,
   overlayPosition,
   ...rest
 }) {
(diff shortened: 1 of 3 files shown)


micmcgrorty:
[Presence avatar tooltips flash and disappear](https://trello.com/c/CcQVR0my/2812-presence-avatar-tooltips-flash-and-disappear)

DavidHollins6 (approved):

PR #1160, opened 2022-11-29
update tertiary button hover style

micmcgrorty:
### 💬 Description
Updating tertiary button hover style to match designs.

### DIFF
--- a/lib/src/modules/Button/ButtonTertiary/styles/buttonTertiary.scss
+++ b/lib/src/modules/Button/ButtonTertiary/styles/buttonTertiary.scss
@@ -2,15 +2,16 @@
   @apply border-transparent border-solid text-neutral-20 bg-white;
 
   &:hover {
-    @apply bg-neutral-95 border-neutral-95;
+    background-color: rgba(103, 128, 152, 0.1);
   }
 
   &:active,
   &.button-tertiary-active {
-    @apply bg-neutral-90 border-neutral-90 shadow-none;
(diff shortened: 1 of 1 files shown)


AmeeMorris (approved):

lib/src/modules/Button/ButtonTertiary/styles/buttonTertiary.scss
```diff
@@ -2,15 +2,16 @@
...
 
   &:hover {
-    @apply bg-neutral-95 border-neutral-95;
+    background-color: rgba(103, 128, 152, 0.1);
   }
 
   &:active,
   &.button-tertiary-active {
-    @apply bg-neutral-90 border-neutral-90 shadow-none;
+    background-color: rgba(103, 128, 152, 0.1);
```
webknit, line 10: Looks fine just noticed you took it out of tailwind. Apologies if you are aware, if you didn't want a tailwind config update you, I think you could use tailwind arbitrary for this also

class="bg-[rgba(103,128,152,0.1)]"

webknit (approved):

lib/src/modules/Button/ButtonTertiary/styles/buttonTertiary.scss
micmcgrorty: Thanks for the tip 😄  I think in this case though it's a bit easier to do it in the stylesheet rather than a class.

PR #1165, opened 2022-12-14
update selection provider with data update function

AmeeMorris:
### 💬 Description
We needed a way to update the additional data in our selection provider, this allows us to do so!


AmeeMorris:
[Multiple Workflow Snags V2](https://trello.com/c/vBI8PfGp/4165-multiple-workflow-snags-v2)

webknit (approved):

PR #1178, opened 2023-03-09
GC-508 add prop to form submit button icon

micmcgrorty:
### 💬 Description
Added the `defaultActiveColor={false}` prop to the form submit button so that the white colour is not overridden.


AmeeMorris (commented):
Always uneasy about adding important flags, so just to check which styles are overriding it?

timbryandev (approved):

PR #1182, opened 2023-03-09
GC-518 - Create parity between old and new due date pickers

timbryandev:
### 💬 Description

Part of https://bynder.atlassian.net/browse/GC-518

* [x] remove grey bg from weekend
* [x] Month/Year on the left with arrows on the right

### DIFF
--- a/lib/src/modules/calendar/Calendar.js
+++ b/lib/src/modules/calendar/Calendar.js
@@ -2,6 +2,8 @@ import React from 'react';
 import DayPicker from 'react-day-picker';
 import moment from 'moment';
 
+import cx from 'classnames';
+
 export function Calendar({
   onDayClick,
   selectedDay,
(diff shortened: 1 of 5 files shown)


lib/src/modules/calendar/Calendar.js
```diff
@@ -2,20 +2,36 @@ import React from 'react';
...
+    };
+  }
+
   const weekdayText = ['Sun', 'Mon', 'Tues', 'Wed', 'Thu', 'Fri', 'Sat'];
 
+  const classes = [
+    enableWeekendDays && 'enable-weekend-days',
+    leftAlignHeader && 'left-align-header'
+  ]
+    .filter(Boolean)
```
webknit, line 30: Is this doing the same as the `cx from 'classnames` package that we use throughout gather-ui?

```
const classes = cx(`calendar`, {
    'enable-weekend-days': enableWeekendDays,
    'left-align-header': leftAlignHeader,
  });
  
  <div className={classes}>
```

timbryandev: Great point - forgot we had access to classnames so I'll get it ammended!

lewishankinson (approved):

PR #1198, opened 2023-04-06
GC-651 add fixed due date copy to due date dropdown

AmeeMorris:
### 💬 Description
We have one for dynamic dates, we need one for fixed dates too


timbryandev (approved):
LGTM 👍🏻

PR #1203, opened 2023-05-15
GC-822 - Create initial AI prompt

timbryandev:
### 💬 Description

Create initial AI prompt
### 🔗 Links

* [GC-822](https://bynder.atlassian.net/browse/GC-822)
* https://www.figma.com/file/t741BgLw93yHaDvIbk90RV/Content-assistant?type=design&node-id=549-23974&t=JjP7hMClgoqizE8G-0

### 📹 GIF (optional)
https://share.getcloudapp.com/jkuAxo5m

[GC-822]: https://bynder.atlassian.net/browse/GC-822?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


AmeeMorris (changes requested):
As we discussed in the call - I totally lead you down the wrong path by saying that it should go in GUI. I assumed there would be new ui components that would be needed for this, but it looks like its actually just a collection of existing ui components, so this should be moved over to webapp 👌 

(apart from the icons!)

PR #1206, opened 2023-05-18
GC-857 - Can we use React refs instead of DOM queries

timbryandev:
### 💬 Description
As part of GC-857 we need to swap dom queries for ref hooks for best practice. Problem is, we don't have access to the refs from within gather-ui so we need to expose the refs via prop drilling.

We tried react's forwardRefs but it seems a little over-engineered for this small task.
### 🔗 Links
[_Links that relate to the PR (Trello, Figma etc.)_](https://bynder.atlassian.net/browse/GC-857)
### 📹 GIF (optional)
### 🚪 Start Points
### 👫 Related PRs (optional)

### DIFF
--- a/lib/src/components/form/FormInput.js
+++ b/lib/src/components/form/FormInput.js
@@ -2,15 +2,10 @@ import * as React from 'react';
 import { Input } from 'lib';
 import { FormContext, statuses } from './Form';
 
-export const FormInput = React.forwardRef(function FormInput(
-  { children, ...rest },
-  ref
-) {
+export function FormInput({ children, ...rest }) {
   const { status } = React.useContext(FormContext);
   const hasFailed = status === statuses.failure;
   const isProcessing = status === statuses.processing;
 
-  return (
-    <Input invalid={hasFailed} disabled={isProcessing} {...rest} ref={ref} />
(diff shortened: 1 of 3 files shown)


webknit:
That looks cool and makes sense. I guess my only question - which I have no idea of the answer - is if we don't need a ref, does wrapping out UI components in this manner have any implications?

AmeeMorris (commented):
You can bypass having to use forwardRef by just passing the ref in as a prop under a different name. 

```
export function FormInput({ children, inputRef, ...rest }) {
  ...
  return (
    <Input invalid={hasFailed} disabled={isProcessing} {...rest} ref={inputRef} />
  );
}
```

This used to be an acceptable way of doing it, I think this may have changed with latest react, but we do it for our Form component in remix-gc ;)

timbryandev:
Thanks @AmeeMorris for the feedback, I'll go back and investigate and if that's the case, I'll get a new branch/pr going

@webknit as far as I know the wrapping shouldn't have any implications if the ref is used or needed as it silently handles `undefined` in a graceful way with no side effects.

AmeeMorris (approved):

PR #1210, opened 2023-05-22
Adding keyDown to Dropdown action

webknit:
I need a keyDown event listener on this dropdown action. Followed similar style to the onClick event


AmeeMorris (approved):
rather than having to manually add in props for each event, do you want to just spread the rest of the props in the component? So in the sharedProps object just add `...props` at the end. Might save a pr in the future

PR #1220, opened 2023-06-20
GC-1083 - Fix console log spam in item editor

timbryandev:
### 💬 Description

Fixing some issues that are causing spammy console logs and actual issues potentially getting missed

#BrokenWindows

### DIFF
--- a/lib/Dropdown/DropdownAction.js
+++ b/lib/Dropdown/DropdownAction.js
@@ -20,9 +20,16 @@ const DropdownAction = ({
   ...props
 }) => {
   const {
-    toggleShowContent = props.toggleShowContent,
-    showContent = props.showContent
+    showContent: showContentProp,
+    toggleShowContent: toggleShowContentProp,
+    ...restProps
+  } = props;
+
+  const {
+    toggleShowContent = toggleShowContentProp,
+    showContent = showContentProp
   } = useContext(DropdownContext) || {};
+
   const classNames = cx(`dropdown__action ${className}`, {
     'dropdown__action--danger': danger,
     'dropdown__action--noBg': noBackground,
@@ -47,7 +54,7 @@ const DropdownAction = ({
       if (actionKeyDown) actionKeyDown(e);
     },
     disabled,
-    ...props
+    ...restProps
   };
 
   return isSubmit ? (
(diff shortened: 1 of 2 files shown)


lib/Dropdown/DropdownAction.js
```diff
@@ -20,9 +20,16 @@ const DropdownAction = ({
   ...props
 }) => {
   const {
-    toggleShowContent = props.toggleShowContent,
-    showContent = props.showContent
+    showContent: showContentProp,
+    toggleShowContent: toggleShowContentProp,
+    ...restProps
+  } = props;
```
AmeeMorris, line 26: same q for the one below, why are we destructuring again here rather than in the main props destructuring above?

lib/src/modules/Button/ButtonBase.js
```diff
@@ -18,6 +18,8 @@ function ButtonBase({
   buttonRef,
   ...rest
 }) {
+  const { defaultFillColor, ...restProps } = rest;
```
AmeeMorris, line 21: same as the one above! why not just add `defaultFillColor` after line 18?

timbryandev: fair point on this one - I'll get it changed

lib/Dropdown/DropdownAction.js
timbryandev: in this case there was a naming collision when I tried to destructing in the function arguments directly with the destructuring that happens on the DropdownContext below.

I guess i could rename the values in the component props when destructuring like this but it didn't feel right somehow 

```
...
  value,
  showContent: showContentProp,
  toggleShowContent: toggleShowContentProp,
  ...props
}) => {
  const {
    toggleShowContent = toggleShowContentProp,
    showContent = showContentProp
  } = useContext(DropdownContext) || {};
...
```


I guess my actual issue is refactoring this bit to prevent the collision:
```jsx
  const {
    toggleShowContent = toggleShowContentProp,
    showContent = showContentProp
  } = useContext(DropdownContext) || {};
```

I'll have a rethink - think it was my late afternoon brain failing me on this one

timbryandev: Ah, the other issue was alongside the naming issue, we are also setting default values so that is the dropdown context returns nothing, we default to the values from the props - that tripped up my other attempts too

AmeeMorris (approved):

micmcgrorty (approved):

PR #1221, opened 2023-06-27
Feature/typescript

AmeeMorris:
### 💬 Description
I used airbnbs handy migration tool but had to do **so many** manual changes too. lots of inserting `//` comments in the middle of jsx, test files were migrated to ts files rather than tsx. Then a bunch of eslint, babel, storybook and dependancy hell 

Still, it seems to be working now! Obviously there's an absolute tonne of ignored ts errors that we'll slowly have to go through


micmcgrorty (approved):
🙌

timbryandev (approved):

webknit (approved):
works for me!

PR #1223, opened 2023-07-03
Feature/stylelint

timbryandev:
### 💬 Description
Based on Bynder using stylelint for linting and standards for CSS/SCSS, here we have a PoC in gather-iu.

It looks like a lot of changes, so I've listed the main files for review below around the setup and config. Everything else should just be style and linting changes but might still be worth some review.

### 🔗 Links

https://miro.com/app/board/uXjVP-x4gHc=/?openCardPanel=3458764556492763681

### 🚪 Start Points

* .github/workflows/node.js.yml
https://github.com/Bynder/gathercontent-gather-ui/pull/1223/files#diff-014228303dff9a1af15f4bbd18401f906380129b10ae2a2c62f8b8be592ff88e
* .stylelintrc.json
https://github.com/Bynder/gathercontent-gather-ui/pull/1223/files#diff-528119d81e5a10738ae49f5e20d333ebc37cea63657eb9069281693e6dae0ad3
* package.json
https://github.com/Bynder/gathercontent-gather-ui/pull/1223/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519


micmcgrorty (approved):

lib/Breadcrumb/styles.scss
```diff
@@ -75,7 +79,7 @@ $breadcrumb-spacing: $layout-spacing-base !default;
       display: block;
     }
 
-    &:after {
+    &::after {
```
AmeeMorris, line 82: What's with this rule? is there any benefit to adding two colons?

AmeeMorris (approved):
Really nice that it's cleaned up some super old scss and got rid of deprecated browser prefixes 🙌

lib/Breadcrumb/styles.scss
timbryandev: I think you could use single colons for psudo-elements originally, but as of the CSS3 spec we should use double-colon selectors for pseudo-elements like `first-letter`, `before` and `after`, and
single-colon selectors for pseudo-selectors such as `nth-child`.

We just used the standard stylelint ruleset for SCSS and it follows the CSS3 spec.
That said, most modern browsers accept both syntaxes without issue anyway.

AmeeMorris: Ah cool, thanks for the explainer :)

PR #1230, opened 2023-07-25
Feature/revision list

AmeeMorris:
### 💬 Description
This is a little bit of a chunker. Its all the ui components for this: https://www.figma.com/file/66iSurBAx32DWrJymo9NKL/GUI---Collections?type=design&node-id=9433-21356&mode=design&t=5xWX0MT9iysHynPA-0 which is being used here: https://www.figma.com/file/cQiZrM4OPBTYrJji3L08HY/Expand-revisions?type=design&node-id=3-4052&mode=design&t=M0l316PouJBVZ8Zz-0 

Its a classic compound component which gives you all the little building blocks to make our revisions list. I tried for a while to make the nested lists show and hide using a details element, but couldn't quite get it to play ball so a classic button hooked up to state for our story example is what we'll roll with.


lib/src/modules/revisionList/RevisionListItem.tsx
```diff
@@ -0,0 +1,25 @@
...
+interface Props extends HTMLAttributes<HTMLLIElement> {
+  leftContent?: React.ReactNode;
+  active?: boolean;
+}
+
+export function RevisionListItem({
+  children,
+  className = "",
+  leftContent,
+  active,
```
timbryandev, line 13: Should we default to false if this is an optional param?

timbryandev (approved):

PR #1232, opened 2023-07-26
Adding flex to second child with the addition of the new drag element

webknit:
As part of the [new designs](https://www.figma.com/file/bPJSeiwE1tQ8dm2F6MMdkm/Selection?node-id=391%3A11237&mode=dev) the first element is now a drag element, so this flex needs to be applied to the second.


webknit:
Closing this as will need to apply styles with a css flag of some kind within webapp as this change will apply something globally.

PR #1243, opened 2023-08-16
closing gap from tooltip

webknit:
[Selection toolbar tooltips are too far away. (After further inspection, this is true across the app)](https://bynder.atlassian.net/browse/GC-1793)

Closing the gap of the tooltips. There doesn't appear to be a way to do this via the [plugin](https://atomiks.github.io/tippyjs/#placements), so I styled with CSS. Some formatting styles snuck in there too.

Screenshot 2023-08-16 at 11 22 23


timbryandev (approved):
Yeah bit weird that some linting has snuck in, but its not functionally changing anything so I guess its okay

PR #1258, opened 2023-08-31
GC-2136 Migrate CheckToggle tests

lewishankinson:
### 💬 Description
- Migrated the CheckToggle tests
- Combined most of the modifier classname tests into one
- Skipped some stinkers (e.g. checking the checkbox is checked when checked)
- Deleted the old test file


lib/CheckToggle/index.tsx
```diff
@@ -117,6 +117,7 @@ export class CheckToggle extends Component {
             type="checkbox"
             id={id}
             className="toggle-switch toggle-switch--inline"
+            role="checkbox"
```
webknit, line 120: role="checkbox" must also include the aria-checked attribute I think @lewishankinson 
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/checkbox_role

But more importantly is there a reason you are adding the role, as i think all native elements have this by default. Eg a checkbox is a checkbox there would be no need to inform users of its role?

webknit (changes requested):

lib/CheckToggle/index.tsx
AmeeMorris: Yeah it should have this by default

webknit: Just seen you probs added it for `screen.getByRole('checkbox')`. I guess it wont select it if the role not added?

Maybe could target the test-id instead?

or as per testing-library docs adding a `name`... Most often, this will be used with the name option like so: `getByRole('button', {name: /submit/i})`

https://testing-library.com/docs/queries/about/#types-of-queries

lewishankinson: @webknit Good spot. The locator wasn't finding the checkbox by role initially but I think it was a red herring, a separate issue. I've removed it and the tests still pass so it's not needed 👍 Thanks for the additional info too!

webknit (approved):

PR #1260, opened 2023-09-01
remove low value SectionHeader test

AmeeMorris:
### 💬 Description
The test is just checking that children we've passed in inside the test are rendering, so lets ditch, we dont need tests checking that react works as we expect


webknit (approved):
Been reading about tests recently and you see so much of this stuff to get more coverage etc and it just seems so pointless to me, so I am glad to see this. :-)

timbryandev (approved):
👋🏻 So long

PR #1263, opened 2023-09-01
move search tests over to vitest

AmeeMorris:
### 💬 Description
Old tests were a lot of not super useful 'it renders' style tests, this one hopefully just tests the functionality a bit more


lib/Search/.specs/Search.spec.tsx
```diff
@@ -0,0 +1,40 @@
...
+        value: "PLOP",
+      },
+    });
+    expect(screen.getByText("I'm some options"));
+    expect(screen.getByText("Waffles"));
+    expect(screen.getByText("Pancakes"));
+
+    fireEvent.click(screen.getByTitle("Clear search"));
+    fireEvent.click(screen.getByTitle("Close search"));
+  });
```
lewishankinson, line 39: Do we need any assertions after these to check whether the search has been cleared and closed?

AmeeMorris: We should! Done :)

lewishankinson (approved):

PR #1272, opened 2023-09-04
GC-2180 Migrate LoadingOverlay tests to Vitest

AmeeMorris:
### 💬 Description
Bosh


lib/LoadingOverlay/.specs/LoadingOverlay.spec.tsx
```diff
@@ -0,0 +1,36 @@
+import { describe, expect, it, vi } from "vitest";
+
+import React from "react";
+
+import { render, screen } from "@testing-library/react";
+import LoadingOverlay from "..";
+
+describe("LoadingOverlay", () => {
+  it("can optionally load the svg, text, fixed modifier, and a loading percentage", () => {
```
timbryandev, line 9: This test seems to be doing/expecting a lot.

Do tests like these make sense to be broken up into smaller chunks of checks/expects as it looks like we're checking?

AmeeMorris: Fair point! I've split them up apart from text/loading percentage as those props are tied to each other in the component

lewishankinson (approved):

PR #1277, opened 2023-09-05
GC-2180 fix types that were getting in the way of Vitest migration

AmeeMorris:
### 💬 Description
An unfinished FormModal test has accidentally snuck in too, will finish in follow up pr 🤦‍♀️ 

I felt bad littering more `@ts-expect-errors` in the code so fixed up some types that snowballed slightly


timbryandev (approved):
image

PR #1288, opened 2023-09-05
GC-2136 Migrate EditableChoiceInput tests

lewishankinson:
### 💬 Description
- Installed the user-event companion library for better simulation of user events
- - This is important for events like a user typing in a disabled textbox, where using fireEvent.change() would ignore the disabled attribute and set the value of the textbox, userEvent.type() does not
- Migrated the tests
- Deleted the old tests


AmeeMorris (approved):
Nice!

timbryandev (approved):

PR #1292, opened 2023-09-06
Dismissible prompt test

webknit:
🥫


micmcgrorty (approved):

lib/DismissiblePrompt/.specs/DismissiblePrompt.spec.tsx
```diff
@@ -0,0 +1,40 @@
...
+      >
+        Hello world
+      </DismissiblePrompt>
+    );
+    expect(screen.getByText("Hello world"));
+    expect(
+      screen
+        .getByTestId("test-dismissible-prompt")
+        .classList.contains("hellooooo")
+    ).toBeTruthy();
```
timbryandev, line 30: I'm not 100% sure on this opinion, but I'll share it anyway in case it's helpful (doesn't necessarily need actioning)...

If asserting that the class has been added, should we use a screen query that grabs element by the given class and check they exist that way?

timbryandev (changes requested):

lib/DismissiblePrompt/.specs/DismissiblePrompt.spec.tsx
webknit: I tried this at first but couldn't figure out how to target the DOM, eg `screen.getElementsByClassName` and `screen.querySelectorAll` you can't do.

Looking through docs but unsure what would be used to specifically target classnames @timbryandev

AmeeMorris: I dont think we need to test that the className prop works, thats just core react so we can probs just ditch imo!

AmeeMorris (approved):

timbryandev (approved):

PR #1293, opened 2023-09-06
Update/vitest/confirmation overlay

webknit:
Swapping `ConfirmationOverlay` tests for vitest. Also adding a spread operator into the component so can add additional props (in this case the testId).


AmeeMorris (approved):

lewishankinson (approved):
👍

micmcgrorty (approved):

PR #1296, opened 2023-09-08
💄  GC-1464 Allow users to resize the sidebar in the Content Hub

timbryandev:
### 💬 Description

* A new component designed to wrap existing elements and make them resizable. In this specific instance we're keeping the sidebar menu in mind.

* Within the vite config, It looks like most of the changes are formatting related. However, there is an actual change on line 11 were I've added tsx (`(x)?`) to the include list: `include: ["**/.specs/**/*.+(spec).ts(x)?"],`

Recording:
https://www.loom.com/share/9bd52f0c9d9248258b4f7eaa57b7efe7

### 🔗 Links

- https://bynder.atlassian.net/browse/GC-1464
- https://www.figma.com/file/UfcvpzNN59iIo3CciO59wb/Finder-resize?type=design&node-id=404-10007&mode=design&t=kszwEB9TC9JsMapJ-0


lib/helpers.ts
```diff
@@ -1,16 +1,54 @@
...
+      user.name
+        .toLowerCase()
+        .split(" ")
+        .filter((subStr: any) => subStr.lastIndexOf(safeTerm, 0) === 0).length >
+        0 ||
+      user.name.toLowerCase().lastIndexOf(safeTerm, 0) === 0 ||
+      user.display.toLowerCase().lastIndexOf(safeTerm, 0) === 0 ||
+      (searchByEmail && user.email.toLowerCase().lastIndexOf(safeTerm, 0) === 0)
   );
 };
```
timbryandev, line 19: These are just linting changes...

Actual changes below:

lib/Resizeable/index.tsx
```diff
@@ -0,0 +1,104 @@
...
+    const gutter = handle.parentElement as HTMLSpanElement;
+    const handleOffset = handle.offsetHeight / 2;
+
+    const y = keepValueWithinRange(
+      evt.clientY - rect.top,
+      handleOffset,
+      gutter.offsetHeight - handleOffset
+    );
+
+    handle.style.top = `${y - handleOffset}px`;
```
timbryandev, line 78: I'm having a brain fart, but I'm sure there is a nicer way of handling the offset required here

micmcgrorty (approved):
Awesome stuff 🙌

lib/Resizable/index.tsx
```diff
@@ -0,0 +1,123 @@
...
+  defaultWidth?: number | string;
+  minWidth?: number | string;
+  maxWidth?: number | string;
+  useGutterOffset?: boolean;
+  containerWidth?: number | string;
+}
+
+/**
+ * A wrapper for making a given child element resizable
+ * @param children The element to be resized
```
kyleharper, line 15: What's the reasoning for having types and comment blocks 🤔 

They are already out of sync and like in this case shouldn't `children` be added to the interface but it isn't it's only described in the comment block? 🤔

kyleharper: For me comment blocks are currently used to explain a complex or smelly piece of implementation, I don't see how the comment blocks add anything other than a maintenance overhead for future refactors. Some ideas to remove the need for the comment blocks...

-  `maxWidth` to `maxResizableWidth`
-  `minWidth` to `minResizedWidth`
- `defaultWidth` to `initialWidth`
- add type for `children` as I am assuming it can't be any react node (e.g. a fragment)

lib/Resizable/index.tsx
```diff
@@ -0,0 +1,123 @@
...
+export function Resizable({
+  children,
+  defaultWidth = "50%",
+  maxWidth,
+  minWidth,
+  useGutterOffset = false,
+  ...rest
+}: PropsWithChildren<ResizableProps>) {
+  const containerWidth = toPixels(
+    rest.containerWidth ?? document.body.offsetWidth
```
kyleharper, line 31: Why is `containerWidth` stored in `rest` but other properties aren't? 🤔

lib/Resizable/index.tsx
```diff
@@ -0,0 +1,123 @@
...
+  };
+
+  const doDrag = (evt: MouseEvent) => {
+    const newWidth = state.startWidth + evt.clientX - state.startX;
+    setWidth(newWidth);
+  };
+
+  const stopDrag = () => {
+    document.removeEventListener("mousemove", doDrag, false);
+    document.removeEventListener("mouseup", stopDrag, false);
```
kyleharper, line 62: These events will never be removed.

This is because on re-render `stopDrag` and `doDrag` will have a different signature and `removeEventListener` won't work.

Talk to @micmcgrorty about this, we had a similar issue with editor server events recently.

lib/.specs/helpers.spec.ts
```diff
@@ -0,0 +1,48 @@
+import { describe, expect, it, vi } from "vitest";
+import { keepValueWithinRange, toPixels } from "../helpers";
+
+describe("helpers: toPixels", () => {
```
kyleharper, line 4: 👍

vite.config.ts
```diff
@@ -1,36 +1,36 @@
 /// <reference types="vitest" />
-import { defineConfig } from 'vite';
```
kyleharper, line 2: This just a file that hasn't been formatted?

kyleharper (changes requested):

lib/Resizable/index.tsx
timbryandev: Thanks Kyle. I think at the time I was having a bit of trouble trying to remember what props I needed and why I needed them as I was working around new problems and iterating. I'll take them out and refactor the props & interface as advised now the solution is "complete".

lib/Resizable/index.tsx
timbryandev: Yeah this doesn't look great. I was lazily looking for a way to dynamically set the default value for `containerWidth` if it was undefined but I didn't want to do it as a default param in case a falsy value such as `0` gets passed in and then overridden by my default.

I'll clean this up!

lib/Resizable/index.tsx
timbryandev: Interesting point! I'll ask Michael about it :)

vite.config.ts
timbryandev: It looks like most of the changes are formatting related. 

However, there is an actual change on line 11 were I've added tsx (`(x)?`) to the include list:
```javascript
include: ["**/.specs/**/*.+(spec).ts(x)?"],
```

timbryandev: Sorry - I should have highlighted this in the description!

lib/Resizable/index.tsx
timbryandev: Just had a look back at the code to refresh my memory and we are actually defining the `children` prop via PropsWithChildren: `}: PropsWithChildren<ResizableProps>) {`

kyleharper (approved):

PR #1298, opened 2023-09-19
fix ReactMentions props and fix tests that stop local linting

AmeeMorris:
### 💬 Description
Newer ReactMentions have changed what props you need to pass to what component, this fixes that.

I also had errors in these tests when linking locally, so i've just fixed the imports for those too


webknit (approved):
How come the `afterEach` has to be imported when it doesn't look to be used?

AmeeMorris:
> How come the `afterEach` has to be imported when it doesn't look to be used?

@webknit it is used 😅 https://github.com/Bynder/gathercontent-gather-ui/pull/1298/files#diff-14214aaac303045d2a034883a558bd72326ec00e0698886dfe25e3eb7cd6e4f0R22 same for all the files i've imported it in. used to be globally available when we had jest, but no longer with vitest. thats why its suddenly an issue

PR #1305, opened 2023-09-26
✨  GC-1464 Remember the last resize position

timbryandev:
### 💬 Description

Builds on the new Resizable component. This PR adds a new useLocalStorage hook to record the last width a component may have been resized to and can be restored to using that locally recorded value.

Uses the new `rememberPosition` boolean prop to enable the functionality, and will use the new optional `id` string prop to record instances uniquely.

Added a new story too to demonstrate usage.

### DIFF
--- a/lib/hooks/useLocalStorage.ts
+++ b/lib/hooks/useLocalStorage.ts
@@ -1,0 +1,28 @@
+import { useState } from "react";
+
+export const useLocalStorage = (key: string, defaultValue: unknown) => {
+  const [localStorageValue, setLocalStorageValue] = useState(() => {
+    try {
+      // Try and retrieve any previously recorded values
+      const value = localStorage.getItem(key);
+      if (value) {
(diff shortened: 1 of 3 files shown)


micmcgrorty (approved):
Awesome stuff 🙌

lib/hooks/useLocalStorage.ts
```diff
@@ -0,0 +1,28 @@
+import { useState } from "react";
+
+export const useLocalStorage = (key: string, defaultValue: unknown) => {
+  const [localStorageValue, setLocalStorageValue] = useState(() => {
+    try {
+      // Try and retrieve any previously recorded values
+      const value = localStorage.getItem(key);
```
AmeeMorris, line 7: I think you could name this `currentValue`/`previousValue` and then just do away with the comments

AmeeMorris (approved):

PR #1308, opened 2023-10-02
Adding new Bynder logo

webknit:
- Adding new SVG
- Restricting max width specifically for bynder logo
- Applying border radius


AmeeMorris:
Whats your plan to featureflag this? might be worth having both in GUI then we can pass in a prop (controlled by a flag) from webapp that will switch between the two

lib/Logo/index.tsx
```diff
@@ -9,11 +9,12 @@ import logoSVG from "../../assets/logo.svg";
 
 export function Logo(props: any) {
   const LogoPath = props.url || logoSVG;
+  console.log(LogoPath);
```
AmeeMorris, line 12: log

lib/LogoGC/index.tsx
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+export function LogoGC(props: any) {
+  const LogoPath = props.url || logoSVG;
+  const image =
+    typeof LogoPath === "string" ? (
+      <img src={props.url} alt={props.alt} className="logo__image" />
+    ) : (
+      <span className="logo__image">
+        <LogoPath />
```
AmeeMorris, line 17: This points to the same svg as the new bynder one, we need the old svg in place :)

webknit: Added old SVG back and also added a new story so it's clear what's what.

timbryandev (approved):

PR #1310, opened 2023-10-05
fix/FSB-6089_App-is-unresponsive-when-used-in-Safari-browser

timbryandev:
### 💬 Description

Safari doesn't seem to like using a tab-index of null so lets opt for the more safe -1 :D

https://developer.apple.com/documentation/safari-release-notes/safari-16_4-release-notes


lewishankinson:
Yas 🙌

lewishankinson (approved):

PR #1319, opened 2023-11-02
🏷️ Prevent collision with native HTML input size property

timbryandev:
This should resolve the type conflict between our custom size prop and the native HTML input attribute - as whitnessed in this CI process https://github.com/Bynder/gathercontent-gather-ui/pull/1318#discussion_r1380081027.


micmcgrorty (approved):

lewishankinson (approved):
Nice! 👍

PR #1327, opened 2023-11-29
GC-3696 update blue primary and add new screen sizes

micmcgrorty:
### 💬 Description
Updated the blue primary colour to match mono, and update the screen sizes object to match mono.

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,7 @@
     "dist"
   ],
   "scripts": {
-    "build": "rm -rf dist && babel --extensions '.ts,.tsx' lib/ --out-dir dist/lib && babel tailwind.config.js --out-dir dist && npm run compile:css && npm run legacy:copy-styles && npm run generate-types",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -57,7 +57,7 @@
     "dist"
   ],
   "scripts": {
-    "build": "rm -rf dist && babel --extensions '.ts,.tsx' lib/ --out-dir dist/lib && babel tailwind.config.js --out-dir dist && npm run compile:css && npm run legacy:copy-styles && npm run generate-types",
```
timbryandev, line 60: Is there a reason we're removing the type generation?
I could be wrong, but I think we import those types in SSR so we can have type safety where we're using the mono components?

micmcgrorty: Oops, didn't mean to commit that.

timbryandev: Ah no worries :) all good!

timbryandev (approved):

